PR #7, opened 2018-08-22
Feature: Enable flag for not raise VM errors on RPC responses in ganache

apanizo:
**Description**
In the near future, we will integrate the tx-history-service following the PersonalSafe contracts approach.

Used here: https://github.com/gnosis/safe-contracts/blob/development/contracts/GnosisSafePersonalEdition.sol#L164

For doing it successfully we have to enable the **--noVMErrorsOnRPCResponse** flag in ganache, otherwise when we calculate the process of calculating needed gas on inner safe transactions will fail

### DIFF
--- a/docker/ganache/Dockerfile
+++ b/docker/ganache/Dockerfile
@@ -2,4 +2,4 @@ FROM node:9-alpine
 
 RUN npm install -g ganache-cli
 
-CMD ["ganache-cli", "-d", "--defaultBalanceEther", "10000", "-a", "10", "-h", "0.0.0.0"]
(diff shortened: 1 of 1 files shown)


giacomolicari:
This can be helpful for development purposes, however make sure your code would work as well when connecting to 3rd party services like infura (mainnet, rinkeby etc..) as you wouldn't own the Ethereum node's configuration in that case. @apanizo @rmeissner

rmeissner:
ganache is the only network to my knowledge that has a different default behaviour, therefore we need the flag for ganache.

giacomolicari:
ok 👍 Les't include it then :)

PR #69, opened 2020-02-25
Closes #68: Make webhook independent from Safe address

### LINKED ISSUE #68: Webhooks should be independent of Safe address
Currently you have to specify a Safe for which you want to receive a webhook call. It should be possible to specify that you want to receive for all Safes a webhook call.

#### Code to change
https://github.com/gnosis/safe-transaction-service/blob/master/safe_transaction_service/history/tasks.py#L201

rmeissner:
Closes #68 

* Address has been removed from the webhook and all webhooks are always triggerd
* It is now possible that multiple webhooks get notified for the same Safe

#### To be discussed
It might make sense to keep the address as an optional to allow adding Safe specific webhooks, but it feels this is not needed right now and just increases complexity


Uxio0:
Implemented here 45f315a604606e94293ff5d55294ceed931b8645, thanks for the PR!

PR #72, opened 2020-03-05
Notify new transaction history releases in slack

davidalbela:
- Add send_slack_notification commandline.
  - Add send message to slack with the name of the app and the version.


Uxio0:
Thanks @davidalbela !

PR #120, opened 2020-07-15
update swagger and error codes in some places

sche:
I updated the swagger doc and 422 error codes on endpoints that we will use in the first public release.
I think it makes sense to move `if not Web3.isChecksumAddress(address)` checks into a decorator.
I didn't change it in all other places.


Uxio0:
Agree, thanks for the PR

PR #340, opened 2021-08-26
Create django.yml

aalee888:
### What was wrong?

Related to Issue #


Uxio0:
I don't see this is needed, we already have a CI workflow

Issue #348, opened 2021-08-31
Enabled modules on setup and other custom calls are not indexed (event indexing)


Uxio0:
**Describe the bug**
When a module is enabled during `setup` (using custom `to` and `data`), event is picked by the `SafeEventIndexer` but not processed by the `TxProcessor`

**To Reproduce**
1. Create a new Safe and enable a module during setup
2. Enable module event will have a lower index, so it will be processed before the `SafeSetup`. As `SafeSetup` has not been seen yet, `EnableModule` event is ignored

**Expected behavior**
Enabled module is shown on the `SafeStatus` model

PR #351, opened 2021-09-02
Fix setup modules

Uxio0:
### What was wrong?

Closes issue #348 

### How was it fixed?
As multiple events can be emitted before the `SafeSetup` event (using custom `to` and `data` during Safe creation) `SafeSetup` event should be processed first

@gnosis/safe-services


fmrsabino (approved):
🚀 🚀 🚀

PR #371, opened 2021-09-15
Add support for native coin to /tokens/price/usd

fmrsabino:
- Add support for native coin fiat price in ` /tokens/<address>/price/usd`
- In order to retrieve the native coin of the chain tied to this service, the address to be used should be `0x0000000000000000000000000000000000000000` so from the client perspective the endpoint path would be `/tokens/0x0000000000000000000000000000000000000000/price/usd`
- If an error occurs when fetching the new price, `0` is returned instead. This is line with the other token prices results in case of failure.

### DIFF
--- a/safe_transaction_service/tokens/services/price_service.py
+++ b/safe_transaction_service/tokens/services/price_service.py
@@ -11,7 +11,6 @@ from cache_memoize import cache_memoize
 from cachetools import TTLCache, cachedmethod
 from celery.utils.log import get_task_logger
 from eth_typing import ChecksumAddress
-from gnosis.eth.constants import NULL_ADDRESS
 from redis import Redis
 
 from gnosis.eth import EthereumClient, EthereumClientProvider
@@ -248,14 +247,6 @@ class PriceService:
         except CannotGetPrice:
             logger.warning('Cannot get network ether price', exc_info=True)
             eth_price = 0
-
-        # If the token is the native one (NULL_ADDRESS) we can return the eth_price immediately
-        if len(token_addresses) == 1 and token_addresses[0] == NULL_ADDRESS:
-            yield FiatPriceWithTimestamp(
-                fiat_price=eth_price,
-                fiat_code=
-                FiatCode.USD, timestamp=timezone.now()
-            )
 
         for token_eth_values_with_timestamp in self.get_cached_token_eth_values(token_addresses):
             yield FiatPriceWithTimestamp(eth_price * token_eth_values_with_timestamp.eth_value,
--- a/safe_transaction_service/tokens/services/price_service.py
+++ b/safe_transaction_service/tokens/services/price_service.py
@@ -11,6 +11,7 @@ from cache_memoize import cache_memoize
 from cachetools import TTLCache, cachedmethod
 from celery.utils.log import get_task_logger
 from eth_typing import ChecksumAddress
+from gnosis.eth.constants import NULL_ADDRESS
 from redis import Redis
 
 from gnosis.eth import EthereumClient, EthereumClientProvider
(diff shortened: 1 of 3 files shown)


safe_transaction_service/tokens/services/price_service.py
```diff
@@ -248,6 +249,14 @@ def get_cached_usd_values(self, token_addresses: Sequence[ChecksumAddress]) -> I
             logger.warning('Cannot get network ether price', exc_info=True)
             eth_price = 0
 
+        # If the token is the native one (NULL_ADDRESS) we can return the eth_price immediately
+        if len(token_addresses) == 1 and token_addresses[0] == NULL_ADDRESS:
```
Uxio0, line 253: I don't think this is ok. I think it shouldn't be important the number of `token_addresses` received. Maybe yielding `1` in `get_cached_token_eth_values` if `token == NULL_ADDRESS`?

Uxio0: Also, we can reduce the cache for `get_eth_usd_price`

Uxio0 (changes requested):

safe_transaction_service/tokens/services/price_service.py
fmrsabino: > I don't think this is ok. I think it shouldn't be important the number of `token_addresses` received. Maybe yielding `1` in `get_cached_token_eth_values` if `token == NULL_ADDRESS`?

Good point! Much better now. Actually `get_cached_token_eth_values` was already doing that so it was a matter of adjusting the condition on the view to not require the native coin to be in the Token table. Also adjusted the test for it.

> Also, we can reduce the cache for get_eth_usd_price

I would do this as part of a different PR if we agree on a new value 👍

Uxio0:
The check for the `NULL_ADDRESS` is still missing https://github.com/gnosis/safe-transaction-service/blob/master/safe_transaction_service/tokens/services/price_service.py#L166 😉

fmrsabino:
> The check for the `NULL_ADDRESS` is still missing https://github.com/gnosis/safe-transaction-service/blob/master/safe_transaction_service/tokens/services/price_service.py#L166 😉

It is checking here and returning it: https://github.com/gnosis/safe-transaction-service/blob/0ce1cdbdfca9d5227cae85cbf9d26f83fb14b162/safe_transaction_service/tokens/services/price_service.py#L227-L228

fmrsabino:
> > The check for the `NULL_ADDRESS` is still missing https://github.com/gnosis/safe-transaction-service/blob/master/safe_transaction_service/tokens/services/price_service.py#L166 😉
> 
> It is checking here and returning it:
> 
> https://github.com/gnosis/safe-transaction-service/blob/0ce1cdbdfca9d5227cae85cbf9d26f83fb14b162/safe_transaction_service/tokens/services/price_service.py#L227-L228
> 
> EDIT: Will double check the condition above ☝️

@Uxio0 Wouldn't the condition that I shared have to change to:

```python
if token_address == NULL_ADDRESS:
```

instead of

```python
if not token_address:
```

Uxio0:
`0x00..00` str does not evaluate to `False`, so that check is not valid. It should be `if token_address == NULL_ADDRESS` and it should be on this method: https://github.com/gnosis/safe-transaction-service/blob/master/safe_transaction_service/tokens/services/price_service.py#L166

Uxio0:
I just pushed a commit, maybe it's easier to explain 😃

fmrsabino:
> I just pushed a commit, maybe it's easier to explain 😃

I see! Thanks yeah that makes sense 👍

Uxio0 (approved):

PR #382, opened 2021-09-21
Retry gas estimation if there's a node connection error

Uxio0:


safe_transaction_service/history/serializers.py
```diff
@@ -241,12 +241,16 @@ def save(self, **kwargs):
...
-            raise NodeConnectionException(f'Node connection error when estimating gas for safe {safe_address}') from exc
-        return {'safe_tx_gas': safe_tx_gas}
+        exc = None
+        # Retry twice to get an estimation
+        for _ in range(2):
+            try:
+                safe_tx_gas = safe.estimate_tx_gas(self.validated_data['to'], self.validated_data['value'],
+                                                   self.validated_data['data'], self.validated_data['operation'])
+                return {'safe_tx_gas': safe_tx_gas}
+            except IOError as _exc:
```
jpalvarezl, line 251: Should we handle this like you did here: https://github.com/gnosis/safe-transaction-service/pull/383

safe_transaction_service/history/serializers.py
```diff
@@ -241,12 +241,16 @@ def save(self, **kwargs):
...
         safe = Safe(safe_address, ethereum_client)
-        try:
-            safe_tx_gas = safe.estimate_tx_gas(self.validated_data['to'], self.validated_data['value'],
-                                               self.validated_data['data'], self.validated_data['operation'])
-        except IOError as exc:
-            raise NodeConnectionException(f'Node connection error when estimating gas for safe {safe_address}') from exc
-        return {'safe_tx_gas': safe_tx_gas}
+        exc = None
+        # Retry twice to get an estimation
+        for _ in range(2):
```
jpalvarezl, line 246: Maybe if you need to retry twice, it would be better to do it in the `expect` block where you would handle the first failure 🤔 It would add a lot of nesting though, and the result is the same but yeah, maybe it would be more explicit.

jpalvarezl (approved):
I would approve to not block you, just a couple of minor comments :)

safe_transaction_service/history/serializers.py
jpalvarezl: The advantage of this, is that you could log the error once and log that you are "retrying" in case that is helpful

safe_transaction_service/history/serializers.py
Uxio0: As I'm not using Multicall here I'm not expecting a `ValueError`, but adding it will not harm. Nice catch!

safe_transaction_service/history/serializers.py
Uxio0: Good suggestion, but I don't want to log as the error is due to using shared nodes and them being busy when asking for the gas estimation 😉

safe_transaction_service/history/serializers.py
Uxio0: Added it

PR #383, opened 2021-09-21
Fix internal server errors on node errors

Uxio0:


jpalvarezl (approved):
Would it make sense to do something with the value or log it somehow?

Uxio0:
> Would it make sense to do something with the value or log it somehow?

`raise NodeConnectionException from exc` will be catched and logged by Django: https://github.com/gnosis/safe-transaction-service/blob/master/safe_transaction_service/history/exceptions.py#L12

PR #385, opened 2021-09-22
Support CELO Safe deployments

Uxio0:
- https://github.com/gnosis/safe-deployments/pull/18


jpalvarezl (approved):
LGTM, just curious that the contracts could no be deployed in the usual address.

Uxio0:
> LGTM, just curious that the contracts could no be deployed in the usual address.

Not possible on some networks as they require the `chainId` on the transactions

PR #389, opened 2021-09-23
Allow delegates to delete themselves

Uxio0:


safe_transaction_service/history/serializers.py
```diff
@@ -258,15 +259,33 @@ class SafeDelegateDeleteSerializer(serializers.Serializer):
...
+
+    def get_valid_delegators(self, ethereum_client: EthereumClient, safe_address: ChecksumAddress,
+                             delegate: ChecksumAddress) -> List[ChecksumAddress]:
+        """
+        :param ethereum_client:
+        :param safe_address:
+        :param delegate:
+        :return: Valid delegators for a Safe. A delegate should be able to remove itself
+        """
+        return self.get_safe_owners(ethereum_client, safe_address) + [delegate]
```
jpalvarezl, line 277: Are owners considered `delegates` in this method?

safe_transaction_service/history/serializers.py
```diff
@@ -321,6 +334,16 @@ def validate(self, data):
...
 
+    def get_valid_delegators(self, ethereum_client: EthereumClient, safe_address: ChecksumAddress,
+                             delegate: ChecksumAddress) -> List[ChecksumAddress]:
+        """
+        :param ethereum_client:
+        :param safe_address:
+        :param delegate:
+        :return: Valid delegators for a Safe. A delegate shouldn't be able to add itself
+        """
+        return self.get_safe_owners(ethereum_client, safe_address)
```
jpalvarezl, line 345: This is a bit confusion as this appears to be returning owners when the method should return delegators 🤔

safe_transaction_service/history/tests/test_views.py
```diff
@@ -1091,6 +1091,17 @@ def test_delete_safe_delegate(self):
         self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
         self.assertEqual(SafeContractDelegate.objects.count(), 1)
 
+        # Check if a delegate can delete itself
+        SafeContractDelegate.objects.all().delete()
+        delegate_account = Account().create()
+        SafeContractDelegateFactory(safe_contract=safe_contract, delegate=delegate_account.address)
+        hash_to_sign = DelegateSignatureHelper.calculate_hash(delegate_account.address)
+        data['signature'] = delegate_account.signHash(hash_to_sign)['signature'].hex()
+        response = self.client.delete(reverse('v1:history:safe-delegate', args=(safe_address, delegate_account.address)),
```
jpalvarezl, line 1100: I am not sure if this already behaves like so, but if we allow to delete several delagators, should we make this a transactional method?

jpalvarezl (approved):
I have a couple of questions, regarding mostly what the difference of an `owner` and a `delegator` is in this PR. I am just a bit confusded by the methods used in some places. I would approve just not to block you.

safe_transaction_service/history/serializers.py
Uxio0: No, owners are delegators

safe_transaction_service/history/serializers.py
Uxio0: Owners are the delegators (they delegate in a `delegate`)

safe_transaction_service/history/tests/test_views.py
Uxio0: What do you mean?

PR #410, opened 2021-10-08
Add minor improvements for testing

Uxio0:


Uxio0:
> and also with mainnet node URL in `.env`

If you set the mainnet node in `.env` it will work, but the tests involving a mainnet node are ignored

PR #451, opened 2021-10-28
Refactor ethereum events

### LINKED ISSUE #445: Extremely long running DB queries affects the availability of services
There are some long queries taking a lot of DB time. From what we see on the logs, most part are from the `EthereumEvent` table:
- It uses a jsonb field, even if it's indexed it seems that is not the most optimal thing to use (at least with postgres 10)
- Table need to be refactored

Uxio0:
### What was wrong?

- Closes #445 
- Refactor `EthereumEvent` table into `ERC20Transfer` and `ERC721Transfer`.
- Don't use `jsonb` field anymore.
- Refactor queries to `EthereumEvent` and optimize them if possible

### How was it fixed?

@gnosis/safe-services


safe_transaction_service/history/models.py
```diff
@@ -645,55 +593,79 @@ def ether_incoming_txs_for_address(self, address: str):
...
 
     def ether_and_token_incoming_txs(self, address: str):
-        tokens_queryset = self.token_incoming_txs_for_address(address)
+        erc20_queryset = ERC20Transfer.objects.incoming(address).token_txs()
+        erc721_queryset = ERC721Transfer.objects.incoming(address).token_txs()
         ether_queryset = self.ether_incoming_txs_for_address(address)
-        return self.union_ether_and_token_txs(tokens_queryset, ether_queryset)
+        return self.union_ether_and_token_txs(
+            erc20_queryset, erc721_queryset, ether_queryset
+        )
```
fmrsabino, line 647: Isn't it possible to use the `QuerySet` returned by `token_incoming_txs_for_address`? The only set missing would be the ether transfers.

safe_transaction_service/history/models.py
```diff
@@ -297,58 +310,145 @@ def update_with_block_and_receipt(
...
+            parameters["value"] = event_data["args"]["value"]
+            return ERC20Transfer(**parameters)
+        else:
+            raise ValueError(
+                f"Not supported EventData, `value` not present {event_data}"
+            )
+
+
+class ERC721TransferManager(TokenTransferManager):
+    # TODO Optimize this
```
fmrsabino, line 430: Is this `TODO` still valid? 😅

safe_transaction_service/history/models.py
```diff
@@ -297,58 +310,145 @@ def update_with_block_and_receipt(
...
         Returns erc721 owned by address, removing the ones sent
         :return: List of tuples(token_address: str, token_id: int)
         """
         # Get all the token history
-        erc721_events = self.erc721_events(address=address)
+        erc721_events = self.to_or_from(address)
         # Get tokens received and remove tokens transferred
         tokens_in: Tuple[str, int] = []
         tokens_out: Tuple[str, int] = []
         for erc721_event in erc721_events:
```
fmrsabino, line 441: Here we are consuming the the `query_set`. Maybe we can have a look into how to do the logic below on top of the `query_set` API

safe_transaction_service/history/tasks.py
```diff
@@ -63,7 +63,7 @@ def index_erc20_events_out_of_sync_task(
     block_process_limit: Optional[int] = None,
     block_process_limit_max: Optional[int] = None,
     addresses: Optional[ChecksumAddress] = None,
-    number_of_addresses: Optional[int] = None,
+    number_of_addresses: Optional[int] = 100,
```
fmrsabino, line 66: Is `100` the maximum number of addresses to sync (if not provided)?

gunicorn.conf.py
```diff
@@ -1,6 +1,6 @@
 access_logfile = "-"
 error_logfile = "-"
-# max_requests = 500000  # Restart a worker after it has processed a given number of requests (for memory leaks)
+max_requests = 500000  # Restart a worker after it has processed a given number of requests (for memory leaks)
```
fmrsabino, line 3: We are changing this back because of the optimizations of this PR?

fmrsabino (approved):
This is a good change 👏 Left some questions.

gunicorn.conf.py
Uxio0: Yes, this was impacting a lot memory usage on production

safe_transaction_service/history/tasks.py
Uxio0: Yes, we don't want to stress the node and by testing around 100-200 seems like a good value

safe_transaction_service/history/models.py
Uxio0: Yes, it is. I would like to find a way to do this directly on SQL

safe_transaction_service/history/models.py
Uxio0: Where?

safe_transaction_service/history/models.py
Uxio0: That's why I have the TODO. I would like to do this on SQL

safe_transaction_service/history/models.py
Uxio0: Ok, I get you, you are right but I prefer to leave this o a previous optimization, as we will be sorting twice

huyngopt1994:
@Uxio0  Hi just concerning, in this logic we will lost all events which doesn't belong to erc20/721 right ?

Uxio0:
There was no logic to index different events than ERC20/721 and Safe events, so nothing is lost with this PR

huyngopt1994:
Oke thanks I asked this because when I upgraded the version, somehow some history are missed when I try to transfer some native tokens before It can track those event history,  but thanks let me check the logic deeply.

Uxio0:
Please update to the last version of the Safe and take a look at the README, there are instructions in case you are missing transactions so you can reindex them

PR #457, opened 2021-11-02
Fix coveralls

Uxio0:
### What was wrong?

Coveralls is not working, `COVERALLS_REPO_TOKEN` is not needed anymore

### How was it fixed?

Don't use `COVERALLS_REPO_TOKEN`

@gnosis/safe-services


fmrsabino (approved):
Interesting... so it breaks if the environment variable is set? 🤔

Uxio0:
> Interesting... so it breaks if the environment variable is set? thinking

Yes, it looks like it's trying to use it but it's not working anymore

PR #464, opened 2021-11-03
Tell apart ERC20/ERC721 looking at token contract

Uxio0:
Sometimes ERC721 don't follow the standard `Transfer` event, not indexing the `tokenId`. Now when processing every event the token address will be check for `decimals`

### DIFF
--- a/safe_transaction_service/history/admin.py
+++ b/safe_transaction_service/history/admin.py
@@ -3,6 +3,7 @@ from typing import Optional
 from django.contrib import admin
 from django.db.models import F, Q
 from django.db.models.functions import Greatest
+from django.db.transaction import atomic
 
 from hexbytes import HexBytes
 
@@ -111,12 +112,26 @@ class TokenTransferAdmin(admin.ModelAdmin):
 
 @admin.register(ERC20Transfer)
 class ERC20TransferAdmin(TokenTransferAdmin):
-    pass
+    actions = ["to_erc721"]
+
+    @admin.action(description="Convert to ERC721 Transfer")
+    @atomic
+    def to_erc721(self, request, queryset):
+        for element in queryset:
+            element.to_erc721_transfer().save()
+        queryset.delete()
 
 
 @admin.register(ERC721Transfer)
 class ERC721TransferAdmin(TokenTransferAdmin):
-    pass
+    actions = ["to_erc20"]
+
+    @admin.action(description="Convert to ERC20 Transfer")
+    @atomic
+    def to_erc20(self, request, queryset):
+        for element in queryset:
+            element.to_erc20_transfer().save()
+        queryset.delete()
 
 
 @admin.register(EthereumTx)
--- a/safe_transaction_service/history/indexers/erc20_events_indexer.py
+++ b/safe_transaction_service/history/indexers/erc20_events_indexer.py
@@ -13,6 +13,8 @@ from web3.types import EventData, LogReceipt
 
 from gnosis.eth import EthereumClient
 
+from safe_transaction_service.tokens.models import Token
+
 from ..models import ERC20Transfer, ERC721Transfer, SafeContract, TokenTransfer
 from .events_indexer import EventsIndexer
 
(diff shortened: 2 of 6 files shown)


safe_transaction_service/history/indexers/erc20_events_indexer.py
```diff
@@ -102,18 +108,15 @@ def _process_decoded_element(self, event: EventData) -> EventData:
...
-            )
-            value = event_args.pop("unknown")
-            if self._is_erc20(event["address"]):
-                event_args["value"] = value
-            else:
-                event_args["tokenId"] = value
+        if "unknown" in event_args:  # Not standard event
+            event_args["value"] = event_args.pop("unknown")
+
+        if self._is_erc20(event["address"]):
```
jpalvarezl, line 114: Judging by line ~110 here in this file `event_args` is equal `event["address"]` , right? You could probably inline it here.

safe_transaction_service/history/models.py
```diff
@@ -509,6 +521,16 @@ def from_decoded_event(cls, event_data: EventData) -> Union["ERC721Transfer"]:
...
         return self.token_id
 
+    def to_erc20_transfer(self):
+        return ERC20Transfer(
+            ethereum_tx=self.ethereum_tx,
+            address=self.address,
+            _from=self._from,
+            to=self.to,
+            log_index=self.log_index,
+            value=self.token_id,
```
jpalvarezl, line 531: should you be assiging `self.value` instead of `self.token_id`?

safe_transaction_service/history/tests/test_erc20_events_indexer.py
```diff
@@ -76,8 +76,22 @@ def test_erc20_events_indexer(self):
...
             event["args"]["unknown"] = event["args"].pop("tokenId")
-            event["args"]["unknown"] = original_event["args"]["tokenId"]
 
             self.assertEqual(
                 erc20_events_indexer._process_decoded_element(event), original_event
             )
+
+        event = self.ethereum_client.erc20.get_total_transfer_history(
+            from_block=block_number, to_block=block_number
+        )[0]
```
jpalvarezl, line 86: Not sure if this is the sort of case you need to `pop` instead of using `index` access to the array 🤔 Just pointing it out :)

safe_transaction_service/history/tests/test_models.py
```diff
@@ -227,6 +227,30 @@ class TestEthereumTx(TestCase):
...
+        erc20_transfer = ERC20TransferFactory()
+        self.assertEqual(ERC721Transfer.objects.count(), 0)
+        erc20_transfer.to_erc721_transfer().save()
+        self.assertEqual(ERC721Transfer.objects.count(), 1)
+        erc721_transfer = ERC721Transfer.objects.get()
+        self.assertEqual(erc721_transfer.ethereum_tx_id, erc20_transfer.ethereum_tx_id)
+        self.assertEqual(erc721_transfer.address, erc20_transfer.address)
+        self.assertEqual(erc721_transfer.log_index, erc20_transfer.log_index)
+        self.assertEqual(erc721_transfer.to, erc20_transfer.to)
+        self.assertEqual(erc721_transfer.token_id, erc20_transfer.value)
```
jpalvarezl, line 240: Ah I see, sometimes ERC721 transfers get indexed as ERC20 events?

safe_transaction_service/history/indexers/erc20_events_indexer.py
Uxio0: No, `event_args = event["args"]`. I could inline in all the method, but I prefer to have a variable defined so I don't have a typo on `["args"]`

safe_transaction_service/history/models.py
Uxio0: No, as I'm converting a `ERC721` to an `ERC20` 😉

safe_transaction_service/history/tests/test_erc20_events_indexer.py
Uxio0: I could use that, but I don't need to modify the list, only access the element

safe_transaction_service/history/tests/test_models.py
Uxio0: Yes, as the event should be `Transfer(indexed address from, indexed address to, indexed uint256 tokenId)`, but sometimes it is `Transfer(indexed address from, indexed address to, uint256 tokenId)` matching the one for `ERC20` and making it impossible to tell apart

fmrsabino (approved):

safe_transaction_service/history/admin.py
```diff
@@ -111,12 +112,26 @@ def block_number(self, obj: MultisigConfirmation) -> Optional[int]:
...
 @admin.register(ERC721Transfer)
 class ERC721TransferAdmin(TokenTransferAdmin):
-    pass
+    actions = ["to_erc20"]
+
+    @admin.action(description="Convert to ERC20 Transfer")
+    @atomic
+    def to_erc20(self, request, queryset):
+        for element in queryset:
+            element.to_erc20_transfer().save()
```
jpalvarezl, line 133: Isn't there something like `build_update` that could be used so we don't necessarily iterate through everything? Maybe mapp everything and then bulk save to have a single DB access instead of one per item.

jpalvarezl: The same could be done in line `119` of this file

Uxio0: Not an easy way, this method should only be used for fixing some elements and not expecting a lot of transfers in the queryset

PR #472, opened 2021-11-08
Auto reindex when an indexing issue is found for a Safe

### LINKED ISSUE #471: Auto reindex when an indexing issue is found for a Safe
# What is needed?
Auto reindex when an indexing issue is found for a Safe

# Background
Currently when an indexing issue is found on a Safe txs are processed again, but not reindexing is triggered

Uxio0:
### What was wrong?

Closes #471

### DIFF
--- a/safe_transaction_service/history/models.py
+++ b/safe_transaction_service/history/models.py
@@ -972,6 +972,7 @@ class MultisigTransactionManager(models.Manager):
         have an invalid `safeNonce`, so `safeTxHash` is not valid and owners recovered from the signatures wouldn't be
         valid. We exclude `Approved hashes` and `Contract signatures` as that owners are not retrieved using the
         signature, so they will show the right owner even if `safeNonce` is not valid
+
         :param safe:
         :return: Last valid indexed transaction mined
         """
@@ -1023,6 +1024,7 @@ class MultisigTransactionManager(models.Manager):
     def not_indexed_metadata_contract_addresses(self):
         """
         Find contracts with metadata (abi, contract name) not indexed
+
         :return:
         """
         return (
@@ -1056,7 +1058,7 @@ class MultisigTransactionQuerySet(models.QuerySet):
         """
         threshold_query = (
             SafeStatus.objects.filter(internal_tx__ethereum_tx=OuterRef("ethereum_tx"))
-            .sorted_reverse_by_internal_tx()
+            .sorted_reverse_by_mined()
             .values("threshold")
         )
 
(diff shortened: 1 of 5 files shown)


safe_transaction_service/history/models.py
```diff
@@ -1491,23 +1494,34 @@ def __str__(self):
...
         adding owners and enabling a Module in the same contract `execTransaction`)
-        :return: True if corrupted, False otherwise
+
+        :return: `True` if corrupted, `False` otherwise
         """
         return (
-            self.__class__.objects.filter(
-                address=self.address, nonce__lte=self.nonce
-            ).count()
+            self.__class__.objects.distinct("nonce")
```
fmrsabino, line 1509: Is this because we can have multiple `SafeStatus` for the same `nonce`?

safe_transaction_service/history/models.py
```diff
@@ -1491,23 +1494,34 @@ def __str__(self):
...
 
+    def previous(self) -> Optional["SafeStatus"]:
+        """
+        :return: SafeStatus with the previous nonce
+        """
+        return (
+            self.__class__.objects.filter(address=self.address, nonce__lt=self.nonce)
+            .sorted_by_mined()
+            .first()
+        )
```
fmrsabino, line 1523: Would it make sense to use `distinct` here too before the `filter()`?

safe_transaction_service/history/models.py
```diff
@@ -1491,23 +1494,34 @@ def __str__(self):
...
+            .count()
             <= self.nonce
         )
 
+    def previous(self) -> Optional["SafeStatus"]:
+        """
+        :return: SafeStatus with the previous nonce
+        """
+        return (
+            self.__class__.objects.filter(address=self.address, nonce__lt=self.nonce)
```
fmrsabino, line 1520: If `self.nonce` is available on the database doesn't that mean that we have indexed up until that point?

safe_transaction_service/history/tasks.py
```diff
@@ -225,16 +225,34 @@ def process_decoded_internal_txs_for_safe_task(
...
+                                    block_number,
+                                )
+                                index_service.reindex_master_copies(
+                                    from_block_number=block_number
+                                )
+                            logger.error(
+                                "Safe-address=%s Processing traces again",
+                                safe_address,
+                            )
+                            index_service.reprocess_addresses([safe_address])
```
fmrsabino, line 255: Shall we trigger the `index_service.reprocess_addresses` even after we triggered `index_service.reindex_master_copies` with the specific block number?

safe_transaction_service/history/models.py
Uxio0: That's right

safe_transaction_service/history/models.py
Uxio0: Why? I just need the previous one, I guess than using `first (LIMIT 1)` is more optimal than calling `distinct`

safe_transaction_service/history/tasks.py
Uxio0: Of course. Traces need to be processed again, as new ones can appear

fmrsabino (approved):

PR #475, opened 2021-11-08
Use `contextlib.suppress` instead of `try...except`

Uxio0:


safe_transaction_service/utils/tests/test_tasks.py
```diff
@@ -1,16 +1,34 @@
+from unittest import mock
```
fmrsabino, line 1: What is the context for these changes? 😅

fmrsabino (commented):
Looks good to me. I didn't know about `contextlib.suppress`.

There are some other changes that I'm not sure that they are related with the PR 🤔

safe_transaction_service/utils/tests/test_tasks.py
Uxio0: Oh that's right, those changes were supposed to go on the increase coverage PR. I realised some of the tasks were not tested at all

Uxio0: Removed

fmrsabino (approved):

PR #478, opened 2021-11-10
Set reindexing limit

Uxio0:
Don't reindex until the current block, only until the next not problematic transaction

### DIFF
--- a/safe_transaction_service/history/services/index_service.py
+++ b/safe_transaction_service/history/services/index_service.py
@@ -272,6 +272,7 @@ class IndexService:
     def reindex_master_copies(
         self,
         from_block_number: int,
+        to_block_number: Optional[int] = None,
         block_process_limit: int = 1000,
         addresses: Optional[ChecksumAddress] = None,
     ):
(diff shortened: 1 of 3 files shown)


safe_transaction_service/history/services/index_service.py
```diff
@@ -280,10 +281,13 @@ def reindex_master_copies(
         while reindexing
 
         :param from_block_number: Block number to start indexing from
+        :param to_block_number: Block number to stop indexing on
         :param block_process_limit: Number of blocks to process each time
         :param addresses: Master Copy or Safes(for L2 event processing) addresses. If not provided,
             all master copies will be used
         """
+        assert (not to_block_number) or to_block_number > from_block_number
```
fmrsabino, line 289: Given that this statement won't be available if we run with the optimiser flag should we change this to `raise` an exception instead?

safe_transaction_service/history/tasks.py
```diff
@@ -233,17 +233,18 @@ def process_decoded_internal_txs_for_safe_task(
                             logger.error(message)
                             index_service = IndexServiceProvider()
                             if previous_safe_status := safe_status.previous():
-                                block_number = (
-                                    previous_safe_status.internal_tx.ethereum_tx.block_id
-                                )
+                                block_number = previous_safe_status.block_number
```
fmrsabino, line 236: Just to confirm: `previous_safe_status.internal_tx.ethereum_tx.block_id` is equivalent to `previous_safe_status.block_number` (ie.: the respective block numbers should match)?

fmrsabino (approved):

safe_transaction_service/history/tasks.py
Uxio0: Yes, that's only a shortcut

PR #482, opened 2021-11-12
Reindex last hour txs

Uxio0:
Prevents indexing issues

### DIFF
--- a/safe_transaction_service/history/management/commands/setup_service.py
+++ b/safe_transaction_service/history/management/commands/setup_service.py
@@ -76,6 +76,12 @@ TASKS = [
         IntervalSchedule.MINUTES,
     ),
     CeleryTaskConfiguration(
+        "safe_transaction_service.history.tasks.reindex_last_hours",
+        "Reindex master copies for the last hours",
+        110,
+        IntervalSchedule.MINUTES,
+    ),
+    CeleryTaskConfiguration(
         "safe_transaction_service.history.tasks.process_decoded_internal_txs_task",
         "Process Internal Txs",
         2,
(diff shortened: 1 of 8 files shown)


fmrsabino (changes requested):
Looks good.

Although I'm trying to understand exactly how this flow would help us: from what I could gather, every 110 minutes we reindex the last 2 hours (default).

While this is ongoing will the indexing of new blocks stop until this task finishes?

Uxio0:
That's correct, but indexing is never stopped. It's run on parallel

safe_transaction_service/history/management/commands/setup_service.py
```diff
@@ -75,6 +75,12 @@ def create_task(self) -> Tuple[PeriodicTask, bool]:
         5,
         IntervalSchedule.MINUTES,
     ),
+    CeleryTaskConfiguration(
+        "safe_transaction_service.history.tasks.reindex_master_copies_last_hours",
```
fmrsabino, line 79: Should this be `safe_transaction_service.history.tasks.reindex_last_hours`

fmrsabino (changes requested):

safe_transaction_service/history/management/commands/setup_service.py
Uxio0: Sure! Nice catch!

Uxio0: Should be fixed now

safe_transaction_service/history/management/commands/setup_service.py
```diff
@@ -75,6 +75,12 @@ def create_task(self) -> Tuple[PeriodicTask, bool]:
...
+        "Reindex master copies for the last hours",
+        110,
```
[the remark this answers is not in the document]
Uxio0: Yes

fmrsabino (approved):

PR #484, opened 2021-11-12
Add Troubleshooting section to README.md

fmrsabino:
- Add instructions on how to set the environment in case the `grpc` fails to build on an M1 chip.


Uxio0 (approved):
Good idea, thanks!

PR #491, opened 2021-11-16
Only one reindexing time

Uxio0:
### What was wrong?

Reindexing of missing transactions were overloading the service

### How was it fixed?
- Only one reindexing is allowed at a time now. It should be enough for quickly fix all the problematic Safes
- Database event pool is always closed, even if an exception occurs
- `safe_transaction_service.history.tasks.process_decoded_internal_txs_task` will only be called every 20 minutes. The task is already called every time a new trace appears either way and the scheduled task is only to make sure everything was processed, so no need to be biminutely


fmrsabino (approved):
LGTM 🚀

PR #496, opened 2021-11-18
Add endpoints for Ethereum RPC info

Uxio0:
# Endpoints
- `/about/ethereum-rpc/`
- `/about/ethereum-tracing-rpc/`

# Result
```json
{
    "version": "EthereumJS TestRPC/v2.13.2/ethereum-js",
    "block_number": 70,
    "chain_id": 1337,
    "chain": "GANACHE",
    "syncing": false
}
```

### DIFF
--- a/safe_transaction_service/history/tests/test_views.py
+++ b/safe_transaction_service/history/tests/test_views.py
@@ -50,6 +50,26 @@ logger = logging.getLogger(__name__)
 
 
 class TestViews(SafeTestCaseMixin, APITestCase):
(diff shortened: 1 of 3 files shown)


fmrsabino (approved):
LGTM. A question though:

why not have a single `about/ethereum-rpc/` and in the response have something like `tracingEnabled : true | false`?

Uxio0:
Because the transaction service has 2 nodes configured, one for tracing and other normal. You can use the tracing node for both, though

Issue #508, opened 2021-11-23
Optimize database queries


Uxio0:
# What is needed?
- [x] Check atomic blocks and make it as little as possible
- [x] Optimize long time consuming queries
- [x] Analyze if we can apply more caching

Uxio0:
Size of databases before #529 (staging):
- Mainnet: 18 Gb
- Rinkeby: 3.597 Gb
- Polygon: 16 Gb

Size of databases after #529:
- Mainnet:
- Rinkeby: 1.304 Gb
- Polygon:

Uxio0:
> Size of databases before #529 (staging):
> 
> * Mainnet: 18 Gb
> * Rinkeby: 3.597 Gb
> * Polygon: 16 Gb
> 
> Size of databases after #529:
> 
> * Mainnet: 8 Gb
> * Rinkeby: 1.304 Gb
> * Polygon: 13 Gb
> 
> Time spent in migrating production data (with staging database, with way less resources than the production one):
> 
> * Mainnet: 20 minutes
> * Rinkeby: 5 minutes
> * Polygon: 15 minutes

I updated the times @gnosis/safe-services

Uxio0:
I think this should be complete for now, as we don't see any database issues on production since the last optimization update

PR #511, opened 2021-11-24
Denormalize ERC20/721Transfer DB tables

Uxio0:
### What was wrong?

- Joining tables to sort by the `block_number` or `timestamp` was so inefficient
- `block_number` and `timestamp` was added to `ERC20Transfer` and `ERC721Transfer`
- Related to #508


safe_transaction_service/history/services/transaction_service.py
```diff
@@ -227,12 +227,11 @@ def get_all_tx_hashes(
             ERC721Transfer.objects.to_or_from(safe_address)
             .exclude(ethereum_tx__in=multisig_and_module_hashes)
             .annotate(
-                execution_date=F("ethereum_tx__block__timestamp"),
-                created=F("ethereum_tx__block__timestamp"),
-                block=F("ethereum_tx__block_id"),
```
rmeissner, line 232: just to fully understand the old code: would `ethereum_tx__block_id` be the same as `ethereum_tx__block__number`, just with an additional internal db lookup?

rmeissner (approved):

safe_transaction_service/history/services/transaction_service.py
Uxio0: Yes. Well, Django is clever enough to replace `ethereum_tx__block__number` to `ethereum_tx__block_id`

PR #521, opened 2021-12-01
Index blocks using `hash` instead of `number`

### LINKED ISSUE #520: BSC service is out of sync
**Describe the bug**
ERC20 transfers are outdated. Looks like there was a reorg and some blocks were missing

Uxio0:
Closes #520

### What was wrong?
A case was detected in production where the block changed due to a reorg while indexing

### How was it fixed?
Now it that happens block creation will fail


safe_transaction_service/history/services/index_service.py
```diff
@@ -162,31 +162,33 @@ def txs_create_or_update_from_tx_hashes(
...
                 raise BlockNotFoundException(
-                    f"Block with number={block_number} was not found"
+                    f"Block with hash={block_hash} was not found"
                 )
-            assert block_number == block["number"]
-            block_dict[block["number"]] = block
+            assert (
+                block_hash == block["hash"].hex()
+            ), f'{block_hash} - {block["hash"].hex()}'
+            block_dict[block["hash"]] = block
```
jpalvarezl, line 186: Here the logic changed a bit (I mean the substraction). Why this operation necessary with `block_hash` but not `block_number`?
Also, is this `=` assignment atomic? Could this be an issue in a concurrent scenario?

safe_transaction_service/history/services/index_service.py
```diff
@@ -162,31 +162,33 @@ def txs_create_or_update_from_tx_hashes(
...
 
         # Create new transactions or update them if they have no receipt
         current_block_number = self.ethereum_client.current_block_number
         for tx, tx_receipt in zip(txs, tx_receipts):
-            block = block_dict.get(tx["blockNumber"])
+            block = block_dict[tx["blockHash"]]
```
jpalvarezl, line 191: Just curious as if using the array accessor instead of the `get` method brings unexpected behaviour 🤔

safe_transaction_service/history/tests/test_internal_tx_indexer.py
```diff
@@ -53,6 +55,10 @@ def test_internal_tx_indexer_provider(self):
                 (InternalTxIndexer, InternalTxIndexerWithTraceBlock),
             )
 
+    def return_sorted_blocks(self, hashes: HexStr):
+        block_dict = {block["hash"].hex(): block for block in block_result}
+        return [block_dict[provided_hash] for provided_hash in hashes]
```
jpalvarezl, line 60: Not sure how to read this :D but just wondering if there is nothing here that is "lazily" evaluated that could return something unexpected

safe_transaction_service/history/services/index_service.py
Uxio0: Oh, that's a debug line I added. I've just removed it. Yes, it's atomic

safe_transaction_service/history/services/index_service.py
Uxio0: No, it was the opposite. `get` would return a `None` if not found (that should never happen). In Python it´s common to use `[]` accessor when you know the value is there

safe_transaction_service/history/tests/test_internal_tx_indexer.py
Uxio0: This is to return the mock. Previously it was easy, as mock returned the blocks sorted by `block_number`. Now `hashes` are used, so sorting of the blocks can vary

jpalvarezl (approved):

PR #522, opened 2021-12-01
Support Boba Network contracts

InoMurko:
### What was wrong?

Gnosis is not officially supporting Boba Network, but I'd rather keep on deploying your tags rather than maintain our fork. 
Not sure if this is OK or not?
### How was it fixed?

Add Boba deterministic deployments.

### DIFF
--- a/safe_transaction_service/history/management/commands/setup_service.py
+++ b/safe_transaction_service/history/management/commands/setup_service.py
@@ -235,6 +235,14 @@ MASTER_COPIES: Dict[EthereumNetwork, List[Tuple[str, int, str]]] = {
         ("0x03A3cF69a09b5B59CDaa84a5cE3a011AC373541c", 198663, "1.3.0+L2"),
         ("0xFaB2a82fd127590f67EE120626E7A47b67E6912c", 198665, "1.3.0"),
     ],
+    EthereumNetwork.BOBA_RINKEBY: [
+        ("0xfb1bffC9d739B8D520DaF37dF666da4C687191EA", 18854, "1.3.0+L2"),
+        ("0x69f4D1788e39c87893C980c06EdF4b7f686e2938", 18855, "1.3.0"),
+    ],
+    EthereumNetwork.BOBA: [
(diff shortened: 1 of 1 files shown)


config/settings/local.py
```diff
@@ -13,11 +13,8 @@
     default="aHdCBMHXuxIxEhfRGFRp7Cp3N9CqEZEEAvwZVlBCazKExkEnzvVs4bYWC8Qqh9lg",
 )
 # https://docs.djangoproject.com/en/dev/ref/settings/#allowed-hosts
-ALLOWED_HOSTS = [
-    "localhost",
-    "0.0.0.0",
-    "127.0.0.1",
-]
+allowed_hosts = os.getenv("DJANGO_ALLOWED_HOSTS", "localhost,0.0.0.0,127.0.0.1,*")
+ALLOWED_HOSTS = [allowed_host.strip() for allowed_host in allowed_hosts.split(",")]
```
rmeissner, line 17: Could you mention why this change is needed?

InoMurko: Oh sorry, this slipped.
In a docker compose setup with all needed gnosis infrastructure the transaction service would reject requests coming from other containers.
So I've added the default `*` (allow all) and reading from env var `DJANGO_ALLOWED_HOSTS` in case one would like to do something custom. Similar code exists in  https://github.com/gnosis/safe-transaction-service/blob/master/config/settings/production.py#L9

Uxio0: Please remove this change from the PR 🙂

Uxio0 (changes requested):
Could you link the PR for BOBA contracts on https://github.com/gnosis/safe-deployments ?

config/settings/local.py
InoMurko: Can I add it in a separate PR?

InoMurko:
> Could you link the PR for BOBA contracts on https://github.com/gnosis/safe-deployments ?

https://github.com/gnosis/safe-deployments/pull/41

Uxio0:
Please rebase master branch 😉

Uxio0 (approved):

PR #529, opened 2021-12-03
Optimize db fields

Uxio0:
### What was wrong?

- Most records of the database are `EthereumAddresses` and `Keccak256Hashes`, stored as a `varchar`. That's not the optimal type.
- Searching of `Keccak256Hashes` in Django admin were not working
- Related to #508

### How was it fixed?

- Migrate Postgresql fields from `varchar` to `bytea` using manual migrations (by default Django just casts `text` -> `bytea`)
- Fix searching of hashes in Django admin
- Fix validation of ethereum address and hashes on view filters


safe_transaction_service/history/services/transaction_service.py
```diff
@@ -417,5 +417,8 @@ def serialize_all_txs(
                 raise ValueError(f"Type={model_type} not expected, cannot serialize")
             serialized = serializer(model)
             # serialized.is_valid(raise_exception=True)
+            print("a", serializer)
+            print(type(model), model)
             results.append(serialized.data)
+            print("b", serializer)
```
fmrsabino, line 423: Forgotten `print`s ? 😅

safe_transaction_service/history/tests/test_views.py
```diff
@@ -333,8 +333,7 @@ def test_get_module_transactions(self):
             + "?transaction_hash=0x2345"
         )
         response = self.client.get(url, format="json")
-        self.assertEqual(response.status_code, status.HTTP_200_OK)
-        self.assertEqual(response.data["count"], 0)
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
```
fmrsabino, line 336: Why did the result of the test change?

safe_transaction_service/utils/admin.py
```diff
@@ -0,0 +1,35 @@
...
+        self, request: HttpRequest, queryset: QuerySet, search_term: str
+    ) -> Tuple[QuerySet, bool]:
+        queryset, may_have_duplicates = super().get_search_results(
+            request, queryset, search_term
+        )
+        if search_term:
+            for search_field in self.get_search_fields(request):
+                try:
+                    if search_field.startswith("="):
+                        may_have_duplicates = True
```
fmrsabino, line 20: Why do we change the `may_have_duplicates` here and not take the one from `super().get_search_results()`?

fmrsabino (changes requested):

safe_transaction_service/history/services/transaction_service.py
Uxio0: Oh, removed

fmrsabino (approved):

PR #530, opened 2021-12-03
Read ALLOWED_HOSTS from env for local enviroment

InoMurko:
### What was wrong?
If a docker compose setup is used for all needed services supporting Gnosis Safe, the transaction service would reject requests coming from other hosts.
Closes #

### How was it fixed?
Flexible `DJANGO_ALLOWED_HOSTS`
@gnosis/safe-services

### DIFF
--- a/config/settings/local.py
+++ b/config/settings/local.py
@@ -13,7 +13,7 @@ SECRET_KEY = env(
     default="aHdCBMHXuxIxEhfRGFRp7Cp3N9CqEZEEAvwZVlBCazKExkEnzvVs4bYWC8Qqh9lg",
 )
 # https://docs.djangoproject.com/en/dev/ref/settings/#allowed-hosts
-allowed_hosts = env("DJANGO_ALLOWED_HOSTS", default="localhost,0.0.0.0,127.0.0.1,*")
+allowed_hosts = env("DJANGO_ALLOWED_HOSTS", default=".localhost,127.0.0.1,[::1]")
 ALLOWED_HOSTS = [allowed_host.strip() for allowed_host in allowed_hosts.split(",")]
 
 # CACHES
--- a/config/settings/local.py
+++ b/config/settings/local.py
@@ -1,5 +1,5 @@
 import socket
-
+import os
 from .base import *  # noqa
 from .base import env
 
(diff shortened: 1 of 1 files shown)


config/settings/local.py
```diff
@@ -1,5 +1,5 @@
 import socket
-
+import os
```
Uxio0, line 2: Could you use `env` instead of `os`. You can see examples on `base.py`

Uxio0 (changes requested):

config/settings/local.py
```diff
@@ -13,11 +13,8 @@
     default="aHdCBMHXuxIxEhfRGFRp7Cp3N9CqEZEEAvwZVlBCazKExkEnzvVs4bYWC8Qqh9lg",
 )
 # https://docs.djangoproject.com/en/dev/ref/settings/#allowed-hosts
-ALLOWED_HOSTS = [
-    "localhost",
-    "0.0.0.0",
-    "127.0.0.1",
-]
+allowed_hosts = env("DJANGO_ALLOWED_HOSTS", default="localhost,0.0.0.0,127.0.0.1,*")
```
fmrsabino, line 16: The `*` will match anything. This should not be the default if `DJANGO_ALLOWED_HOSTS` is not set. Since this is the `local.py` configuration I'd change it to `.localhost,127.0.0.1,[::1]`

fmrsabino (changes requested):

config/settings/local.py
InoMurko: I agree!

fmrsabino (approved):

config/settings/local.py
```diff
@@ -13,11 +13,8 @@
     default="aHdCBMHXuxIxEhfRGFRp7Cp3N9CqEZEEAvwZVlBCazKExkEnzvVs4bYWC8Qqh9lg",
 )
 # https://docs.djangoproject.com/en/dev/ref/settings/#allowed-hosts
-ALLOWED_HOSTS = [
-    "localhost",
-    "0.0.0.0",
-    "127.0.0.1",
-]
+allowed_hosts = env("DJANGO_ALLOWED_HOSTS", default=".localhost,127.0.0.1,[::1]")
```
Uxio0, line 16: Please use:

`ALLOWED_HOSTS = env("DJANGO_ALLOWED_HOSTS", default=[".localhost", "127.0.0.1", "[::1]"])`

No need for the next line

Uxio0 (changes requested):

config/settings/local.py
Uxio0: Take a look at `production.py` 😉

InoMurko:
https://github.com/gnosis/safe-transaction-service/pull/546

PR #551, opened 2021-12-13
Require 2FA for admin login

### LINKED ISSUE #534: Analyze authentication alternatives for the admin interface (2FA, SSO...)
# What is needed?
We should enforce 2fa for all accounts that can use the admin interface.

Uxio0:
- Closes #534

### DIFF
--- a/config/urls.py
+++ b/config/urls.py
@@ -5,9 +5,14 @@ from django.http import HttpResponse
 from django.urls import path, re_path
 from django.views import defaults as default_views
 
+from django_otp.admin import OTPAdminSite
 from drf_yasg import openapi
 from drf_yasg.views import get_schema_view
 from rest_framework import permissions
+
+if not settings.DEBUG:
+    admin.site.__class__ = OTPAdminSite
+
 
 schema_view = get_schema_view(
     openapi.Info(
--- a/config/urls.py
+++ b/config/urls.py
@@ -10,7 +10,7 @@ from drf_yasg import openapi
 from drf_yasg.views import get_schema_view
 from rest_framework import permissions
 
-if not settings.DEBUG:
+if settings.ENABLE_2FA:
     admin.site.__class__ = OTPAdminSite
 
 
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,4 @@
+[pytest]
+env =
+    DJANGO_SETTINGS_MODULE=config.settings.test
+    DJANGO_DOT_ENV_FILE=.env.test
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,6 +11,7 @@ django-debug-toolbar-force
 django-environ==0.8.1
 django-filter==21.1
 django-model-utils==4.2.0
+django-otp[qrcode]==1.1.3
 django-redis==5.2.0
 django-s3-storage==0.13.5
 django-timezone-field==4.2.1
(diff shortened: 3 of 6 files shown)


config/urls.py
```diff
@@ -5,10 +5,17 @@
...
 
+from django_otp.admin import OTPAdminSite
 from drf_yasg import openapi
 from drf_yasg.views import get_schema_view
 from rest_framework import permissions
 
+if not settings.DEBUG:
+    admin.site.__class__ = OTPAdminSite
+
+admin.site.__class__ = OTPAdminSite
```
fmrsabino, line 16: We are setting the `OTPAdminSite` regardless of `settings.DEBUG`

config/urls.py
```diff
@@ -5,10 +5,17 @@
 from django.urls import path, re_path
 from django.views import defaults as default_views
 
+from django_otp.admin import OTPAdminSite
 from drf_yasg import openapi
 from drf_yasg.views import get_schema_view
 from rest_framework import permissions
 
+if not settings.DEBUG:
+    admin.site.__class__ = OTPAdminSite
```
fmrsabino, line 14: Should we use the `AdminConfig` for this?

example:

```python
from django.contrib.admin.apps import AdminConfig

class OTPAdminConfig(AdminConfig):
    default_site = 'django_otp.admin.OTPAdminSite'
```

and then install the `OTPAdminConfig`:
```python
INSTALLED_APPS = [
    ...
    'safetransactionservice.admin.OTPAdminConfig',  # replaces 'django.contrib.admin'
    ...
]
```

See https://docs.djangoproject.com/en/dev/ref/contrib/admin/#overriding-the-default-admin-site

pytest.ini
```diff
@@ -0,0 +1,4 @@
+[pytest]
+env =
+    DJANGO_SETTINGS_MODULE=config.settings.test
+    DJANGO_DOT_ENV_FILE=.env.test
```
fmrsabino, line 4: Is this required?

requirements.txt
```diff
@@ -27,6 +28,7 @@ packaging>=21.0
 pillow==8.4.0
 psycogreen==1.0.2
 psycopg2-binary==2.9.2
+qrcode==7.3.1
```
fmrsabino, line 31: I think we can merge this with the `django-otp` by using `django-otp[qrcode]=1.1.3`

fmrsabino (changes requested):

config/urls.py
Uxio0: I did it for testing and leave it there 🤦🏻‍♂️

config/urls.py
Uxio0: Didn't know about that way to do that... What way do you think is more explicit?

pytest.ini
Uxio0: No, I realised tests where not picking the environment variables and got into this PR

requirements.txt
Uxio0: Good proposal

config/urls.py
fmrsabino: I would say that by using `INSTALLED_APPS` we make it more configurable also in the future in case we want to deploy an instance without the OTP (might be the case also for other teams doing their own deployments).

I would not add right now an environment variable for it but by having it in `INSTALLED_APPS` seems to be a better approach regarding general configuration.

Uxio0: Ok, I did it that way 👍🏻

Uxio0:
Replaced by https://github.com/gnosis/safe-transaction-service/pull/753

PR #553, opened 2021-12-14
Add xDai Master Copy and Proxy Factory v1.0.0

Uxio0:


fmrsabino (approved):
Looks good.

However we are missing the deployments for xDAI for `v1.0.0` on https://github.com/gnosis/safe-deployments 🤔

PR #557, opened 2021-12-15
Fix token migration

Uxio0:
### What was wrong?

Token address migration was only casting `varchar` instead of converting to `bytea`

### DIFF
--- a/safe_transaction_service/tokens/migrations/0004_ethereum_address_field_v2_20211201_1512.py
+++ b/safe_transaction_service/tokens/migrations/0004_ethereum_address_field_v2_20211201_1512.py
@@ -13,7 +13,10 @@ class Migration(migrations.Migration):
 
     operations = [
         migrations.RunSQL(
-            "DROP INDEX IF EXISTS tokens_token_address_18ef94ca_like",
+            """
+            DROP INDEX IF EXISTS tokens_token_address_18ef94ca_like;
+            ALTER TABLE "tokens_token" ALTER COLUMN "address" TYPE bytea USING DECODE(SUBSTRING("address", 3), 'hex');
+            """,
             reverse_sql=migrations.RunSQL.noop,
         ),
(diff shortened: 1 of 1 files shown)


safe_transaction_service/tokens/migrations/0004_ethereum_address_field_v2_20211201_1512.py
```diff
@@ -13,7 +13,10 @@ class Migration(migrations.Migration):
 
     operations = [
         migrations.RunSQL(
-            "DROP INDEX IF EXISTS tokens_token_address_18ef94ca_like",
+            """
+            DROP INDEX IF EXISTS tokens_token_address_18ef94ca_like;
+            ALTER TABLE "tokens_token" ALTER COLUMN "address" TYPE bytea USING DECODE(SUBSTRING("address", 3), 'hex');
```
rmeissner, line 18: :thinking: what is this doing?

rmeissner: why substring 3?

Uxio0: Currently we are storing addresses as `varchar` with `0x` prefix. `0x` prefix is not valid for the postgres `DECODE` function. So this line is converting current `varchar` addresses to `bytea` binary ones

rmeissner: Apparently SQL uses 1 based indexes .... this is what confused me: https://www.w3schools.com/sql/trymysql.asp?filename=trysql_func_mysql_substring

image

rmeissner (approved):

Uxio0:
> Apparently SQL uses 1 based indexes

Good old Pascal

PR #558, opened 2021-12-15
Fix celery concurrency

Uxio0:
### What was wrong?

- Concurrency was too low (`120`) to process all the needed tasks. A higher number can be used as we are using `gevent` (light threads)

### How was it fixed?

- Make concurrency configurable so we can adjust it without a deployment
- Set `500` as a default for concurrency (x5)
- Increase `CELERY_BROKER_POOL_LIMIT` as stated here: https://docs.celeryproject.org/en/stable/userguide/optimizing.html#broker-connection-pools

@gnosis/safe-services

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -206,6 +206,10 @@ INSTALLED_APPS += [
 ]
 # http://docs.celeryproject.org/en/latest/userguide/configuration.html#std:setting-broker_url
 CELERY_BROKER_URL = env("CELERY_BROKER_URL", default="django://")
+# https://docs.celeryproject.org/en/stable/userguide/optimizing.html#broker-connection-pools
+CELERY_BROKER_POOL_LIMIT = env(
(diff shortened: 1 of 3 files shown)


config/settings/base.py
```diff
@@ -206,6 +206,10 @@
 ]
 # http://docs.celeryproject.org/en/latest/userguide/configuration.html#std:setting-broker_url
 CELERY_BROKER_URL = env("CELERY_BROKER_URL", default="django://")
+# https://docs.celeryproject.org/en/stable/userguide/optimizing.html#broker-connection-pools
+CELERY_BROKER_POOL_LIMIT = env(
+    "CELERY_BROKER_POOL_LIMIT", default=env("CELERYD_CONCURRENCY", default=500)
+)
```
fmrsabino, line 212: How is this used in the application?

fmrsabino (approved):

config/settings/base.py
Uxio0: It's used on the client side, as the other `CELERY_BROKER` configuration

PR #559, opened 2021-12-15
Refactor celery configuration

Uxio0:
@fmrsabino 😉

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -204,12 +204,13 @@ MANAGERS = ADMINS
 INSTALLED_APPS += [
     "django_celery_beat",
 ]
+
+CELERYD_CONCURRENCY = env("CELERYD_CONCURRENCY", default=500)
+CELERYD_POOL = env("CELERYD_POOL", default="gevent")
 # http://docs.celeryproject.org/en/latest/userguide/configuration.html#std:setting-broker_url
 CELERY_BROKER_URL = env("CELERY_BROKER_URL", default="django://")
 # https://docs.celeryproject.org/en/stable/userguide/optimizing.html#broker-connection-pools
-CELERY_BROKER_POOL_LIMIT = env(
-    "CELERY_BROKER_POOL_LIMIT", default=env("CELERYD_CONCURRENCY", default=500)
-)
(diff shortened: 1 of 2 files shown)


config/settings/base.py
```diff
@@ -204,12 +204,13 @@
...
+
+CELERYD_CONCURRENCY = env("CELERYD_CONCURRENCY", default=500)
+CELERYD_POOL = env("CELERYD_POOL", default="gevent")
 # http://docs.celeryproject.org/en/latest/userguide/configuration.html#std:setting-broker_url
 CELERY_BROKER_URL = env("CELERY_BROKER_URL", default="django://")
 # https://docs.celeryproject.org/en/stable/userguide/optimizing.html#broker-connection-pools
-CELERY_BROKER_POOL_LIMIT = env(
-    "CELERY_BROKER_POOL_LIMIT", default=env("CELERYD_CONCURRENCY", default=500)
-)
+CELERY_BROKER_POOL_LIMIT = env("CELERY_BROKER_POOL_LIMIT", default=CELERYD_CONCURRENCY)
```
fmrsabino, line 213: Should this be `BROKER_POOL_LIMIT`? https://docs.celeryproject.org/en/stable/userguide/configuration.html?highlight=CELERYD_CONCURRENCY#broker-pool-limit

config/settings/base.py
```diff
@@ -204,12 +204,13 @@
 INSTALLED_APPS += [
     "django_celery_beat",
 ]
+
+CELERYD_CONCURRENCY = env("CELERYD_CONCURRENCY", default=500)
```
fmrsabino, line 208: For this one, shouldn't it match the number of CPU cores available (default) on the host? Wondering how `500` would affect the number of worker processes spawn per host 🤔

config/settings/base.py
Uxio0: Celery on Python uses the `CELERY` namespace for every configuration variable

config/settings/base.py
Uxio0: But it was a good catch 😉

fmrsabino (approved):

PR #560, opened 2021-12-15
Updated issue templates

jpalvarezl:
### What was wrong?

Included more explicit requests of context for features and bugs
Closes #(there is no associated issue in this repo)

### How was it fixed?
Added similar instructions to what's available in the `safe-config-service` and the `safe-client-gateway`

@gnosis/safe-services


Uxio0 (approved):

Uxio0:
Thanks @jpalvarezl !

PR #561, opened 2021-12-15
Fix migration

Uxio0:
- Using COMMIT breaks Django migration system


safe_transaction_service/history/migrations/0050_ethereum_address_field_v2_20211201_1507.py
```diff
@@ -71,7 +71,6 @@ class Migration(migrations.Migration):
             ALTER TABLE "history_multisigtransaction" ALTER COLUMN "safe" TYPE bytea USING DECODE(SUBSTRING("safe", 3), 'hex');
             ALTER TABLE "history_multisigtransaction" ALTER COLUMN "to" TYPE bytea USING DECODE(SUBSTRING("to", 3), 'hex');
             ALTER TABLE "history_proxyfactory" ALTER COLUMN "address" TYPE bytea USING DECODE(SUBSTRING("address", 3), 'hex');
-            DROP INDEX IF EXISTS "history_safecontract_address_b3109ec5_like";
```
jpalvarezl, line 74: How come this is no longer needed?

jpalvarezl (approved):
Would approve to not block, but just left a single question :)

safe_transaction_service/history/migrations/0050_ethereum_address_field_v2_20211201_1507.py
Uxio0: Because it's duplicated, it's already up

Uxio0: https://github.com/gnosis/safe-transaction-service/pull/561/files#diff-961fa78220cce1b53909bd592793e8df50ab63078c837a8da4bbdef4ca81c73cL40

PR #570, opened 2021-12-20
Fix native coin price on token fiat endpoint

Uxio0:
- When providing `0x00.00` on `tokens/<address>/prices/usd/ fiat value for the native coin was not retrieved


rmeissner (approved):

Uxio0:
I added more tests

PR #583, opened 2021-12-23
Use psycopg2 instead of psycopg2-binary

Uxio0:
https://www.psycopg.org/docs/install.html#psycopg-vs-psycopg-binary


fmrsabino (approved):
🚀 

I also need to do this on the safe-config-service side

PR #585, opened 2021-12-23
Increase number of DB `MAX_CONNS` from 4 -> 50

Uxio0:
### What was wrong?

- There were deadlocks on DB access (detected on production xDai)
- Number of max connections to the database was reduced to `4` when there were performance issues on DB level, so DB woulnd't get overloaded. Now we can safetely increase it again.

### How was it fixed?

- Increase number of DB `MAX_CONNS` from 4 -> 50
- A task was not closing DB connections, it was fixed
- `MAX_CONNS` and `REUSE_CONNS` are now configurable via environment var


rmeissner (approved):

Uxio0:
@luarx Do you see any issues? We will increase number of Database connections from about 20 to 200

luarx:
It shoud not be a problem! Let's try in staging before

PR #595, opened 2021-12-25
Make sure tasks close postgres connections

Uxio0:
### What was wrong?

It looks like sometimes postgres connections can stay open

### How was it fixed?

Close postgres connections in every task


rmeissner (approved):
I checked that the new decorator was used in all the touched methods.

PR #622, opened 2022-01-10
Disable authentication in OwnersView

fmrsabino:
- `OwnersView` no longer requires authentication
- Tokens can still be created as before (the API token feature was not removed)

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -420,3 +420,9 @@ AWS_CONFIGURED = bool(
 
 ETHERSCAN_API_KEY = env("ETHERSCAN_API_KEY", default=None)
 IPFS_GATEWAY = env("IPFS_GATEWAY", default="https://cloudflare-ipfs.com/")
+
+SWAGGER_SETTINGS = {
+    "SECURITY_DEFINITIONS": {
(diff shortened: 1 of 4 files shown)


config/settings/base.py
```diff
@@ -416,9 +416,3 @@
 
 ETHERSCAN_API_KEY = env("ETHERSCAN_API_KEY", default=None)
 IPFS_GATEWAY = env("IPFS_GATEWAY", default="https://cloudflare-ipfs.com/")
-
-SWAGGER_SETTINGS = {
```
Uxio0, line 420: Maybe we can leave this? It doesn't harm to have it, right?

Uxio0 (approved):

config/settings/base.py
fmrsabino: Yes makes sense. I don't think this was being used before on our Swagger.

fmrsabino: ae8b48a

Uxio0: Looks good to me

Uxio0:
It seems that there's an issue with the tests

fmrsabino:
> It seems that there's an issue with the tests

The PR reverted some adjustments made on `test_transfers_view`. No I reverted it correctly.

Issue #620, opened 2022-01-10
Separate the celery tasks into different queues that should have different workers


fmrsabino:
@rmeissner will fill here the details

Uxio0:
As a lot of pending tasks were related to webhook http requests, I increased the number of concurrent connections https://github.com/gnosis/safe-transaction-service/pull/625 (previously we had 10).

Also, priority for notifications/webhooks and contracts indexing tasks was lowered (this change is already on production as it was included on `v4.0.0`: https://github.com/gnosis/safe-transaction-service/pull/623

I will keep monitoring production task queue and investigate on what would be the best solution to have it monitorized: https://github.com/danihodovic/celery-exporter/blob/master/images/grafana_dashboard.png

PR #623, opened 2022-01-10
Set low priority for webhook/notification tasks

Uxio0:
Regular priority for Celery is 3, we are setting 1 for these tasks (the lowest the less priority)

This is a first fix for #620

### DIFF
--- a/safe_transaction_service/history/signals.py
+++ b/safe_transaction_service/history/signals.py
@@ -247,9 +247,13 @@ def process_webhook(
     )
     for payload in payloads:
         if address := payload.get("address"):
-            send_webhook_task.delay(address, payload)
+            send_webhook_task.apply_async(
+                args=(address, payload), priority=4
+            )  # Almost the lowest priority
             if is_relevant_notification(sender, instance, created):
-                send_notification_task.apply_async(args=(address, payload), countdown=5)
+                send_notification_task.apply_async(
+                    args=(address, payload), countdown=5, priority=4
+                )
             else:
                 logger.debug(
                     "Notification will not be sent for created=%s object=%s",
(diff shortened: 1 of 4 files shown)


safe_transaction_service/history/signals.py
```diff
@@ -247,9 +247,13 @@ def process_webhook(
...
     for payload in payloads:
         if address := payload.get("address"):
-            send_webhook_task.delay(address, payload)
+            send_webhook_task.apply_async(
+                args=(address, payload), priority=1
+            )  # Almost the lowest priority
             if is_relevant_notification(sender, instance, created):
-                send_notification_task.apply_async(args=(address, payload), countdown=5)
+                send_notification_task.apply_async(
+                    args=(address, payload), countdown=5, priority=1
```
fmrsabino, line 255: Wouldn't we need to configure `app.conf.broker_transport_options` to make Redis aware of the priority queues?

> To start scheduling tasks based on priorities you need to configure queue_order_strategy transport option.

https://docs.celeryproject.org/en/stable/userguide/routing.html#redis-message-priorities

Uxio0: We already do, take a look at `base.py` line 234

fmrsabino, line 255: Isn't the highest priority `0`? If we only use `0` so far then I think `1` makes sense (or If it needs to be a higher number ie.: lower priority)

Uxio0: It looks like for Redis is the opposite 🤦🏻 . Thanks for noticing @fmrsabino, I have changed it

fmrsabino (approved):

PR #627, opened 2022-01-11
Fix parsing when Token name/symbol are bytes

### LINKED ISSUE #626: 500 error for getting balances
**User support issue:**
https://discord.com/channels/907968493134155807/913377923308871680/927209590888759296

**Describe the bug**
500 error returned for the safe https://gnosis-safe.io/app/eth:0x921DCc4ba6872B4736eB1A22351E2a8Ce671C3E2/balances
image

fmrsabino:
### What was wrong?

Closes #626 

### How was it fixed?

- `name` and `symbol` can either be a `str` or `bytes`. If it happens to be `bytes` (as it happens with the the token `0xBB9bc244D798123fDe783fCc1C72d3Bb8C189413`) then the trimming of the `name` and/or `symbol` will not work and a DB constraint exception is thrown – `django.db.utils.DataError: value too long for type character varying(60)`
- Additionally, if `bytes` is correctly converted into a `str`, `bytes` might contain `NULL` characters (`\x00`) which are not valid for under DB constraints (for `name` and `symbol`)
- Therefore `NULL` characters are replaced by �. More specifically for the token `0xBB9bc244D798123fDe783fCc1C72d3Bb8C189413`:

```
name: b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01'
symbol: b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01'
```

would be saved as 

```
name: �������������������������������\x01
symbol: �������������������������������\x01
```

### DIFF
--- a/safe_transaction_service/tokens/tests/test_models.py
+++ b/safe_transaction_service/tokens/tests/test_models.py
@@ -126,3 +126,24 @@ class TestModels(TestCase):
         self.assertEqual(
             token.name, "Balancer Pool Token " + get_metadata_mock.return_value.name
         )
+
+    from gnosis.eth.ethereum_client import Erc20Info, Erc20Manager
+
+    @mock.patch.object(
+        Erc20Manager,
+        "get_info",
+        autospec=True,
+        return_value=Erc20Info(
+            name=b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01",
+            symbol=b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01",
+            decimals=18,
+        ),
+    )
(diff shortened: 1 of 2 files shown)


Uxio0 (approved):

Uxio0 (changes requested):
Test? 😉

safe_transaction_service/tokens/tests/test_models.py
```diff
@@ -126,3 +126,24 @@ def test_fix_balancer_pool_tokens(self, get_metadata_mock: MagicMock):
         self.assertEqual(
             token.name, "Balancer Pool Token " + get_metadata_mock.return_value.name
         )
+
+    from gnosis.eth.ethereum_client import Erc20Info, Erc20Manager
```
Uxio0, line 130: Why do you do the import here?

fmrsabino: My bad. Editor's view and I thought that with the commit hooks it'd put them in the correct place 😬

fmrsabino: 919bfdc

Uxio0 (approved):

PR #637, opened 2022-01-18
Fix ERC721 token uris error

Uxio0:
### What was wrong?

- For some ERC721 tokens there's an error retrieving the uri that makes the service crash

### How was it fixed?

- Updating gnosis-py to the last version


fmrsabino (approved):
🚀

Issue #434, opened 2021-10-20
Metrics about the queries to the ethereum nodes


Uxio0:
Ideally, we would like to get some information about the RPC queries done to the ethereum nodes. Ideally:
- Request body
- Response body/errors
- Elapsed time

That way we could debug a lot further issues with the nodes

Uxio0:
This is still pending (cc @luarx 😉 )

Uxio0:
This should do it for now. https://github.com/gnosis/safe-transaction-service/pull/642

PR #643, opened 2022-01-19
Log missing RPC calls

Uxio0:
- Related to #434


Uxio0:
Version was already updated in other PR

PR #653, opened 2022-01-25
Refactor tx decoder

### LINKED ISSUE #638: Data is decoded using a wrong ABI
**Describe the bug**
Data is decoded using a wrong ABI

**To Reproduce**
1. Use xDai omnibridge setting a receiver
2. Check the executed transaction. Parameters have wrong names

**Expected behavior**
Data is decoded using the proper ABI

**Additional context**
- Currently there's a big database with all the ABI and methods are decoded using the most important ABIs, but by default all ABIs have a relevance of 100, so if there's a collision a random ABI will be used
- Checking the contract when decoding will be an overhead (multisend decoding and then getting the ABIs for the contracts...)

**Open questions**
- What happens to the proxies?

Uxio0:
### What was wrong?

- When there's an ABI function selector colliding, sometimes a wrong decoder is used
- Closes #638

### How was it fixed?

Refactor `SafeTxDecoder`:
- Add a method `get_abi_function` to get the proper abi for data and an optional contract `address`.
- Pass always the `contract address` with the `data` to be decoded

Refactor `DbTxDecoder`:
- If the contract address has an ABI on database and can decode the `data` it will be used first 

Add `to` for the data decoding endpoint

### DIFF
--- a/safe_transaction_service/contracts/tx_decoder.py
+++ b/safe_transaction_service/contracts/tx_decoder.py
@@ -1,4 +1,5 @@
 import itertools
+import operator
 from functools import cache, cached_property
 from logging import getLogger
 from typing import (
@@ -16,6 +17,7 @@ from typing import (
 )
 
 import gevent
+from cachetools import TTLCache, cachedmethod
 from eth_abi.exceptions import DecodingError
 from eth_typing import ChecksumAddress, HexStr
 from eth_utils import function_abi_to_4byte_selector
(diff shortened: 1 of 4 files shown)


safe_transaction_service/contracts/tx_decoder.py
```diff
@@ -452,11 +527,49 @@ class DbTxDecoder(TxDecoder):
...
+        :param data: transaction data
+        :param address: contract address in case of ABI colliding
+        :return: Abi function for data if it can be decoded, `None` if not found
+        """
+        selector = data[:4]
+        # Check first that selector is supported on our database
+        if selector in self.fn_selectors_with_abis:
+            # Try to use specific ABI if address provided
+            if address:
+                contract_selectors_with_abis = self.get_contract_abi(address)
```
rmeissner, line 559: Maybe a small comment here: `If the selector is available in the abi specific for the address we will use that one, otherwise we fallback to the general abi that matches the selector`

rmeissner (approved):
As far as I can tell this makes sense. It would be helpful to add to the description the changes made to the `DbTxDecoder` as they are the important part of the PR as far as I understood

safe_transaction_service/contracts/tx_decoder.py
Uxio0: Agree

Uxio0:
@rmeissner Added `to` for the data decoding endpoint

rmeissner (approved):

PR #676, opened 2022-02-09
Fix S3 uploads timing out

Uxio0:
- There was an issue with gevent due to the threading used on boto3 library. More info https://github.com/etianen/django-s3-storage/pull/136


requirements.txt
```diff
@@ -12,14 +13,15 @@ django-environ==0.8.1
...
-django-s3-storage==0.13.5
+# django-s3-storage==0.13.6
 django-timezone-field==4.2.3
 djangorestframework==3.13.1
 djangorestframework-camel-case==1.3.0
 docutils==0.18.1
 drf-yasg[validation]==1.20.0
 ethereum==2.3.2
 firebase-admin==5.2.0
+git+https://github.com/Uxio0/django-s3-storage
```
fmrsabino, line 24: I think we can pin the commit here.

`git+https://github.com/Uxio0/django-s3-storage.git@3a1a8fb33c3aecee8af4f40fa35dca44e5bca5b4`

What do you think?

Uxio0: Done 👍🏻

requirements.txt
```diff
@@ -12,14 +13,15 @@ django-environ==0.8.1
 django-filter==21.1
 django-model-utils==4.2.0
 django-redis==5.2.0
-django-s3-storage==0.13.5
+# django-s3-storage==0.13.6
```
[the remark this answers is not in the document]
Uxio0: Good suggestion

PR #677, opened 2022-02-09
Invalidate contracts cache when contract/ABI is updated

### LINKED ISSUE #665: Add option to invalidate ContractView cache
# What is needed?
When a `Contract` is updated, there's a view level cache set for 1 hour so we can only see the updated changes after the view cache is outdated.

# Background
It should be possible to validate this endpoint either manually or once a `Contract` is updated via a Django signal.

# Endpoint
`GET /api/v1/contracts/<address>`

Uxio0:
Closes #665


fmrsabino (approved):
🚀

PR #682, opened 2022-02-14
Bump django-s3-storage to 0.13.6

Uxio0:


requirements.txt
```diff
@@ -13,15 +13,14 @@ django-environ==0.8.1
 django-filter==21.1
 django-model-utils==4.2.0
 django-redis==5.2.0
-# django-s3-storage==0.13.6  # https://github.com/etianen/django-s3-storage/pull/136
+django-s3-storage==0.13.6
```
fmrsabino, line 16: Had to check why we were not updating the version but it seems that we updated it before the release? 😅

fmrsabino (approved):

PR #685, opened 2022-02-14
Bump gnosis-py to v3.8.0

Uxio0:
- Remove `pyethereum` dependency, as it's not needed anymore

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -19,9 +19,8 @@ djangorestframework==3.13.1
 djangorestframework-camel-case==1.3.0
 docutils==0.18.1
 drf-yasg[validation]==1.20.0
-ethereum==2.3.2
 firebase-admin==5.2.0
-gnosis-py[django]==3.7.8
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -19,9 +19,8 @@ djangorestframework==3.13.1
 djangorestframework-camel-case==1.3.0
 docutils==0.18.1
 drf-yasg[validation]==1.20.0
-ethereum==2.3.2
```
fmrsabino, line 22: Why was this one removed?

Uxio0: It was not being used, and also conflicting with the newest version of py-evm

fmrsabino: For context I think it's better to leave a comment about these in the PR or in the commit message.

fmrsabino (approved):

requirements.txt
Uxio0: Makes sense, thanks

PR #688, opened 2022-02-15
Allow tokens to mimic price from other tokens

### LINKED ISSUE #681: Get token price using another token as reference
# What is needed?
Get token price using another token as reference

# Background
Some tokens price cannot be retrieved, but they have the same virtual price as other tokens (for example `lGNO` and `GNO`)

# Test case
lGNO `0x4f8AD938eBA0CD19155a835f617317a6E788c868` should be equal to `GNO` price

Uxio0:
`Token` model has now a `copy_price` nullable field. If it is set to an `ethereum address`, that `ethereum address` will be used when fetching the token price

Closes #681


safe_transaction_service/tokens/migrations/0006_auto_20220214_1629.py
```diff
@@ -0,0 +1,54 @@
...
+                help_text="If provided, copy the price from the token", null=True
+            ),
+        ),
+        migrations.AlterField(
+            model_name="token",
+            name="decimals",
+            field=models.PositiveSmallIntegerField(
+                blank=True,
+                db_index=True,
+                help_text="Number of decimals. For ERC721 tokens decimals must be `None`",
```
rmeissner, line 28: Help text changes trigger db migrations, interesting

rmeissner (approved):

safe_transaction_service/tokens/migrations/0006_auto_20220214_1629.py
Uxio0: Yes, I don't like that...

PR #689, opened 2022-02-15
Fix contract caching error

Uxio0:
- When redis cache backend is used response should be returned after calling `add_post_render_callback`. [More info](https://docs.djangoproject.com/en/4.0/ref/template-response/#django.template.response.SimpleTemplateResponse.add_post_render_callback)

### DIFF
--- a/safe_transaction_service/__init__.py
+++ b/safe_transaction_service/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "4.2.1"
+__version__ = "4.2.2"
 __version_info__ = tuple(
     [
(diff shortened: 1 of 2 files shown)


rmeissner (approved):
From the linked docs and after looking up the cache implementation I would have expected to be the default behaviour, but not sure :D

Uxio0:
> From the linked docs and after looking up the cache implementation I would have expected to be the default behaviour, but not sure :D

I would too... but when setting up the redis backend, it breaks. I need to change the tests to use redis, currently caching is disabled for tests as it breaks a lot of stuff

PR #691, opened 2022-02-15
Add Vscode default configuration

Uxio0:
- Running/debugging tests should be working
- `.vscode` folder is excluded on `.gitignore` as this is the basic template but it needs to be modified by the user, e.g. adding information about Python environment. That will be done by `vscode` run running it for the first time

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -124,3 +124,4 @@ safe_transaction_service/media
 
 # User-specific stuff:
 .idea/
+.vscode/
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,17 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Django: Run tests",
+            "type": "python",
+            "request": "launch",
+            "module": "pytest",
+            "args": [
+                "--lf",
+                "-rxXs"
+            ],
+            "django": true,
+            "envFile": ".env.test"
+        }
+    ]
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,8 @@
+{
+    "python.formatting.provider": "black",
+    "python.linting.flake8Enabled": true,
+    "python.testing.pytestArgs": [],
+    "python.testing.pytestEnabled": true,
+    "python.testing.unittestEnabled": false,
+    "python.linting.enabled": true
+}


.gitignore
```diff
@@ -124,3 +124,4 @@ safe_transaction_service/media
 
 # User-specific stuff:
 .idea/
+.vscode/
```
fmrsabino, line 127: I think we can add `launch.json` and `settings.json` as part of the files to be tracked (else any update to those files will be ignored from now on)

```suggestion
.vscode/
!.vscode/launch.json
!.vscode/settings.json
```

fmrsabino (changes requested):

.gitignore
Uxio0: That's the idea. When you configure vscode for yourself you will get into it some variables only meaninful to you (like the path for the virtualenvs or some other configurations). I don't want to be ignoring `.vscode` folder in every commit I do. This way if someone wants to update `.vscode` folders it can be done, but it needs to be added using `git add -f`

Uxio0: Updated the description

fmrsabino: But this doesn't affect that other paths under `.vscode` (`.vscode` continues to be excluded). It just says that any updates to `launch.json` and `settings.json` won't be ignored or do we want them to be ignored?

.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Django: Run tests",
+            "type": "python",
+            "request": "launch",
+            "module": "pytest",
+            "args": [
+                "--lf", // Rerun only failed tests
```
fmrsabino, line 10: Doesn't this comment invalidate the json?

Uxio0: Well, it's working, but you are right

.gitignore
Uxio0: We want them to be ignored. All the user changes will be on `settings.json` and `launch.json`, not on different files

.vscode/launch.json
Uxio0: I've removed it

Uxio0: Noooooooo

Uxio0: Now it is

fmrsabino (approved):
🚀

PR #694, opened 2022-02-16
Auto resize token/contract logos to 256x256

### LINKED ISSUE #693: Resize to images to 512x512 maximum and convert to PNG
# What is needed?
To prevent UI issues, token and contracts logos should be resized to 512x512 and converted to PNG

Uxio0:
- Closes #693


Uxio0:
@fmrsabino Maybe we can do this for the `config-service` too

jpalvarezl (approved):

PR #698, opened 2022-02-17
Be less restrictive when setting service out of sync

Uxio0:
- If less than 10% of contracts are out of sync for ERC20/ERC721, consider service synced
- That will prevent a lot of alerts when only one new or two contracts are out of sync

### DIFF
--- a/safe_transaction_service/history/services/index_service.py
+++ b/safe_transaction_service/history/services/index_service.py
@@ -48,6 +48,7 @@ class IndexServiceProvider:
                 EthereumClientProvider(),
                 settings.ETH_REORG_BLOCKS,
                 settings.ETH_L2_NETWORK,
+                settings.ALERT_OUT_OF_SYNC_EVENTS_THRESHOLD,
             )
         return cls.instance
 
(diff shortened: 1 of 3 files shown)


safe_transaction_service/history/services/index_service.py
```diff
@@ -104,10 +104,15 @@ def is_service_synced(self) -> bool:
...
         ).count()
         if out_of_sync_contracts > 0:
-            logger.error(
-                "%d Safe Contracts have ERC20/721 out of sync", out_of_sync_contracts
-            )
-            synced = False
+            total_number_of_contracts = SafeContract.objects.all().count()
+            proportion_out_of_sync = out_of_sync_contracts / total_number_of_contracts
+            # Ignore less than 10% of contracts out of sync
+            if proportion_out_of_sync >= 0.1:
```
fmrsabino, line 110: I think we can have this as an environment variable what do you think?

fmrsabino (commented):
In the future this can be an idea for a Prometheus metric: `SafeContract.objects.all().count()`
And then we do this computation on the alert system side.

safe_transaction_service/history/services/index_service.py
Uxio0: Makes sense

PR #700, opened 2022-02-18
Redact token value from hook logs

fmrsabino:
- We are using `full_url` if the Webhook called failed with an error status – we should use `base_url` so that the token is not logged


Uxio0:
Even if I feel that's the cleanest solution:
- I think doing a regex for *every* log will kill performance (complexity for regex is O(n), but can require up to O(2^m))

Also, we should always use `re.compile` when using regular expressions 😉

Uxio0 (approved):

PR #709, opened 2022-02-21
Bump Django to v4

Uxio0:
- Bump django from 3.2.11 to 4.0.2
- Remove not needed django-redis


fmrsabino (changes requested):
Nice 👏 

We can also remove `django-redis` right?

Uxio0:
> Nice 👏
> 
> We can also remove `django-redis` right?

Yes, I don't know why it wasn't saved 😕 . Updated now

fmrsabino (approved):

PR #710, opened 2022-02-22
Update Django coverage configuration

Uxio0:
- Refactor coverage configuration, remove `.coveragerc` file and move configuration to `setup.cfg`


requirements-test.txt
```diff
@@ -1,5 +1,6 @@
 -r requirements.txt
 coverage==6.3.2
+django-coverage-plugin==2.0.2
```
fmrsabino, line 3: It seems that it is mostly related with Django templates but where we use those in the `safe-transaction-service`?

Uxio0: I will use it either way to update configuration

Uxio0:
Updated only changing the configuration and removing the unecessary file .coveragerc

fmrsabino (approved):

PR #714, opened 2022-02-25
Add prometheus metrics endpoint

Uxio0:
WIP about adding a prometheus exporter to the transaction service.

https://github.com/korfuri/django-prometheus looked very promising, but it does a lot under the hood and I was afraid about having a lot of code executed in every request. I prefer to go implementing step by step what we need, so this PR uses `prometheus_client` library.

# Still pending

- Add authentication for the `metrics/` endpoint


Uxio0:
Datadog is covering this for us now

PR #720, opened 2022-03-02
Update checkout and setup-python github actions to v3

Uxio0:


.github/workflows/python.yml
```diff
@@ -33,18 +33,13 @@ jobs:
...
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v3
       with:
         python-version: ${{ matrix.python-version }}
-    - uses: actions/cache@v2
-      name: Cache pip
-      with:
-        path: ~/.cache/pip
-        key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements-test.txt') }}-${{ hashFiles('**/requirements.txt') }}
-        restore-keys: |
```
jpalvarezl, line 46: Is caching now handled for you in the new version of `setup-python`? You no longer need to restore manually?

jpalvarezl (approved):

.github/workflows/python.yml
Uxio0: That's right

.github/workflows/python.yml
```diff
@@ -33,18 +33,13 @@ jobs:
...
-          ${{ runner.os }}-pip-
+        cache: 'pip'
```
Uxio0, line 41: @jpalvarezl Caching is handled here

PR #726, opened 2022-03-07
Fix creating token with null bytes on name/symbol

Uxio0:
- Fix creating token with null bytes on name/symbol
- Set version 4.2.6

### DIFF
--- a/safe_transaction_service/tokens/tests/test_models.py
+++ b/safe_transaction_service/tokens/tests/test_models.py
@@ -13,7 +13,8 @@ from ..clients.zerion_client import (
     ZerionPoolMetadata,
     ZerionUniswapV2TokenAdapterClient,
 )
-from ..models import Token
+from ..models import Token, TokenManager
+from ..models import logger as token_model_logger
 from .factories import TokenFactory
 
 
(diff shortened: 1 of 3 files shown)


jpalvarezl (approved):
Looks good 🙂

safe_transaction_service/tokens/tests/test_models.py
```diff
@@ -157,3 +157,22 @@ def test_create_from_blockchain_with_byte_data(self, get_info: MagicMock):
         token = Token.objects.get(address="0xBB9bc244D798123fDe783fCc1C72d3Bb8C189413")
         self.assertEqual(token.name, "�������������������������������\x01")
         self.assertEqual(token.symbol, "�������������������������������\x01")
+
+    @mock.patch.object(
+        Erc20Manager,
+        "get_info",
+        autospec=True,
+        return_value=Erc20Info(
+            name="Build\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00",
```
rmeissner, line 166: if this should be handled like a c-type memory for strings, then 0-bytes terminate the string. Therefore I would not replace them but look for the first 0-byte and cut of everything afterwards

Uxio0: Yeah, but are we sure `0-bytes` are not allowed anywhere else?

Uxio0:
@rmeissner PR was updated with more cases and always ignoring and logging an error inserting a token so we don't 500 again

Uxio0:
I will merge it as I'm seeing 500 in production and this should fix them

PR #730, opened 2022-03-08
Fix admin search for integer fields

Uxio0:
- When trying to search for a field using a string in the admin field, there was an error 500
- For example, if searching for a `blockHash` there was an error when trying to format the string to a integer


rmeissner (approved):
I find the title of the PR misleading. Might be nice to have a better description here. 

As far as I understood the PR allows to search for a block by block hash, right?

PR #743, opened 2022-03-16
Configure all-auth

Uxio0:


Uxio0:
Replaced by https://github.com/gnosis/safe-transaction-service/pull/753

PR #753, opened 2022-03-21
Add buzzfeed sso support

Uxio0:


Uxio0:
Replaces #551

config/settings/base.py
```diff
@@ -196,7 +196,8 @@
 ADMIN_URL = "admin/"
 # https://docs.djangoproject.com/en/dev/ref/settings/#admins
 ADMINS = [
-    ("""Gnosis""", "dev@gnosis.pm"),
+    ("uxio.fuentefria", "uxio.fuentefria@gnosis.pm"),
+    ("richard", "richard.meissner@gnosis.pm"),
```
fmrsabino, line 200: I think this is better to be set on an environment variable.

With the amount of public usage by different orgs this is definitely something that admins would configure.

fmrsabino (changes requested):

config/settings/base.py
Uxio0: You are right, what I would do it's make it an environment variable but leave ourselves as the default ones:
- We already had gnosis configuration in place in that variable
- Removing us would imply modifying 12 deployments in production + 6 in staging

Uxio0: Done 👍🏻

fmrsabino (approved):

safe_transaction_service/history/views.py
```diff
@@ -66,13 +66,20 @@ class AboutView(APIView):
 
     renderer_classes = (JSONRenderer,)
 
-    @method_decorator(cache_page(60 * 60))  # 1 hour
+    # @method_decorator(cache_page(60 * 60))  # 1 hour
```
fmrsabino, line 69: I only saw this now but was this intended? @Uxio0

Uxio0: No. I disabled cache to test the headers

Uxio0: Thanks for noticing

Uxio0: https://github.com/gnosis/safe-transaction-service/pull/757

PR #767, opened 2022-03-30
Add support for non-TLD urls

### LINKED ISSUE #765: URL field validation applied to new Web Hook is too strict
# What is needed?
The URL field validation from the Django admin panel prevents the addition of URLs such as `http://nginx:8000/cgw/...` or any URL that contains a name that potentially leverages the name resolution for containers in the same `docker-compose.yml` files.

The lack of a top level domain (TLD) value in the URL, renders the URL invalid

# Background
This is required to prevent workarounds the admin panel for the purpose of this guide: https://github.com/gnosis/safe-infrastructure/

# Related issues
- User feedback highlighting pain points of the setup: https://github.com/gnosis/safe-infrastructure/issues/22
- This has been addressed in the config service successfully: https://github.com/gnosis/safe-config-service/issues/231

# Endpoint
There is no specific endpoint. This is rather an admin panel configuration task.

## URL
N/A

## Response
N/A

fmrsabino:
Closes #765

- `WebHook.url` now accepts any url as long the scheme is `http` or `https` and a network location is set
- This allows this field to have values like `http://tx-service` which were not valid under `URLField`


safe_transaction_service/history/models.py
```diff
@@ -1646,10 +1647,25 @@ def matching_for_address(self, address: str):
         return self.filter(Q(address=address) | Q(address=None))
 
 
+def _validate_webhook_url(url: str) -> None:
```
Uxio0, line 1650: Maybe we can refactor this method and tests at some point to `gnosis-py`, maybe even a new `ModelField` integrating this serializer and the `max_lenght` set to 255 by default. Just an idea for the future

Uxio0 (approved):

PR #771, opened 2022-03-31
Don't notify owner again if it has a delegate set

Uxio0:
- Owners were notified again if a delegate was set due to a typo


Uxio0:
Related to #1725

fmrsabino (approved):

PR #815, opened 2022-05-05
Maintain order for owners when adding/removing/swapping

### LINKED ISSUE #814: Make sure owners are processed and stored in blockchain order
# What is needed?
Follow the same logic than https://github.com/safe-global/safe-contracts/blob/v1.3.0/contracts/base/OwnerManager.sol when storing owners

Uxio0:
- Guarantee we keep owners in the database in the same order they are kept in blockchain
- Closes #814

### DIFF
--- a/safe_transaction_service/history/indexers/tx_processor.py
+++ b/safe_transaction_service/history/indexers/tx_processor.py
@@ -191,22 +191,22 @@ class SafeTxProcessor(TxProcessor):
                 return True
         return False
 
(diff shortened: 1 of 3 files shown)


rmeissner (approved):

rmeissner:
Should we have tests for this?

Uxio0:
> Should we have tests for this?

You are right, we have for all the process but not for this

Uxio0:
Tests added

PR #824, opened 2022-05-09
Use flower for celery monitoring

Uxio0:
- Run flower on scheduler machine (default port 5555)

### DIFF
--- a/docker/web/celery/scheduler/run.sh
+++ b/docker/web/celery/scheduler/run.sh
@@ -10,8 +10,7 @@ else
 fi
 
 
-echo "==> $(date +%H:%M:%S) ==> Installing flower for monitoring <=="
-pip install 'celery[flower]'
-sleep 5
+sleep 10
 echo "==> $(date +%H:%M:%S) ==> Running Celery beat <=="
+celery -C -A config.celery_app flower &
 exec celery -C -A config.celery_app beat -S django_celery_beat.schedulers:DatabaseScheduler --loglevel $log_level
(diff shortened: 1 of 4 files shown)


docker/web/celery/scheduler/run.sh
```diff
@@ -10,8 +10,6 @@ else
 fi
 
 
-echo "==> $(date +%H:%M:%S) ==> Installing flower for monitoring <=="
-pip install 'celery[flower]'
-sleep 5
+sleep 10
```
fmrsabino, line 13: Why do we need this?

docker/web/celery/worker/run.sh
```diff
@@ -16,7 +16,7 @@ echo "==> $(date +%H:%M:%S) ==> Setting up service... "
 python manage.py setup_service
 
 MAX_MEMORY_PER_CHILD="${WORKER_MAX_MEMORY_PER_CHILD:-2097152}"
-MAX_TASKS_PER_CHILD="${MAX_TASKS_PER_CHILD:-500000}"
+MAX_TASKS_PER_CHILD="${MAX_TASKS_PER_CHILD:-1000000}"
```
fmrsabino, line 19: What is the reason for this change?

fmrsabino, line 19: What is the reason for this change>

docker/web/celery/scheduler/run.sh
Uxio0: Waiting for migrations to be run (this was already here, was removed this monday when doing testing with flower)

docker/web/celery/worker/run.sh
Uxio0: Prevent memory leaks when running a lot of tasks

Uxio0: 1M is a number that we should not reach soon, but my plan is to implement metrics with flower/grafana/prometheus to check that

fmrsabino (approved):

PR #830, opened 2022-05-12
Use Rabbitmq instead of Redis for task queues

Uxio0:
## Background
Redis is not working pretty well when we have a lot of tasks

## Solution
- Use RabbitMQ as the recommended queue manager
- Priority numbers are reversed when using Redis or RabbitMQ (take a look at https://docs.celeryq.dev/en/stable/userguide/routing.html -> Priority Order)


docker-compose.yml
```diff
@@ -23,6 +23,11 @@ services:
       - --save ""
       - --appendonly no
 
+  rabbitmq:
+    image: rabbitmq:alpine
+    # ports:
+    #  - "5555:5555"
```
fmrsabino, line 29: Shall we specify the ports here?

fmrsabino (approved):
🐇

docker-compose.yml
Uxio0: No, as I don't think we should expose them

Uxio0: Removed

PR #844, opened 2022-05-19
Optimize internal tx processing

Uxio0:
- Optimize internal tx processing
- Set version v4.3.1


safe_transaction_service/history/models.py
```diff
@@ -1283,13 +1284,8 @@ def __str__(self):
...
     @property
-    def execution_date(self) -> Optional[datetime.datetime]:
-        if (
-            self.internal_tx.ethereum_tx_id
-            and self.internal_tx.ethereum_tx.block_id is not None
-        ):
-            return self.internal_tx.ethereum_tx.block.timestamp
-        return None
+    def execution_date(self) -> datetime.datetime:
+        return self.internal_tx.timestamp
```
fmrsabino, line 1288: This field was optional before. Do we have guarantees that we will always have a value here?

fmrsabino (approved):

safe_transaction_service/history/models.py
Uxio0: Yes, as `traces` are always executed and never pending. I don't know why it was optional before

fmrsabino: Got it 👌

PR #887, opened 2022-06-08
Return only relevant master copies

Uxio0:
- Modifies master copy endpoint
- Don't return not L2 MasterCopies on L2 networks


moisses89 (approved):
Tested locally

PR #890, opened 2022-06-09
Rename SafeLastStatus get_or_create -> get_or_generate

Uxio0:
- There's no need to override Django ORM method


moisses89 (approved):
I agree ;)

PR #891, opened 2022-06-10
Disable pending blocks for owner retrieval

Uxio0:
- It failed with some nodes and doesn't make a big change
- Refactored some code


rmeissner (approved):

Uxio0:
@dasanra @rmeissner @moisses89 
More info about this:
- `pending` was working on public node https://mainnet.optimism.io/ : `Geth/v1.9.10-stable/linux-amd64/go1.18`
- `pending` is not working on Infura even if they are using the same version: `Geth/v1.9.10-stable/linux-amd64/go1.15.15`

PR #894, opened 2022-06-10
Add endpoint to retrieve Safes by enabled module

### LINKED ISSUE #893: Endpoint for fetching safes with a given module enabled.
# What is needed?
An endpoint to the transaction service that is similar to `/owners/{address}/safes/`, but is for modules, something like `/modules/{address}/safes/`. So that you can query for all of the safes that have enabled a given module.

# Background
There are many cases where a given module might be enabled on many safes. For example, some modules (like the spending limits module) are deployed as singletons to be used by many safes or a DAO framework might be enabled as a module to multiple safes it controls.

# Endpoint
Described above.

## URL
`GET /api/v1/modules/{address}/safes/`

## Response
```
{
  "safes": [
    "0xdead...beef",
    "0x1234...5678"
  ]
}
```

Uxio0:
- Closes #893


safe_transaction_service/history/views.py
```diff
@@ -1012,6 +1012,36 @@ def get_queryset(self):
...
+            return Response(
+                status=status.HTTP_422_UNPROCESSABLE_ENTITY,
+                data={
+                    "code": 1,
+                    "message": "Checksum address validation failed",
+                    "arguments": [address],
+                },
+            )
+
+        safes_for_owner = SafeLastStatus.objects.addresses_for_module(address)
```
rmeissner, line 1039: `for_module`

rmeissner (approved):

Issue #895, opened 2022-06-11
Duplicate txs returned


tschubotz:
**Describe the bug**
Duplicate transactionsreturned

**To Reproduce**
1. Open https://gnosis-safe.io/app/eth:0xa4487e1abc196e80f35d0ce8639e97bcd96f3ea9/transactions/history
2. Check tx history

-> Seemslike txs are returned duplicated.

![image](https://user-images.githubusercontent.com/1380529/173200495-2c706228-0c38-49c1-b9b1-5ff4a952922d.png)

**Expected behavior**
All txs are there once

**Environment (please complete the following information):**
 - Staging or production? prod
 - Which chain? mainnet
 - OS: MacO
 - Browser Chrome

**Additional context**

https://safe-transaction.gnosis.io/api/v1/safes/0xa4487e1aBc196E80f35d0ce8639e97bCd96F3EA9/multisig-transactions/

Uxio0:
It's the first time I see that. Maybe it's a really old bug as it's the first time I see that. I reprocessed the Safe and now it's fixed

Uxio0:
We had the same info on staging, so that leads me to think it was a bug on the service long time ago

moisses89:
I think that we should create an issue to track it.

PR #901, opened 2022-06-13
Improve tx-processor testing

Uxio0:
- Wrote a new test while fixing #895


moisses89 (commented):
There is only included the test where is the fix?

Uxio0:
> There is only included the test where is the fix?

There's no fix. Check the name of the PR 😉

moisses89:
> 

Agree was to confirm.

moisses89 (approved):
Looks good

PR #907, opened 2022-06-16
Optimize EIP55 address calculation

Uxio0:
- Related to https://github.com/safe-global/safe-eth-py/pull/269


moisses89 (approved):
Looks good

Issue #909, opened 2022-06-16
Gunicorn web worker is using all the provided RAM


Uxio0:
**Describe the bug**
Gunicorn web worker is using all the provided RAM

**To Reproduce**
Run the web worker for mainnet.

**Expected behavior**
Service uses all the provided RAM and at some point is out of memory

**Environment**
 Production Mainnet

**Additional context**
We must analyze where the RAM is spent and if we have a memory leak

Uxio0:
Fixed by e06cca5bbd2458204688b91ba052f1eb9ae505d5

PR #910, opened 2022-06-17
Mitigate gunicorn memory leaks

Uxio0:
- Mitigate gunicorn memory leaks
- Set version 4.3.8


luarx (approved):

Uxio0:
Related to https://github.com/safe-global/safe-transaction-service/issues/909

PR #925, opened 2022-06-28
Optimize collectibles DB query

Uxio0:
- Use a proper query instead of returning all elements and do the processing on Python
- Related to https://github.com/safe-global/safe-pm/issues/89

### DIFF
--- a/safe_transaction_service/history/models.py
+++ b/safe_transaction_service/history/models.py
@@ -23,7 +23,7 @@ from django.conf import settings
 from django.contrib.postgres.fields import ArrayField
 from django.contrib.postgres.indexes import GinIndex
 from django.core.exceptions import ValidationError
-from django.db import IntegrityError, models, transaction
+from django.db import IntegrityError, connection, models, transaction
 from django.db.models import Case, Count, Index, JSONField, Max, Q, QuerySet
 from django.db.models.expressions import F, OuterRef, RawSQL, Subquery, Value, When
 from django.db.models.functions import Coalesce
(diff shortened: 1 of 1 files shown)


safe_transaction_service/history/models.py
```diff
@@ -499,41 +500,45 @@ def to_erc721_transfer(self):
...
+                                WHERE  "_from" = %s AND "to" != "_from"
+                                GROUP  BY address,
+                                          token_id) Q2
+                            ON Q1.address = Q2.address
+                               AND Q1.token_id = Q2.token_id
+                WHERE Q1.count > COALESCE(Q2.count, 0)
+                """,
+                [hex_address, hex_address],
+            )
+            return [
```
moisses89, line 538: What would it happen if there is any change on the database model? Would be detected by unit test to change the query acordingly?

Uxio0: Yes

moisses89: Ok, I would like to suggest catch the exception (if model changes) and mange it using orm queries like a second chance. What do you think?

Uxio0: I don't agree. If model changes we will notice due to the tests. If we do as you propose and model changes, tests will keep passing and we will not notice that we are using a suboptimal version of the query

Uxio0: I'm not very fond of using raw queries unless it's totally needed, but I'm alright if they are well tested

moisses89: Make sense, I hope that unit test get the change before integration test.

moisses89 (approved):

safe_transaction_service/history/models.py
Uxio0: Of course. You can try as you don't seem convinced 😉

PR #950, opened 2022-07-07
Add filters for collectibles SQL query

Uxio0:
- Filter `trusted` and `spam` tokens on the collectibles query


moisses89 (approved):
Looks good

Uxio0:
> Looks good

I think it can be more optimal using a JOIN instead of WHERE IN, I will check tomorrow before merging

Uxio0:
@moisses89 Added a DB index for SPAM tokens

PR #966, opened 2022-07-18
Auto reindex ERC20/721 periodically

Uxio0:
- Use the same mechanism that's currently in place for auto reindexing
  txs (every 2 hours last 2 hours are reindexed)


safe_transaction_service/history/tests/test_commands.py
```diff
@@ -258,6 +258,69 @@ def test_reindex_master_copies(self, current_block_number_mock: PropertyMock):
...
+                    f"--block-process-limit={block_process_limit}",
+                    f"--from-block-number={from_block_number}",
+                    stdout=buf,
+                )
+                self.assertIn(
+                    f"Start reindexing addresses {[safe_contract.address]}",
+                    cm.output[0],
+                )
+                self.assertIn("found 0 traces/events", cm.output[1])
+                self.assertIn(
```
moisses89, line 308: Maybe include a test with 1 or more events? What do you think?

safe_transaction_service/history/services/index_service.py
```diff
@@ -348,15 +327,16 @@ def reindex_master_copies(
...
+            logger.info("Start reindexing addresses %s", addresses)
+            current_block_number = self.ethereum_client.current_block_number
             stop_block_number = (
                 min(current_block_number, to_block_number)
                 if to_block_number
                 else current_block_number
             )
             block_number = from_block_number
             while block_number < stop_block_number:
+                block_number = min(block_number, stop_block_number)
```
moisses89, line 339: When stop_block_number could be less than block_number?

safe_transaction_service/history/indexers/events_indexer.py
```diff
@@ -36,9 +36,6 @@ def __init__(self, *args, **kwargs):
         kwargs.setdefault(
             "blocks_to_reindex_again", 10
         )  # Reindex last 10 blocks every run of the indexer
-        kwargs.setdefault(
-            "confirmations", 2
-        )  # Due to reorgs, wait for the last 2 blocks
         kwargs.setdefault("query_chunk_size", settings.ETH_EVENTS_QUERY_CHUNK_SIZE)
```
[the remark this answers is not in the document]
Uxio0: No need for confirmations if we enable reindexing, if by chance there's any transaction missing it will appear later

safe_transaction_service/history/services/index_service.py
Uxio0: You are right, refactored this part with a `for` instead of a `while`

safe_transaction_service/history/tests/test_commands.py
Uxio0: I would like to refactor `IndexService` at some point (to fix the circular imports, between other things) so I don't want to spend much time on it for now 😃

moisses89 (approved):
We are going to see how it works on staging :)

hectorgomezv (approved):

PR #969, opened 2022-07-19
Improve performance collectibles pagination

moisses89:
### What was wrong?
The new place where pagination is applied is going to be better to reduce the queries to database or blockchain node. 

@gnosis/safe-services

### DIFF
--- a/safe_transaction_service/history/services/collectibles_service.py
+++ b/safe_transaction_service/history/services/collectibles_service.py
@@ -312,18 +312,18 @@ class CollectiblesService:
         if not addresses_with_token_ids:
             return [], 0
 
+        count = len(addresses_with_token_ids)
+        # TODO Paginate on DB
+        if limit is not None:
+            addresses_with_token_ids = addresses_with_token_ids[offset : offset + limit]
+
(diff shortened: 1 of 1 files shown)


safe_transaction_service/history/services/collectibles_service.py
```diff
@@ -312,19 +312,19 @@ def _get_collectibles(
         if not addresses_with_token_ids:
             return [], 0
 
+        count = len(addresses_with_token_ids)
+        # TODO Paginate on DB
+        if limit is not None:
```
rmeissner, line 317: Isn't this on db level? Django orm maps this to a sql query, right.

Uxio0: I'm doing some magic to get all the collectibles for a Safe from DB: https://github.com/safe-global/safe-transaction-service/blob/master/safe_transaction_service/history/models.py#L516

Even for the Safes with most collectibles query is pretty fast so we decided not to write a new `COUNT(*)` query or even modify this one for using `LIMIT` and `OFFSET` for now. It's it not a difficult change and we should do it at some point

PR #974, opened 2022-07-21
Refactor configuration

Uxio0:
- Remove security configurations already handled on nginx or gunicorn
- Don't force cookies to be secure
- Remove not used email configuration
- Remove custom configuration for templates, as they are not used
- Rename users/emails for safe.global
- Enable DEBUG always for `local` configuration file


config/settings/local.py
```diff
@@ -1,43 +1,36 @@
...
-        "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
-        "LOCATION": "",
+        "BACKEND": "django_redis.cache.RedisCache",
+        "LOCATION": REDIS_URL,
+        "OPTIONS": {
+            "CLIENT_CLASS": "django_redis.client.DefaultClient",
+            # Mimicing memcache behavior.
+            # http://niwinz.github.io/django-redis/latest/#_memcached_exceptions_behavior
+            "IGNORE_EXCEPTIONS": True,
+        },
```
fmrsabino, line 30: So for the local one caching layer we assume Redis as the default one?

config/settings/production.py
```diff
@@ -85,40 +58,17 @@
...
-    "DJANGO_DEFAULT_FROM_EMAIL",
-    default="Gnosis Safe Transaction Service <noreply@gnosis.pm>",
-)
-# https://docs.djangoproject.com/en/dev/ref/settings/#server-email
-SERVER_EMAIL = env("DJANGO_SERVER_EMAIL", default=DEFAULT_FROM_EMAIL)
-# https://docs.djangoproject.com/en/dev/ref/settings/#email-subject-prefix
-EMAIL_SUBJECT_PREFIX = env(
-    "DJANGO_EMAIL_SUBJECT_PREFIX", default="[Gnosis Safe Push Service]"
-)
+    SSO_ADMINS = env.list("SSO_ADMINS", default=["richard", "uxio"])
```
fmrsabino, line 61: Maybe we can remove this completely and have it set on the deployment side?

config/settings/local.py
Uxio0: Yes, same as we use Postgres for local servers we use redis too

hectorgomezv (approved):

PR #996, opened 2022-08-02
Update to flake8 v5

Uxio0:


setup.cfg
```diff
@@ -1,7 +1,7 @@
 [flake8]
 max-line-length = 88
 select = C,E,F,W,B,B950
-extend-ignore = E203, E501, F841
+extend-ignore = E203,E501,F841,W503
```
moisses89, line 4: commas without space? 
W503 should be ignored?

Uxio0: It's INI format, not python 😉 , so better without spaces: https://en.wikipedia.org/wiki/INI_file

Uxio0: `503` was ignored by default in `flake8 < 4`, I don't know why they enabled it as `503` and `504` are incompatible

Uxio0: https://docs.python.org/3/library/configparser.html

moisses89 (approved):

PR #1002, opened 2022-08-03
Disable index_erc20_events_out_of_sync_task scheduled task

Uxio0:
- This task was created when we were indexing Safe addresses one by one and some of them missed sync while others were syncing
- As we are indexing erc20 events in big batches, I think this task is not needed anymore and can only cause conflicts with the regular indexing task, as this task is called once every 5 minutes and has no limit on how many times can be spammed. It can always be called from manage.py


moisses89:
🤔 then transaction-service was indexing the same transaction two times? This could be a big improvement

moisses89 (approved):

PR #1006, opened 2022-08-05
Fix swagger for TokenPrice endpoint

Uxio0:
- It was showing limit and offset on Swagger


moisses89 (approved):
Yes! didn't make sense :)

PR #1038, opened 2022-08-24
Implement prices for the Cronos chain

pkakelas:
### What was wrong?

The Cronos network had wrong prices.

### How was it fixed?

Implemented a `CRO-USDT` price pair on Kucoin. 
And by the way, refactored the code of KuCoin to match that of the BinanceClient.


Uxio0 (approved):
Really nice code

PR #1040, opened 2022-08-25
Recover and store again last price if cannot get price

### LINKED ISSUE #1019: Improve our retrieve prices for tokens
# What is needed?
Our current solution to retrieve the prices is returning `0.0` values when cannot get the price.  
The `0.0` price is an acceptable value when we don't know the real price but it is weird for an user that suddenly pass from `1700$` to `0$` several times in the same day.
For this I would like to research solutions that keep an acceptable value without back to 0 with each error.
   
We have two things to improve here:
- We shoudln't have any errors when connecting to the nodes to retrieve the pricing
- If some error happens we should keep old value if price cannot be calculated, it's better than nothing

# Related issues
https://github.com/safe-global/safe-transaction-service/issues/311

# Endpoint
`https://safe-transaction.mainnet.gnosis.io/api/v1/tokens/<token-address>/prices/usd/`

moisses89:
### What was wrong?
Closes #1019 
### How was it fixed?
- Store and return the last valid price if cannot obtain prices from external services. 
- Refactor calculate_token_eth_price_task

@gnosis/safe-services

### DIFF
--- a/safe_transaction_service/tokens/services/price_service.py
+++ b/safe_transaction_service/tokens/services/price_service.py
@@ -360,3 +360,30 @@ class PriceService:
                 FiatCode.USD,
                 token_eth_values_with_timestamp.timestamp,
             )
+
+    def get_eth_price_oracles(self, token_address: ChecksumAddress):
+        """
+        Calculate eth_price from oracles
+        :param token_address
+        """
+        return (
+            self.get_token_eth_value(token_address)
+            or self.get_token_usd_price(token_address)
+            / self.get_native_coin_usd_price()
+        )
+
+    def get_eth_price_composed_oracles(self, token_address: ChecksumAddress):
+        """
+        Calculate eth_price from composed oracles
+        :param token_address
+        """
+        eth_price = 0
+        if underlying_tokens := self.get_underlying_tokens(token_address):
+            for underlying_token in underlying_tokens:
+                # Find underlying token price and multiply by quantity
+                address = underlying_token.address
+                eth_price += (
+                    self.get_eth_price_oracles(address) * underlying_token.quantity
+                )
+
+        return eth_price
--- a/safe_transaction_service/tokens/services/price_service.py
+++ b/safe_transaction_service/tokens/services/price_service.py
@@ -361,10 +361,11 @@ class PriceService:
                 token_eth_values_with_timestamp.timestamp,
             )
 
-    def get_eth_price_oracles(self, token_address: ChecksumAddress):
+    def get_eth_price_from_oracles(self, token_address: ChecksumAddress) -> float:
         """
         Calculate eth_price from oracles
         :param token_address
+        :return: Current token price
         """
         return (
             self.get_token_eth_value(token_address)
@@ -372,10 +373,13 @@ class PriceService:
             / self.get_native_coin_usd_price()
         )
 
-    def get_eth_price_composed_oracles(self, token_address: ChecksumAddress):
+    def get_eth_price_from_composed_oracles(
+        self, token_address: ChecksumAddress
+    ) -> float:
         """
         Calculate eth_price from composed oracles
         :param token_address
+        :return: Current token price
         """
         eth_price = 0
         if underlying_tokens := self.get_underlying_tokens(token_address):
@@ -383,7 +387,7 @@ class PriceService:
                 # Find underlying token price and multiply by quantity
                 address = underlying_token.address
                 eth_price += (
-                    self.get_eth_price_oracles(address) * underlying_token.quantity
+                    self.get_eth_price_from_oracles(address) * underlying_token.quantity
                 )
 
         return eth_price
--- a/safe_transaction_service/tokens/tasks.py
+++ b/safe_transaction_service/tokens/tasks.py
@@ -63,24 +63,10 @@ def calculate_token_eth_price_task(
     )  # Expire in 15 minutes
     if key_was_set or force_recalculation:
         price_service = PriceServiceProvider()
-        eth_price = (
-            price_service.get_token_eth_value(token_address)
-            or price_service.get_token_usd_price(token_address)
-            / price_service.get_native_coin_usd_price()
-        )
-        if not eth_price:  # Try composed oracles
-            if underlying_tokens := price_service.get_underlying_tokens(token_address):
-                eth_price = 0
-                for underlying_token in underlying_tokens:
-                    # Find underlying token price and multiply by quantity
-                    address = underlying_token.address
-                    eth_price += (
-                        calculate_token_eth_price_task(
-                            address,
-                            f"price-service:{address}:eth-price",  # TODO Refactor all the calculation logic
-                        ).eth_value
-                        * underlying_token.quantity
-                    )
+        eth_price = price_service.get_eth_price_oracles(token_address)
+        if not eth_price:
+            eth_price = price_service.get_eth_price_composed_oracles(token_address)
+
         logger.debug("Calculated eth-price=%f for token=%s", eth_price, token_address)
         if not eth_price:
             logger.warning(
--- a/safe_transaction_service/tokens/tasks.py
+++ b/safe_transaction_service/tokens/tasks.py
@@ -82,7 +82,7 @@ def calculate_token_eth_price_task(
                         * underlying_token.quantity
                     )
         if not eth_price:
-            logger.warning("Cannot calculate eth price for token=%s", token_address)
+            logger.warning("Cannot calculate eth price for token=%s - Trying to use previous price", token_address)
             last_redis_value = redis.get(redis_key)
             if last_redis_value:
                 eth_price = EthValueWithTimestamp.from_string(
--- a/safe_transaction_service/tokens/tasks.py
+++ b/safe_transaction_service/tokens/tasks.py
@@ -81,18 +81,26 @@ def calculate_token_eth_price_task(
                         ).eth_value
                         * underlying_token.quantity
                     )
-        if eth_price:
-            eth_value_with_timestamp = EthValueWithTimestamp(eth_price, now)
-            redis.setex(redis_key, redis_expiration_time, str(eth_value_with_timestamp))
-            if not getattr(settings, "CELERY_ALWAYS_EAGER", False):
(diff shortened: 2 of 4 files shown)


safe_transaction_service/tokens/tasks.py
```diff
@@ -81,18 +81,26 @@ def calculate_token_eth_price_task(
...
-            if not getattr(settings, "CELERY_ALWAYS_EAGER", False):
-                # Recalculate price before cache expires and prevents recursion checking Celery Eager property
-                calculate_token_eth_price_task.apply_async(
-                    (token_address, redis_key),
-                    {"force_recalculation": True},
-                    countdown=redis_expiration_time - 300,
-                )
-        else:
+        if not eth_price:
             logger.warning("Cannot calculate eth price for token=%s", token_address)
```
Uxio0, line 85:
```suggestion
            logger.warning("Cannot calculate eth price for token=%s - Trying to use previous price", token_address)
```

safe_transaction_service/tokens/tasks.py
```diff
@@ -81,18 +81,26 @@ def calculate_token_eth_price_task(
...
+            last_redis_value = redis.get(redis_key)
+            if last_redis_value:
+                eth_price = EthValueWithTimestamp.from_string(
```
Uxio0, line 88: logger.warning("Using previous eth price for token=%s", token_address)

safe_transaction_service/tokens/tasks.py
```diff
@@ -81,18 +81,26 @@ def calculate_token_eth_price_task(
...
+                    last_redis_value.decode()
+                ).eth_value
+            else:
+                return EthValueWithTimestamp(eth_price, now)
```
Uxio0, line 92: logger.warning("Cannot calculate eth price for token=%s", token_address)

Uxio0 (changes requested):

safe_transaction_service/tokens/tasks.py
```diff
@@ -81,18 +81,26 @@ def calculate_token_eth_price_task(
                         ).eth_value
                         * underlying_token.quantity
                     )
-        if eth_price:
-            eth_value_with_timestamp = EthValueWithTimestamp(eth_price, now)
```
Uxio0, line 85: logger.debug("Calculated eth-price=%f for token=%s", eth_price, token_address)

safe_transaction_service/tokens/tasks.py
```diff
@@ -81,18 +81,32 @@ def calculate_token_eth_price_task(
...
-            if not getattr(settings, "CELERY_ALWAYS_EAGER", False):
-                # Recalculate price before cache expires and prevents recursion checking Celery Eager property
-                calculate_token_eth_price_task.apply_async(
-                    (token_address, redis_key),
-                    {"force_recalculation": True},
-                    countdown=redis_expiration_time - 300,
-                )
-        else:
-            logger.warning("Cannot calculate eth price for token=%s", token_address)
+        logger.debug("Calculated eth-price=%f for token=%s", eth_price, token_address)
```
hectorgomezv, line 84: Maybe I'm getting this wrong, but this message is logged even if `eth_price` wasn't calculated, right?

safe_transaction_service/tokens/tasks.py
```diff
@@ -81,18 +81,32 @@ def calculate_token_eth_price_task(
...
+        if not eth_price:
```
hectorgomezv, line 85: Personal opinion: I feel this `calculate_token_eth_price_task` function is getting way longer, what about splitting into several small private functions?

Uxio0: Agree, I think this function can use some big refactor

safe_transaction_service/tokens/tasks.py
moisses89: yes, when can't get price the previous functions returns `0.0`

safe_transaction_service/tokens/services/price_service.py
```diff
@@ -360,3 +360,30 @@ def get_cached_usd_values(
...
+        Calculate eth_price from oracles
+        :param token_address
+        """
+        return (
+            self.get_token_eth_value(token_address)
+            or self.get_token_usd_price(token_address)
+            / self.get_native_coin_usd_price()
+        )
+
+    def get_eth_price_composed_oracles(self, token_address: ChecksumAddress):
```
moisses89, line 375: I don't feel confident with this function and the unique test that I could see is checking that the result is greater than 0 `self.assertGreater(eth_value_with_timestamp.eth_value, 0.0)` 
Maybe add a new test for this function mocking the composed oracles? What do you think @Uxio0 ?

safe_transaction_service/tokens/tasks.py
moisses89: > Personal opinion: I feel this `calculate_token_eth_price_task` function is getting way longer, what about splitting into several small private functions?

@hectorgomezv looks better for you?

safe_transaction_service/tokens/services/price_service.py
Uxio0: Agree, please add a better test, you can use a well known BALANCER pool like https://app.balancer.fi/#/pool/0xa6f548df93de924d73be7d25dc02554c6bd66db500020000000000000000000e

moisses89: > Agree, please add a better test, you can use a well known BALANCER pool like https://app.balancer.fi/#/pool/0xa6f548df93de924d73be7d25dc02554c6bd66db500020000000000000000000e

Finally I mocked the test because I think that we should test the function that calculates the prices from composed oracles instead test the oracles. Test oracles should be done in safe-eth-py project.

safe_transaction_service/tokens/tasks.py
hectorgomezv: Ok, so I assume this would print `Calculated eth-price=0.0 for token=whatever` it's ok to me, just verifying this should print the line in both scenarios 👍🏻

safe_transaction_service/tokens/tasks.py
```diff
@@ -63,36 +63,36 @@ def calculate_token_eth_price_task(
...
+                token_address,
+            )
+            last_redis_value = redis.get(redis_key)
+            if last_redis_value:
+                logger.warning("Using previous eth price for token=%s", token_address)
+                eth_price = EthValueWithTimestamp.from_string(
+                    last_redis_value.decode()
+                ).eth_value
+            else:
+                logger.warning("Cannot calculate eth price for token=%s", token_address)
```
hectorgomezv, line 83: Could `eth_price` still being `0.0` at this point? It could happend in some (rare) scenarios, right? In that case I assume we prefer sending `0.0` to the client and don't throw an exception.

moisses89: This is currently returning `0.0` and don't throwing an exception :) `logger.warning` is a log warning level information.

hectorgomezv: Yes, I know, I mean, we chose returning 0.0 over throwing an exception, right? My doubt is if it could be better to actually throw the exception 🙂

hectorgomezv: Ok, I read some old discussions about this and the conclusion was that 0 is an adequate value in these cases, so nothing to argue on my side, great job @moisses89 🙂

hectorgomezv (approved):

safe_transaction_service/tokens/tasks.py
Uxio0: If you return an exception here there will never be a price for the token and calculation will be retried

moisses89: @uxio0 I'm not sure if we want this, could increase a lot the tasks for tokens that have not price.

Uxio0: Maybe I was not clear, we don't want to raise an exception 😉

safe_transaction_service/tokens/services/price_service.py
```diff
@@ -360,3 +360,30 @@ def get_cached_usd_values(
                 FiatCode.USD,
                 token_eth_values_with_timestamp.timestamp,
             )
+
+    def get_eth_price_oracles(self, token_address: ChecksumAddress):
```
Uxio0, line 364: Please add return types for the functions (and also add `:return:` to the docs)

Uxio0: Maybe the name could be `get_eth_price_from_oracles`?

safe_transaction_service/tokens/services/price_service.py
Uxio0, line 375: `get_eth_price_from_composed_oracles`

safe_transaction_service/tokens/tests/test_price_service.py
```diff
@@ -215,3 +216,20 @@ def test_token_eth_value_mocked(self, kyber_get_price_mock: MagicMock):
...
+    )
+    @mock.patch.object(
+        PriceService, "get_token_eth_value", autospec=True, return_value=1.0
+    )
+    def test_eth_price_from_composed_oracles(
+        self, get_token_eth_value_mock: MagicMock, price_service_mock: MagicMock
+    ):
+        price_service = self.price_service
+        token_one = UnderlyingToken("0xdF5e0e81Dff6FAF3A7e52BA697820c5e32D806A8", 0.5)
+        token_two = UnderlyingToken("0xdF5e0e81Dff6FAF3A7e52BA697820c5e32D806A8", 0.5)
```
Uxio0, line 231: At least mock something realistic, not the same token on the same percentage 😉

Uxio0 (changes requested):
Really good work overall

safe_transaction_service/tokens/tests/test_price_service.py
moisses89: Ok, I mocked this pool with the current percentages. https://pools.balancer.exchange/#/pool/0xe7ce624c00381b4b7abb03e633fb4acac4537dd6/

Uxio0 (approved):

PR #1049, opened 2022-08-30
Fix cronos mainnet address

pkakelas:
### What was wrong?

The cronos mainnet address for v1.3.0 l2 was not right. 
Check out the address in safe deployments: https://github.com/safe-global/safe-deployments/pull/93/files


Uxio0 (approved):

Uxio0:
Sounds good

PR #1058, opened 2022-09-06
Adjust defaults for indexer

Uxio0:
- Make block process limit adjustment more optimistic (increase faster)
- By default query ERC20 transfers by batches of 1,000 instead of 5,000 (some networks cannot handle 5,000, and it can always be adjusted by envar)


fmrsabino (approved):

hectorgomezv (approved):
LGTM 🙂

PR #1060, opened 2022-09-09
Implement ERC20 indexing status endpoint

Uxio0:
Add a new endpoint `api/v1/about/erc20-indexing` that returns information about the ERC20 indexing status. Response will be:

```json
{
    "current_block_number": "<str>",
    "minimum_indexed_block_number": "<int>",
    "synced": "<bool>",
}
```

- `current_block_number`: Current RPC block number.
- `minimum_indexed_block_number`: Block number for the less synced Safe (the one with the minimum indexed block number)
- `synced`: `True` if service is considered synced, `False` otherwise. Service will be considered `synced` if `minimum_indexed_block_number` is at least `ETH_REORG_BLOCKS` behind the `current_block`

### DIFF
--- a/safe_transaction_service/history/views.py
+++ b/safe_transaction_service/history/views.py
@@ -3,7 +3,7 @@ import logging
 from typing import Any, Dict, Tuple
 
 from django.conf import settings
-from django.db.models import Count
+from django.db.models import Count, Min
 from django.utils.decorators import method_decorator
 from django.views.decorators.cache import cache_page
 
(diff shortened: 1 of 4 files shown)


safe_transaction_service/history/views.py
```diff
@@ -148,6 +148,41 @@ def get(self, request, format=None):
...
+
+    def get(self, request):
+        """
+        Get current indexing status for ERC20/721 events
+        """
+        current_block_number = EthereumClientProvider().current_block_number
+        min_erc20_block_number = SafeContract.objects.aggregate(
+            min_erc20_block_number=Min("erc20_block_number")
+        )["min_erc20_block_number"]
+        if min_erc20_block_number is None:  # Still nothing indexed
```
fmrsabino, line 162: The default for `erc20_block_number` is `0`. Wouldn't the aggregate never be `None`?

fmrsabino (approved):

safe_transaction_service/history/views.py
Uxio0: You can see the line is covered by the tests 😉 https://coveralls.io/builds/52347914/source?filename=safe_transaction_service%2Fhistory%2Fviews.py#L162

PR #1068, opened 2022-09-22
Optimize check index problems

Uxio0:
- Optimize nonce retrieval
- Print block number where the first issue was found
- Nonce query batch size was increased from `100` to `1000`, performance seems better this way

### DIFF
--- a/safe_transaction_service/history/management/commands/check_index_problems.py
+++ b/safe_transaction_service/history/management/commands/check_index_problems.py
@@ -1,8 +1,6 @@
-from typing import Any, Dict, Iterable
-
 from django.core.management.base import BaseCommand
 
-from gnosis.eth import EthereumClientProvider
+from gnosis.eth import EthereumClient
 from gnosis.eth.constants import NULL_ADDRESS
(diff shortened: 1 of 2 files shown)


safe_transaction_service/history/management/commands/check_index_problems.py
```diff
@@ -13,57 +11,37 @@
...
+        ).functions.nonce()
 
     def handle(self, *args, **options):
         fix = options["fix"]
 
         queryset = SafeLastStatus.objects.all()
         count = queryset.count()
-        batch = 100
-        ethereum_client = EthereumClientProvider()
+        batch = 1000
```
fmrsabino, line 29: The batch changed from 100 to 1000. Was this intended?

fmrsabino (approved):

safe_transaction_service/history/management/commands/check_index_problems.py
Uxio0: Yes, I was testing and it can safetly increased 👍🏻 . I will edit the PR to address this

PR #1073, opened 2022-09-23
Add /about/indexing endpoint

Uxio0:
- Refactor ERC20Indexing endpoint into service

### DIFF
--- a/safe_transaction_service/history/services/index_service.py
+++ b/safe_transaction_service/history/services/index_service.py
@@ -1,8 +1,9 @@
 import logging
+from dataclasses import dataclass
 from typing import Collection, List, Optional, OrderedDict, Union
 
 from django.db import IntegrityError, transaction
-from django.db.models import Q
+from django.db.models import Min, Q
 
 from eth_typing import ChecksumAddress
 from hexbytes import HexBytes
@@ -22,6 +23,23 @@ from ..models import (
 )
 
 logger = logging.getLogger(__name__)
+
+
+@dataclass
+class IndexingStatus:
+    current_block_number: int
+    erc20_block_number: int
+    erc20_synced: bool
+    master_copies_block_number: int
+    master_copies_synced: bool
+    synced: bool
+
+
+@dataclass
+class ERC20IndexingStatus:
+    current_block_number: int
+    erc20_block_number: int
+    erc20_synced: bool
 
 
 class IndexingException(Exception):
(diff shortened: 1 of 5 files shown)


safe_transaction_service/history/services/index_service.py
```diff
@@ -24,6 +25,23 @@
 logger = logging.getLogger(__name__)
 
 
+@dataclass
+class IndexingStatus:
+    current_block_number: int
+    erc20_block_number: int
+    erc20_synced: bool
+    master_copies_block_number: bool
```
moisses89, line 33: int ;)

safe_transaction_service/history/services/index_service.py
```diff
@@ -88,6 +106,51 @@ def block_get_or_create_from_block_hash(self, block_hash: int):
                 block, confirmed=confirmed
             )
 
+    def get_indexing_status(self) -> IndexingStatus:
+        current_block_number = self.ethereum_client.current_block_number
+        erc20_block_number = SafeContract.objects.aggregate(
+            min_erc20_block_number=Min("erc20_block_number")
+        )["min_erc20_block_number"]
+        if erc20_block_number is None:  # Still nothing indexed
+            erc20_block_number = current_block_number
```
moisses89, line 115: I didn't get this.
`current_block_number` is last block indexed? and `erc20_block_number` is last indexed block, I'd said that the last block indexed should be max instead min, but I'm sure that I'm missing something.

Uxio0: `current_block_number` is the current block on the chain, and `erc20_block_number` is the block where erc20 indexing is stuck, so, the minimum block number for all the Safes

safe_transaction_service/history/services/index_service.py
Uxio0: Fixed

safe_transaction_service/history/services/index_service.py
moisses89: Ok, then `current_block_number` means the position where we are on chain, and for this reason
`if erc20_block_number is None -> erc20_block_number = current block.` that means that there wasn't anything to index until such `current_block_number`, true?

safe_transaction_service/history/tests/test_views.py
```diff
@@ -93,29 +94,119 @@ def test_erc20_indexing_view(self, current_block_number_mock: PropertyMock):
...
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.data["current_block_number"], 2000)
+        self.assertEqual(response.data["erc20_block_number"], 10)
+        self.assertEqual(response.data["erc20_synced"], False)
+
+        SafeContractFactory(erc20_block_number=12)
+        response = self.client.get(url, format="json")
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.data["current_block_number"], 2000)
+        self.assertEqual(response.data["erc20_block_number"], 10)
```
moisses89, line 119: Shouldn't return 12?

Uxio0: No, as the minimum is 10

moisses89 (approved):

PR #1081, opened 2022-09-26
Allow forcing batch call for nonce retrieval

Uxio0:
- Erigon has a builtin EVM timeout of 5 seconds that can be exceeded when using Multicall


moisses89 (approved):
Look good for me!

PR #1083, opened 2022-09-29
Optimize all transactions query

Uxio0:
- Don't use a join for multisig hashes and module hashes. Check is faster when using the actual tables with their indexes
- Use `UNION ALL` instead of `UNION`. There are already checks for duplicated elements
- Force a pre-sort of tables by timestamp DESC. That way PostgreSQL optimizer will use a heapsort instead of quicksort

Before:
```
Disk: 643080kB (for ERC20)
Planning Time: 1.834 ms
Execution Time: 20388.061 ms
```
After:
```
Disk usage: 154936kB (for ERC20) 
Planning Time: 1.904 ms
Execution Time: 6629.214 ms
```

### DIFF
--- a/safe_transaction_service/history/services/transaction_service.py
+++ b/safe_transaction_service/history/services/transaction_service.py
@@ -157,13 +157,14 @@ class TransactionService:
                 "block",
                 "safe_nonce",
             )
+            .order_by("-execution_date")
         )
         # Block is needed to get stable ordering
 
         if not queued:  # Filter out txs with nonce >= Safe nonce
             last_nonce_query = (
                 MultisigTransaction.objects.filter(safe=safe_address)
-                .exclude(ethereum_tx=None)
+                .executed()
                 .order_by("-nonce")
                 .values("nonce")
             )
@@ -181,8 +182,8 @@ class TransactionService:
         module_tx_ids = (
             ModuleTransaction.objects.filter(safe=safe_address)
             .annotate(
-                execution_date=F("internal_tx__ethereum_tx__block__timestamp"),
-                block=F("internal_tx__ethereum_tx__block_id"),
+                execution_date=F("internal_tx__timestamp"),
+                block=F("internal_tx__block_number"),
                 safe_nonce=Value(0, output_field=Uint256Field()),
             )
             .values(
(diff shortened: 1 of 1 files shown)


safe_transaction_service/history/services/transaction_service.py
```diff
@@ -256,14 +262,15 @@ def get_all_tx_hashes(
...
         queryset = (
-            multisig_safe_tx_ids.union(erc20_tx_ids)
-            .union(erc721_tx_ids)
-            .union(internal_tx_ids)
-            .union(module_tx_ids)
+            multisig_safe_tx_ids.union(erc20_tx_ids, all=True)
+            .union(erc721_tx_ids, all=True)
+            .union(internal_tx_ids, all=True)
+            .union(module_tx_ids, all=True)
             .order_by("-execution_date", "-safe_nonce", "block", "-created")
```
moisses89, line 274: > Use UNION instead of UNION ALL. There are already checks for duplicated elements   

Maybe the PR description should be UNION ALL instead UNION?

Uxio0: Done

safe_transaction_service/history/services/transaction_service.py
```diff
@@ -181,8 +182,8 @@ def get_all_tx_hashes(
         module_tx_ids = (
             ModuleTransaction.objects.filter(safe=safe_address)
             .annotate(
-                execution_date=F("internal_tx__ethereum_tx__block__timestamp"),
-                block=F("internal_tx__ethereum_tx__block_id"),
+                execution_date=F("internal_tx__timestamp"),
+                block=F("internal_tx__block_number"),
```
moisses89, line 186: I'm not sure but seems that the timestamp before was a block timestamp but now is internal tx timestamp. 
This is not going to affect the result?

moisses89 (commented):
Looks good only one question. :)

safe_transaction_service/history/services/transaction_service.py
Uxio0: Because that table was refactored to also have timestamp, to optimize some queries, but I forgot to change this one 😉

fmrsabino (approved):

PR #1084, opened 2022-09-29
Collectibles cache invalidation

### LINKED ISSUE #1027: Cache MetadataRetrievalException for less than 24 hours
**Describe the bug**
I could check how when the tx-service cannot get the metadata from an ipfs, it is storing the exception on redis for 24hours.
I think that is related with the following code:
   ```
 @cache_memoize(
        60 * 60 * 24,
        prefix="collectibles-_retrieve_metadata_from_uri",
        cache_exceptions=(MetadataRetrievalException,),
    )  
```
The cache_memoize documentation says that cache_exceptions enables the function to store the exception and returns the exceptions in the subsequent function calls. Then the expected behaviour is return for 24hours an exception.
```
Exceptions. If the cached function raises any of these exceptions is the
    exception cached and raised as normal. Subsequent cached calls will
    immediately re-raise the exception and the function will not be executed.
    this tuple will be cached, all other will be propagated.
```
**Expected behavior**
I would like to don't store this exceptions or at least store it for less time.

**Additional context**
https://5afe.slack.com/archives/C01GL8X8L3H/p1660666286167059
Some issues are expected getting MetaData for a while (1 hour, 2...) then we should avoid consume requests trying it.

**Proposed solution**
* [x] Increase for success requests of MetaData to 2 Days.
* [x] Exception will be a variable time that is increasing with the retries, for example (1st 30m, 2nd 1 h, 3rd 3h) 
* [x] Number of retries 5.

### PR DESCRIPTION
**Pull request by moisses89**
Closes #1027 
- Remove cache_memoize
- The task execution is delayed 30 seconds
- The retry_backoff is configured to 60 seconds and 4 retries as maximum , the retries is going to happen like following:
  - first try at 30 seconds
  - second try at 60 seconds
  - third try at 120 secs
  - fourth try at 240.
  - fifth try at 480 secs

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -230,6 +230,10 @@ CELERY_BROKER_TRANSPORT_OPTIONS = {}
 CELERY_ROUTES = (
     [
         (
+            "safe_transaction_service.history.tasks.retry_get_metadata_task",
+            {"queue": "tokens"},
+        ),
+        (
             "safe_transaction_service.history.tasks.send_webhook_task",
             {"queue": "webhooks"},
         ),
@@ -237,10 +241,6 @@ CELERY_ROUTES = (
         ("safe_transaction_service.contracts.tasks.*", {"queue": "contracts"}),
         ("safe_transaction_service.notifications.tasks.*", {"queue": "notifications"}),
         ("safe_transaction_service.tokens.tasks.*", {"queue": "tokens"}),
-        (
-            "safe_transaction_service.history.tasks.retry_get_metadata_task",
-            {"queue": "tokens"},
-        ),
     ],
 )
 
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -237,6 +237,10 @@ CELERY_ROUTES = (
         ("safe_transaction_service.contracts.tasks.*", {"queue": "contracts"}),
         ("safe_transaction_service.notifications.tasks.*", {"queue": "notifications"}),
         ("safe_transaction_service.tokens.tasks.*", {"queue": "tokens"}),
+        (
+            "safe_transaction_service.history.tasks.retry_get_metadata_task",
+            {"queue": "tokens"},
+        ),
     ],
 )
 
--- a/safe_transaction_service/history/services/collectibles_service.py
+++ b/safe_transaction_service/history/services/collectibles_service.py
@@ -162,7 +162,7 @@ class CollectiblesService:
 
     @cachedmethod(cache=operator.attrgetter("cache_uri_metadata"))
     @cache_memoize(
-        60 * 60 * 24,
+        60 * 60 * 24 * 2,
         prefix="collectibles-_retrieve_metadata_from_uri",
         cache_exceptions=(MetadataRetrievalException,),
     )  # 1 day
--- a/safe_transaction_service/history/services/collectibles_service.py
+++ b/safe_transaction_service/history/services/collectibles_service.py
@@ -162,7 +162,7 @@ class CollectiblesService:
         self.ens_service: EnsClient = EnsClient(self.ethereum_network.value)
 
         self.cache_uri_metadata = TTLCache[str, Optional[Dict[str, Any]]](
-            maxsize=4096, ttl=self.COLLECTIBLE_EXPRIATION
+            maxsize=4096, ttl=self.COLLECTIBLE_EXPIRATION
         )  # 1 day of caching
         self.cache_token_info: TTLCache[ChecksumAddress, Erc721InfoWithLogo] = TTLCache(
             maxsize=4096, ttl=60 * 30
@@ -431,7 +431,7 @@ class CollectiblesService:
                 )
                 retry_get_metadata_task.apply_async(
                     kwargs={"address": collectible.address, "id": collectible.id},
-                    countdown=5,
+                    countdown=30,  # 30 seconds later
                 )
 
             collectible_with_metadata = CollectibleWithMetadata(
@@ -599,7 +599,7 @@ class CollectiblesService:
             }
             pipe.mset(redis_map_to_store)
             for key in redis_map_to_store.keys():
-                pipe.expire(key, self.COLLECTIBLE_EXPRIATION)  # 1 day of caching
+                pipe.expire(key, self.COLLECTIBLE_EXPIRATION)  # 1 day of caching
             pipe.execute()
             found_uris.update(blockchain_token_uris)
 
--- a/safe_transaction_service/history/services/collectibles_service.py
+++ b/safe_transaction_service/history/services/collectibles_service.py
@@ -154,6 +154,7 @@ class CollectiblesService:
     COLLECTIBLE_EXPIRATION = int(
         60 * 60 * 24 * 2
     )  # Keep collectibles by 2 days in cache
+    TOKEN_EXPIRATION = int(60 * 30)
 
     def __init__(self, ethereum_client: EthereumClient, redis: Redis):
         self.ethereum_client = ethereum_client
@@ -165,7 +166,7 @@ class CollectiblesService:
             maxsize=4096, ttl=self.COLLECTIBLE_EXPIRATION
         )  # 1 day of caching
         self.cache_token_info: TTLCache[ChecksumAddress, Erc721InfoWithLogo] = TTLCache(
-            maxsize=4096, ttl=60 * 30
+            maxsize=4096, ttl=self.TOKEN_EXPIRATION
         )  # 2 hours of caching
 
     def get_redis_metadata_key(self, address: str, id: int):
@@ -366,7 +367,6 @@ class CollectiblesService:
         """
         from ..tasks import retry_get_metadata_task
 
-        redis_pipe = self.redis.pipeline()
         collectibles_with_metadata: List[CollectibleWithMetadata] = []
         collectibles, count = self.get_collectibles(
             safe_address,
@@ -403,8 +403,8 @@ class CollectiblesService:
                 collectibles_with_metadata.append(None)  # Keeps the order
 
         _ = gevent.joinall(jobs)
-        collectibles_with_metadata_no_cached = []
-        redis_map_to_store = {}
+        collectibles_with_metadata_not_cached = []
+        redis_pipe = self.redis.pipeline()
         for collectible, job in zip(collectibles_no_cached, jobs):
             try:
                 metadata = job.get()
@@ -443,7 +443,7 @@ class CollectiblesService:
                 collectible.uri,
                 metadata,
             )
-            collectibles_with_metadata_no_cached.append(collectible_with_metadata)
+            collectibles_with_metadata_not_cached.append(collectible_with_metadata)
             redis_pipe.set(
                 self.get_redis_metadata_key(collectible.address, collectible.id),
                 json.dumps(collectible_with_metadata.__dict__),
@@ -457,7 +457,7 @@ class CollectiblesService:
             if collectibles_with_metadata[collectible_metadata_cached_index] is None:
                 collectibles_with_metadata[
                     collectible_metadata_cached_index
-                ] = collectibles_with_metadata_no_cached[collectibles_no_cached_index]
+                ] = collectibles_with_metadata_not_cached[collectibles_no_cached_index]
                 collectibles_no_cached_index += 1
 
         return collectibles_with_metadata, count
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: safe_transaction_service/history/services/collectibles_service.py [hunk 1]
```diff
@@ -160,55 +164,73 @@ def __init__(self, ethereum_client: EthereumClient, redis: Redis):
             maxsize=4096, ttl=60 * 30
         )  # 2 hours of caching
 
+    def get_redis_metadata_timeout_key(self, uri: str):
+        return f"metadata_timeout:{uri}"
+
     @cachedmethod(cache=operator.attrgetter("cache_uri_metadata"))
     @cache_memoize(
-        60 * 60 * 24,
+        60 * 60 * 24 * 2,
```
moisses89, line 172: Increase the metadata time because is immutable, we should decide which would be an appropriate value.

safe_transaction_service/history/services/collectibles_service.py
```diff
@@ -160,55 +164,73 @@ def __init__(self, ethereum_client: EthereumClient, redis: Redis):
             maxsize=4096, ttl=60 * 30
         )  # 2 hours of caching
 
+    def get_redis_metadata_timeout_key(self, uri: str):
+        return f"metadata_timeout:{uri}"
+
     @cachedmethod(cache=operator.attrgetter("cache_uri_metadata"))
     @cache_memoize(
```
moisses89, line 171: I would like remove this decorator and do it with django_cache, what do you think @Uxio0 ?

moisses89: The second exception shouldn't be necessary if we manage the cache without the decorator.

safe_transaction_service/history/services/collectibles_service.py
```diff
@@ -160,55 +164,73 @@ def __init__(self, ethereum_client: EthereumClient, redis: Redis):
...
+                        )
+
+                    return response.json()
+            except (IOError, ValueError) as e:
+                django_cache.set(
+                    self.get_redis_metadata_timeout_key(uri),
+                    "Metadata timeout",
+                    60 * 60 * 24 * 2,
+                )
+                chance_to_retrieve_metadata_from_uri.apply_async(
```
moisses89, line 228: The idea here is to give more opportunities when the error is a timeout, currently configured at 15 seconds.

safe_transaction_service/history/services/collectibles_service.py
Uxio0: Sure, it's inmutable, but why going from 1 day to 2 days?

safe_transaction_service/history/services/collectibles_service.py
```diff
@@ -160,55 +164,73 @@ def __init__(self, ethereum_client: EthereumClient, redis: Redis):
...
-                    )
-
-                if "application/json" not in content_type:
-                    raise MetadataRetrievalException(
-                        f"Content-type={content_type} for uri={uri} is not valid, "
-                        f'expected "application/json"'
-                    )
-
-                logger.debug("Got metadata for uri=%s", uri)
+        from ..tasks import chance_to_retrieve_metadata_from_uri
```
Uxio0, line 183: Task name should end `_task`. Also, `chance_to_retrieve` does not look clear to me

safe_transaction_service/history/services/collectibles_service.py
```diff
@@ -160,55 +164,73 @@ def __init__(self, ethereum_client: EthereumClient, redis: Redis):
...
 
-                # Some requests don't provide `Content-Length` on the headers
-                if len(response.content) > self.METADATA_MAX_CONTENT_LENGTH:
-                    raise MetadataRetrievalException(
-                        f"Retrieved content for uri={uri} is too big"
-                    )
+        uri_http = ipfs_to_http(uri)
+        if (
+            not django_cache.get(self.get_redis_metadata_timeout_key(uri))
```
Uxio0, line 187: You retrieve information from cache but you don't use it?

safe_transaction_service/history/tasks.py
```diff
@@ -472,3 +473,27 @@ def send_webhook_task(address: Optional[str], payload: Dict[str, Any]) -> int:
...
+    autoretry_for=(MetadataRetrievalExceptionTimeout,),
+    time_limit=LOCK_TIMEOUT,
+    retry_backoff=5,
+    retry_kwargs={"max_retries": 5},
+)
+def chance_to_retrieve_metadata_from_uri(self, uri: str) -> bool:
+    """
+    Give more chances to get the metadata of collectible
+    """
+    from django.core.cache import cache as django_cache
```
Uxio0, line 490: This can be imported outside the function. The only reason to import stuff inside a function is to prevent recursive imports

safe_transaction_service/history/tasks.py
```diff
@@ -472,3 +473,27 @@ def send_webhook_task(address: Optional[str], payload: Dict[str, Any]) -> int:
...
+
+    from .services import CollectiblesServiceProvider
+
+    collectibles_service = CollectiblesServiceProvider()
+    collectibles_service._retrieve_metadata_from_uri.invalidate(uri)
```
Uxio0, line 495: Why invalidating, if maybe you manage to get the actual metadata from the service?

safe_transaction_service/history/services/collectibles_service.py
moisses89: Imeant that we could increase the time that we storage the metadata to for examples 2 days, what do you think?

safe_transaction_service/history/services/collectibles_service.py
moisses89: Got it

safe_transaction_service/history/tasks.py
moisses89: The last commit changes how we manage the metadata cache

safe_transaction_service/history/services/collectibles_service.py
Uxio0: No strong opinion there, but I think 2 days should be alright too

safe_transaction_service/history/tasks.py
```diff
@@ -472,3 +479,58 @@ def send_webhook_task(address: Optional[str], payload: Dict[str, Any]) -> int:
 
         sent_requests += 1
     return sent_requests
+
+
+@app.shared_task(
+    bind=True,
+    soft_time_limit=SOFT_TIMEOUT,
+    autoretry_for=(MetadataRetrievalExceptionTimeout,),
```
hectorgomezv, line 487: First time this retry task is executed by an explicit call to `retry_get_metadata_task` and then it auto-retried every time `MetadataRetrievalExceptionTimeout` is thrown, right?

hectorgomezv (approved):
Looks good to me 🙂

safe_transaction_service/history/tasks.py
moisses89: yes, the first retry task is going to be executed after countdown, and then auto-retried following `exponential backoff` delay between retries, here more context, https://docs.celeryq.dev/en/stable/userguide/tasks.html#automatic-retry-for-known-exceptions. 
Reading again the documentation maybe is better instead to define a `maximum number of retries`, define a `maximum number of seconds`. What do you think? I'm worried about default value that is less than the time for the fourth retry.

hectorgomezv: Hmm, according the docs, setting `retry_backoff` to `3`:
```
For example, if this option is set to 3, the first retry will delay 3 seconds, the second will delay 6 seconds, the third will delay 12 seconds, the fourth will delay 24 seconds, and so on.
```

So it depends on which time we think it's appropriate to be retrying. Maybe it's more readable by setting the `retry_backoff_max` to a number of seconds as you said (one hour for example).

moisses89: It's true, then I must modify the 3 for a higher one then.

safe_transaction_service/history/services/collectibles_service.py
```diff
@@ -154,25 +162,23 @@ def __init__(self, ethereum_client: EthereumClient, redis: Redis):
         self.ens_service: EnsClient = EnsClient(self.ethereum_network.value)
 
         self.cache_uri_metadata = TTLCache[str, Optional[Dict[str, Any]]](
-            maxsize=4096, ttl=60 * 60 * 24
+            maxsize=4096, ttl=self.COLLECTIBLE_EXPIRATION
         )  # 1 day of caching
         self.cache_token_info: TTLCache[ChecksumAddress, Erc721InfoWithLogo] = TTLCache(
             maxsize=4096, ttl=60 * 30
```
Uxio0, line 168: I think we should also create a variable for `TOKEN_EXPIRATION`

safe_transaction_service/history/tasks.py
```diff
@@ -472,3 +479,58 @@ def send_webhook_task(address: Optional[str], payload: Dict[str, Any]) -> int:
...
+    Give more chances to get the metadata of collectible
+    """
+    from .services import CollectiblesServiceProvider
+
+    collectibles_service = CollectiblesServiceProvider()
+    redis_key = collectibles_service.get_redis_metadata_key(address, id)
+    redis = get_redis()
+    redis_collectible = redis.get(redis_key)
+
+    if not redis_collectible:
```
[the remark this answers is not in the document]
Uxio0: Shouldn't you check that `redis_collectible` holds no metadata? Just in case this task is triggered at the same time from 2 different web workers and metadata is already retrieved by one worker

Uxio0 (changes requested):

safe_transaction_service/history/services/collectibles_service.py
```diff
@@ -387,18 +422,44 @@ def _get_collectibles_with_metadata(
...
-                    collectible.id,
+            except MetadataRetrievalExceptionTimeout:
+                metadata = {}
+                logger.warning(
+                    "Timeout retrieving metadata on token-uri=%s for token-address=%s, retrying asyncronous ",
                     collectible.uri,
-                    metadata,
+                    collectible.address,
                 )
+                retry_get_metadata_task.apply_async(
```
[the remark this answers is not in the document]
moisses89: I could check that work.

safe_transaction_service/history/tasks.py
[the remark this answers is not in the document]
moisses89: Agree with the comments, but not sure if celery workers can take the same task at the same task, do you think that this is possible?

safe_transaction_service/history/services/collectibles_service.py
```diff
@@ -442,7 +504,7 @@ def get_collectibles_with_metadata_paginated(
         )
 
     @cachedmethod(cache=operator.attrgetter("cache_token_info"))
-    @cache_memoize(60 * 60, prefix="collectibles-get_token_info")  # 1 hour
+    @cache_memoize(TOKEN_EXPIRATION, prefix="collectibles-get_token_info")  # 1 hour
```
moisses89, line 507: In the code above token_expiration was 30 minutes here 1 hour, what is the correct value? @Uxio0

safe_transaction_service/history/tasks.py
moisses89: Changed to 60 seconds

safe_transaction_service/history/services/collectibles_service.py
```diff
@@ -387,18 +422,44 @@ def _get_collectibles_with_metadata(
...
+        redis_pipe.execute()
+
+        # Creates a collectibles metadata keeping the initial order
+        collectibles_no_cached_index = 0
+        for collectible_metadata_cached_index in range(len(collectibles_with_metadata)):
+            if collectibles_with_metadata[collectible_metadata_cached_index] is None:
+                collectibles_with_metadata[
+                    collectible_metadata_cached_index
+                ] = collectibles_with_metadata_not_cached[collectibles_no_cached_index]
+                collectibles_no_cached_index += 1
```
Uxio0, line 461:
```suggestion
        for collectible_metadata_cached_index in range(len(collectibles_with_metadata)):
            if collectibles_with_metadata[collectible_metadata_cached_index] is None:
                collectibles_with_metadata[collectible_metadata_cached_index] = collectibles_with_metadata_not_cached.pop(0)
```

Uxio0 (changes requested):

config/settings/base.py
```diff
@@ -237,6 +237,10 @@
         ("safe_transaction_service.contracts.tasks.*", {"queue": "contracts"}),
         ("safe_transaction_service.notifications.tasks.*", {"queue": "notifications"}),
         ("safe_transaction_service.tokens.tasks.*", {"queue": "tokens"}),
+        (
```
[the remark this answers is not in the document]
moisses89: You meant before history tasks, it's correct?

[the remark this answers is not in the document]
Uxio0: Yes, but please check it first if possible as I don't trust this documentation too much

Uxio0 (approved):

config/settings/base.py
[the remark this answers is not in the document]
moisses89: Ok I will verify

PR #1120, opened 2022-10-17
Perform ERC20 indexing concurrently

Uxio0:
## ERC20 indexing before this PR
All the Safe addresses reach the indexer, it splits the addresses in batches of `ETH_EVENTS_QUERY_CHUNK_SIZE`, queries the node and everything happens sequentially, so for `ETH_EVENTS_QUERY_CHUNK_SIZE=10` and `105 Safes` it will take 11 queries to the node.

## ERC20 indexing after this PR
As we are using round robin balancing with internal and external nodes multiple queries can be sent to multiple nodes, and that's why we will do it concurrently (configurable on the `ETH_EVENTS_GET_LOGS_CONCURRENCY` parameter, so we can adjust it depending on network conditions)

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -388,6 +388,9 @@ ETH_EVENTS_BLOCK_PROCESS_LIMIT = env.int(
 ETH_EVENTS_BLOCK_PROCESS_LIMIT_MAX = env.int(
     "ETH_EVENTS_BLOCK_PROCESS_LIMIT_MAX", default=0
 )  # Maximum number of blocks to process together when searching for events. 0 == no limit.
+ETH_EVENTS_GET_LOGS_CONCURRENCY = env.int(
+    "ETH_EVENTS_GET_LOGS_CONCURRENCY", default=20
+)  # Number of concurrent requests to `getLogs`
 ETH_EVENTS_QUERY_CHUNK_SIZE = env.int(
     "ETH_EVENTS_QUERY_CHUNK_SIZE", default=1_000
 )  # Number of addresses to use as `getLogs` parameter. `0 == no limit`. By testing `1_000` looks like a good default
--- a/safe_transaction_service/history/indexers/erc20_events_indexer.py
+++ b/safe_transaction_service/history/indexers/erc20_events_indexer.py
@@ -3,10 +3,12 @@ from collections import OrderedDict
 from logging import getLogger
 from typing import Iterator, List, Sequence
 
+import gevent
 from cache_memoize import cache_memoize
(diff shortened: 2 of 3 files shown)


safe_transaction_service/history/indexers/erc20_events_indexer.py
```diff
@@ -79,18 +81,25 @@ def _do_node_query(
...
+                addresses_chunk,
+                from_block=from_block_number,
+                to_block=to_block_number,
+            )
+            for addresses_chunk in addresses_chunks
+        ]
+
+        with self.auto_adjust_block_limit(from_block_number, to_block_number):
+            # Check how long the first job takes
+            gevent.joinall(jobs[:1])
```
moisses89, line 99: AFAIK before we was adjusting the `block_limit` for every `addresses_chunk` but this new implementation adjusts only for the first job, could this affect to the performance? I mean, could the node change his performance and the first block_limit don't be the best choose for the next jobs?

hectorgomezv: I assume the node for the in each batch first job would be chosen using round-robin, so maybe it's not accurate in the first execution but I think it will be converge to the average (if I'm not missing something).

hectorgomezv (approved):
LGTM 🚀

moisses89 (approved):

config/settings/base.py
```diff
@@ -388,6 +388,9 @@
 ETH_EVENTS_BLOCK_PROCESS_LIMIT_MAX = env.int(
     "ETH_EVENTS_BLOCK_PROCESS_LIMIT_MAX", default=0
 )  # Maximum number of blocks to process together when searching for events. 0 == no limit.
+ETH_EVENTS_GET_LOGS_CONCURRENCY = env.int(
+    "ETH_EVENTS_GET_LOGS_CONCURRENCY", default=20
+)  # Number of concurrent requests to `getLogs`
```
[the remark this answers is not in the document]
Uxio0: It's a little arbitrary, based on what I've seen in the nodes (we were using concurrent fetching some time ago). We can always adjust it depending on the node provider/network

safe_transaction_service/history/indexers/erc20_events_indexer.py
Uxio0: @moisses89 It's exactly what @hectorgomezv exposed. Maybe first time it's not very accurate, but at more requests are queued into the node it should stabilize

Uxio0:
Thanks guys for the quick review

PR #1137, opened 2022-10-27
Reduce number of errors when can't get metadata

### LINKED ISSUE #1136: Reduce number of ERRORS in retry_get_metadata_task
**Describe the bug**
`retry_get_metadata_task` is producing a lot of noise due to `MetadataRetrievalExceptionTimeout` being trigerred when retries exceed

**Expected behavior**
Don't log error, metadata retrieval exceptions are expected

 - Production
 - Mainnet and other networks

moisses89:
Closes #1136 

### How was it fixed? 🎯
- catch the exception
- implements a custom backoff retry 
- write a log message only when all retries were consumed
- the task finish always successfully

### DIFF
--- a/safe_transaction_service/history/tasks.py
+++ b/safe_transaction_service/history/tasks.py
@@ -1,6 +1,7 @@
 import contextlib
 import dataclasses
 import json
+import random
 from functools import cache
 from typing import Any, Dict, Optional
 from urllib.parse import urlparse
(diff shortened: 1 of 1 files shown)


Uxio0 (approved):

safe_transaction_service/history/tasks.py
```diff
@@ -534,25 +533,42 @@ def retry_get_metadata_task(
...
+                countdown=int(
+                    random.uniform(55, 65) * retry_get_metadata_task.request.retries
+                )
+            )
+        else:
+            logger.debug(
+                "Cannot get the metadata from %s after %i retries",
+                collectible.uri,
+                retry_get_metadata_task.request.retries,
+            )
```
Uxio0, line 572:
```suggestion
            logger.debug(
                "Cannot get the metadata from %s after %i retries",
                collectible.uri,
                retry_get_metadata_task.request.retries,
                exc_info=True
            )
```

moisses89: Changed the message to include that was a timeout exception

PR #1138, opened 2022-10-28
Add endpoints for creating EIP1271 signatures

Uxio0:
- Add `GET /messages/<id>/` endpoint, detail view for a `safe-message`
- Add `POST /messages/<id>/signatures/` endpoint, to add a signature for a `safe-message`
- Add `GET /messages/safes/<safe_address>/` endpoint, to get a list for all the safe messages sorted by date for a `safe`
- Add `POST /messages/safes/<safe_address>/` endpoint, to create a new Safe message for a Safe


safe_transaction_service/safe_messages/models.py
```diff
@@ -0,0 +1,68 @@
...
+
+from gnosis.eth.django.models import EthereumAddressV2Field, HexField, Keccak256Field
+from gnosis.safe.safe_signature import SafeSignatureType
+
+logger = getLogger(__name__)
+
+
+class SafeMessage(TimeStampedModel):
+    """
+    Safe Message (EIP-191 or EIP-712) to build an EIP-1271 signature from
```
mikhailxyz, line 17: nit: the message doesn't have to be an EIP-191, standards like [SIWE](https://eips.ethereum.org/EIPS/eip-4361) don't use EIP-191

safe_transaction_service/safe_messages/models.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+
+    safe = EthereumAddressV2Field(db_index=True)
+    # Message hash is tied to Safe domain, so it's guaranteed to be unique
+    message_hash = Keccak256Field(db_index=True, unique=True)
+    message = JSONField()  # String if EIP191, object if EIP712
+    proposed_by = EthereumAddressV2Field()  # Owner proposing the message
+    description = models.CharField(max_length=200, blank=True)
```
mikhailxyz, line 25: What do we plan to store in the description?

safe_transaction_service/safe_messages/models.py
Uxio0: For EIP-4361 unless I'm missing something I see that they EIP191 for encoding the message

safe_transaction_service/safe_messages/serializers.py
```diff
@@ -0,0 +1,153 @@
...
+            )
+
+        safe_signatures = SafeSignature.parse_signature(signature, safe_message_hash)
+        if len(safe_signatures) != 1:
+            raise ValidationError(
+                f"1 owner signature was expected, {len(safe_signatures)} received"
+            )
+
+        owners = get_safe_owners(safe_address)
+        proposed_by = safe_signatures[0].owner
```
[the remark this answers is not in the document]
Uxio0: That's right, but in this case we only allow one signature (check is a little above)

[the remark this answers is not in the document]
moisses89: `if len(safe_signatures) != 1` understood :smile:

safe_transaction_service/safe_messages/pagination.py
```diff
@@ -0,0 +1,11 @@
+from rest_framework.pagination import LimitOffsetPagination
+
+
+class DefaultPagination(LimitOffsetPagination):
+    max_limit = 200
+    default_limit = 100
+
+
+class SmallPagination(LimitOffsetPagination):
```
moisses89, line 9: This function is never used.

moisses89 (approved):
A minor comment!

PR #1140, opened 2022-10-28
Use docker latest tag for releases

Uxio0:
- `staging` images were tagged as `latest`
- Only releases will be tagged as `latest`


moisses89 (approved):
Looks good to me!

PR #1147, opened 2022-10-31
Use python 3.11

Uxio0:


moisses89:
Cache-tools added python 3.11 support see https://github.com/tkem/cachetools/blob/master/CHANGELOG.rst#v521-2023-01-08

moisses89 (approved):

PR #1162, opened 2022-11-08
Remove tokens blockchain retrieval on 404

hectorgomezv:
- [X] You ran `./run_tests.sh`
- [X] You ran `pre-commit run -a`

### What was wrong? 👾
Since tokens are indexed through an scheduled process, there's no need to make ad-hoc costly calls to the blockchain. This PR removes that functionality, so when a unknown token is requested, the service just responds 404 and does no further tasks.

Closes #

### How was it fixed? 🎯
I removed both the task and the unit tests which covered it.

### DIFF
--- a/safe_transaction_service/tokens/tasks.py
+++ b/safe_transaction_service/tokens/tasks.py
@@ -111,21 +111,3 @@ def fix_pool_tokens_task() -> Optional[int]:
         if number:
             logger.info("%d pool token names were fixed", number)
         return number
-
-
-@app.shared_task()
-@close_gevent_db_connection_decorator
-def get_token_info_from_blockchain_task(token_address: ChecksumAddress) -> bool:
-    """
-    Retrieve token information from blockchain
-
-    :param token_address:
-    :return: `True` if found, `False` otherwise
-    """
-    redis = get_redis()
-    key = f"token-task:{token_address}"
-    if result := redis.get(key):
-        return bool(int(result))
-    token_found = bool(Token.objects.create_from_blockchain(token_address))
-    redis.setex(key, 60 * 60 * 6, int(token_found))  # Cache result 6 hours
-    return token_found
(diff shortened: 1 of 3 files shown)


safe_transaction_service/tokens/tasks.py
```diff
@@ -111,21 +111,3 @@ def fix_pool_tokens_task() -> Optional[int]:
         if number:
             logger.info("%d pool token names were fixed", number)
         return number
-
-
-@app.shared_task()
-@close_gevent_db_connection_decorator
-def get_token_info_from_blockchain_task(token_address: ChecksumAddress) -> bool:
```
hectorgomezv, line 118: @Uxio0 your review here is appreciated 🙂 I removed the task completely but I'm not sure the cache management is correct. It seems this function also stores an `int` to indicate the token was retrieved (to avoid requesting it several time if I understand correctly), so I suppose the actual caching (if it's being cached) of the value is done outside this function 🤔 so it should be OK to remove this code completely.

Uxio0: No worries, it shouldn't have any conflicts. If we need that function we can just revert or just write it again

safe_transaction_service/tokens/views.py
```diff
@@ -37,11 +35,7 @@ def get(self, request, *args, **kwargs):
                 },
             )
 
-        try:
-            return super().get(request, *args, **kwargs)
-        except Http404 as exc:  # Try to get info about the token
-            get_token_info_from_blockchain_task.delay(address)
-            raise exc
+        return super().get(request, *args, **kwargs)
```
Uxio0, line 38: Maybe we can rename `Cache 1 hour, this should never change` to `Cache 1 hour, this does not change often`

Uxio0 (approved):
Good job!

safe_transaction_service/tokens/views.py
hectorgomezv: 👍🏻 changed in [50506cc](https://github.com/safe-global/safe-transaction-service/pull/1162/commits/50506cc3a626cd0bd6eb31ecf182bce222aba16c)

PR #1176, opened 2022-11-14
Update 1271 endpoints

Uxio0:
- Use `message_hash` as `private key` for `SafeMessage`
- Add `created` and `modified` to `SafeMessage` and `SafeMessageConfirmation`
- Add `preparedSignature` field to `SafeMessage`
- Fix swagger


fmrsabino:
I suggest holding off on this one until we clarify the requirements on the client side.

safe_transaction_service/safe_messages/serializers.py
```diff
@@ -134,12 +144,15 @@ def get_signature_type(self, obj: SafeMessageConfirmation) -> str:
...
 class SafeMessageResponseSerializer(serializers.Serializer):
+    created = serializers.DateTimeField()
+    modified = serializers.DateTimeField()
     safe = eth_serializers.EthereumAddressField()
     message_hash = eth_serializers.Sha3HashField()
     message = serializers.JSONField()
     proposed_by = eth_serializers.EthereumAddressField()
     description = serializers.CharField()
     confirmations = serializers.SerializerMethodField()
+    prepared_signature = serializers.SerializerMethodField()
```
fmrsabino, line 155: The naming on this one is a bit strange but I also don't have better suggestions... `computedSignature` or just `signature`? Maybe something that we can ask the team.

fmrsabino (approved):

safe_transaction_service/safe_messages/serializers.py
iamacook: How about `concatenated_signature`?

Uxio0: I don't have an strong opinion, but I'm between `computedSignature`, `preparedSignature` or `proposedSignature`. `concatenatedSignature` sounds that we are just pasting it together, and there's further logic involved

fmrsabino: Let's go with `preparedSignature` I'd say (no strong opinion). Maybe we can ask in the team chat but shouldn't block this PR.

The only requirement I'd say is that we should be aligned on this before a production release.

Uxio0:
@iamacook changes are already on staging

iamacook:
> @iamacook changes are already on staging

Alright, thanks for letting me know!

PR #1177, opened 2022-11-14
Add analytic transactions per safe app

### LINKED ISSUE #1165: [Safe Apps Analytics] Transactions count per Safe App
# Query
The following query is a generic idea of the query, it doesn't work on django orm:
`result = Multisigtransaction.filter(origin__contains='url').group_by(origin)`

# Result example:
The result is a list of Safe Apps and Custom Apps with the total count of transactions.
```
[
  {
    "url": "https://wry-graph.net",
    "name": "aperiam",
    "total-transactions": 56,
    "transactions-last-week": 51,
    "transactions-last-month": 14,
    "transactions-last-year": 79,
    "id": "1"
  },
  {
    "url": "https://smooth-feast.biz",
    "name": "quis",
    "total-transactions": 42,
    "transactions-last-week": 64,
    "transactions-last-month": 42,
    "transactions-last-year": 29,
    "id": "2"
  }
]
```

# Tasks
- [x] migrate origin from charField to JsonField
- [ ] create periodic task each 1 week to get the data from DB
- [x] store the data in cache
- [x] create an endpoint to return the result

# Comments
We should check that the same origin is grouped only one time, avoiding repeated origins, for example because the parameter order:
```
{'origin': '{"url":"https://app.uniswap.org/","name":"Uniswap"}'
{'origin': '{"name":"Uniswap", "url":"https://app.uniswap.org/"}'
```

moisses89:
Closes #1165 

- Add safe apps analytics V2
- Change field model origin to Jsonfield
- Add Environment variable to enable/disable analytics endpoints
- Add migration test


safe_transaction_service/history/models.py
```diff
@@ -1238,8 +1238,8 @@ class MultisigTransaction(TimeStampedModel):
     signatures = models.BinaryField(null=True, blank=True)  # When tx is executed
     nonce = Uint256Field(db_index=True)
     failed = models.BooleanField(null=True, blank=True, default=None, db_index=True)
-    origin = models.CharField(
-        null=True, blank=True, default=None, max_length=200
+    origin = models.JSONField(
+        null=True, blank=True, default=None
```
moisses89, line 1242: @Uxio0 What do you think about this change? In our staging database there are Null and random strings.

Uxio0: Random strings are valid jsons, but null are not. I would make `{}` the new `None`, so I won't allow `null=True` and we need to run a migration before so every `None` becomes `{}` (I don't know if Django can handle that)

moisses89: `None` is converted to `{}` if default is `dict`, but we have an issue migrating the strings without `{` like following:
Before migration
```
 {'origin': '{"url": "https://example2.com", "name": "afeApp2"}'},
 {'origin': 'eoo'},
 {'origin': None}]
```
When try to migrate the following error is showed:
```
django.db.utils.DataError: invalid input syntax for type json
DETAIL:  Token "eoo" is invalid.
CONTEXT:  JSON data, line 1: eoo
```
I can fix it doing converting `'` to `"` :
```
In : tx.origin
Out: 'eoo'
In: tx.origin = "eoo"
In: txs.save()
```
@Uxio0 this could be handle in the migration but maybe is going to be a heavy migration, what do you think?

moisses89: Done here https://github.com/safe-global/safe-transaction-service/pull/1177/commits/923a47b3697580c4ea5556cd662f8e92c26ac4a5

safe_transaction_service/history/migrations/0068_alter_multisigtransaction_origin.py
```diff
@@ -0,0 +1,48 @@
...
+
+def convert_str_to_json(apps, schema_editor):
+    MultisigTransaction = apps.get_model("history", "MultisigTransaction")
+    transactions = (
+        MultisigTransaction.objects.exclude(origin__icontains="{")
+        .filter(origin__isnull=False)
+        .iterator()
+    )
+    for transaction in transactions:
+        transaction.origin = json.dumps(transaction.origin)
```
Uxio0, line 16: - Shouldn't this be `json.loads`? You have a string and you want to load it as a Json. Is this tested? Unless `JSONField` is doing some conversion, you are just storing `str`.
- Users can store anything on this field, and `origin__icontains="{"` is not a good validation, so you should check for exceptions when loading the json
- Empty strings should be stored as `{}`
- Maybe we can try https://github.com/wemake-services/django-test-migrations

safe_transaction_service/history/migrations/0068_alter_multisigtransaction_origin.py
```diff
@@ -0,0 +1,48 @@
...
+def convert_jsonfield_to_charfield(apps, schema_editor):
+    MultisigTransaction = apps.get_model("history", "MultisigTransaction")
+    MultisigTransaction.objects.filter(origin__exact="{}").update(origin=None)
+    transactions = (
+        MultisigTransaction.objects.exclude(origin__icontains="{")
+        .filter(origin__isnull=False)
+        .iterator()
+    )
+    for transaction in transactions:
+        transaction.origin = json.loads(transaction.origin)
```
Uxio0, line 29: Here you have an `object` and you try to call `json.loads` , this will fail

safe_transaction_service/history/models.py
```diff
@@ -1238,8 +1238,8 @@ class MultisigTransaction(TimeStampedModel):
     signatures = models.BinaryField(null=True, blank=True)  # When tx is executed
     nonce = Uint256Field(db_index=True)
     failed = models.BooleanField(null=True, blank=True, default=None, db_index=True)
-    origin = models.CharField(
-        null=True, blank=True, default=None, max_length=200
+    origin = models.JSONField(
+        null=False, default=dict
```
Uxio0, line 1242: Remove `null=False`, as it's the default behaviour

safe_transaction_service/history/serializers.py
```diff
@@ -234,7 +235,14 @@ def validate(self, data):
 
     def save(self, **kwargs):
         safe_tx_hash = self.validated_data["contract_transaction_hash"]
-        origin = self.validated_data["origin"]
+        # Origin field on db is a JsonField
+        if self.validated_data["origin"] is not None:
+            try:
+                origin = json.loads(self.validated_data["origin"])
+            except ValueError:
+                origin = json.loads(json.dumps(self.validated_data["origin"]))
```
Uxio0, line 243:
```suggestion
                origin = self.validated_data["origin"]
```

This is the same of what you are writting

safe_transaction_service/history/serializers.py
```diff
@@ -234,7 +235,14 @@ def validate(self, data):
 
     def save(self, **kwargs):
         safe_tx_hash = self.validated_data["contract_transaction_hash"]
-        origin = self.validated_data["origin"]
+        # Origin field on db is a JsonField
+        if self.validated_data["origin"] is not None:
```
Uxio0, line 239:
```suggestion
        if self.validated_data["origin"]:
```

This way we will be storing also empty strings as `{}`, so it's better for our database consistency

safe_transaction_service/history/services/index_service.py
```diff
@@ -323,7 +323,7 @@ def _reprocess(self, addresses: List[str]):
 
         logger.info("Remove transactions automatically indexed")
         queryset = MultisigTransaction.objects.exclude(ethereum_tx=None).filter(
-            Q(origin=None) | Q(origin="")
+            Q(origin__exact={}) | Q(origin="")
```
Uxio0, line 326: We shouldn't have empty strings on database
```suggestion
            Q(origin__exact={})
```

Uxio0 (changes requested):

safe_transaction_service/analytics/tasks.py
```diff
@@ -0,0 +1,36 @@
...
+            tx_last_month=Count("origin__name", filter=Q(created__gt=last_month)),
+            tx_last_year=Count("origin__name", filter=Q(created__gt=last_year)),
+        )
+        .order_by("-total_tx")
+    )
+
+    if queryset:
+        redis_key = "analytics_transactions_per_safe_app"
+        redis = get_redis()
+        redis.set(redis_key, json.dumps(list(queryset)))
```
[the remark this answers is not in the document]
moisses89: yes, `MultisigTransactionQuerySet` is not an object serializable because the `values()`.

safe_transaction_service/analytics/tasks.py
```diff
@@ -0,0 +1,36 @@
...
+        .annotate(
+            total_tx=Count("origin__name"),
+            tx_last_week=Count("origin__name", filter=Q(created__gt=last_week)),
+            tx_last_month=Count("origin__name", filter=Q(created__gt=last_month)),
+            tx_last_year=Count("origin__name", filter=Q(created__gt=last_year)),
+        )
+        .order_by("-total_tx")
+    )
+
+    if queryset:
```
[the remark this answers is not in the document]
moisses89: Done [here](https://github.com/safe-global/safe-transaction-service/pull/1177/commits/36e2571f462ce4fda01edc71d867f1b69e3a7dc0)

safe_transaction_service/analytics/tasks.py
```diff
@@ -0,0 +1,36 @@
...
+from dateutil.relativedelta import relativedelta
+
+from safe_transaction_service.history.models import MultisigTransaction
+from safe_transaction_service.utils.redis import get_redis
+from safe_transaction_service.utils.tasks import LOCK_TIMEOUT, SOFT_TIMEOUT
+
+
+@app.shared_task(soft_time_limit=SOFT_TIMEOUT, time_limit=LOCK_TIMEOUT)
+def get_transactions_per_safe_app_task():
+    today = datetime.date.today()
```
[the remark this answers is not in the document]
moisses89: Done [here](https://github.com/safe-global/safe-transaction-service/pull/1177/commits/36e2571f462ce4fda01edc71d867f1b69e3a7dc0)

safe_transaction_service/history/models.py
moisses89: Done [here](https://github.com/safe-global/safe-transaction-service/pull/1177/commits/36e2571f462ce4fda01edc71d867f1b69e3a7dc0)

safe_transaction_service/history/serializers.py
moisses89: Done [here](https://github.com/safe-global/safe-transaction-service/pull/1177/commits/36e2571f462ce4fda01edc71d867f1b69e3a7dc0)

safe_transaction_service/history/migrations/0068_alter_multisigtransaction_origin.py
moisses89: > Shouldn't this be json.loads? You have a string and you want to load it as a Json. Is this tested? Unless JSONField is doing some conversion, you are just storing str.

It has mentioned here https://github.com/safe-global/safe-transaction-service/pull/1177#discussion_r1021943193

> Users can store anything on this field, and origin__icontains="{" is not a good validation, so you should check for exceptions when loading the json

It's an exclude the `alterJsonField` should convert any json string to json without issues. 

> Empty strings should be stored as {}

I agree I will add it in the migration. 

> Maybe we can try https://github.com/wemake-services/django-test-migrations

Sounds good, lets me try.

safe_transaction_service/history/migrations/0068_alter_multisigtransaction_origin.py
moisses89: AFAIK the reverse migration undo the migration executing the operations in a reverse way then we have not a object we have an string.

moisses89: The reason for this is because the django migration from `jsonfield` to `charfield` applies a json.loads for each row, then for example for "hello" is going to store in database '"hello"' instead  of "hello":
```
json.dumps("hello")
Out[34]: '"hello"'
```

safe_transaction_service/history/services/index_service.py
moisses89: > We shouldn't have empty strings on database

@Uxio0  I agree but could be because "" is a valid json and the object creation accepts it, removing the query filter the following test is going to fail [L158](https://github.com/safe-global/safe-transaction-service/pull/1177/files#diff-ed6ce143344e1aefcda7bdce62a260fd78d4569e90ffbc4d097eea27f911b337L158) because when create the multisigtransaction the origin is "". Null origins is formatted to {} but empty strings "" doesn't change when insert.

safe_transaction_service/history/serializers.py
moisses89: Done [here](https://github.com/safe-global/safe-transaction-service/pull/1177/commits/a653677a60cc3b11930ffaf85fec5f50ee8a4dec)

safe_transaction_service/analytics/tests/test_tasks.py
```diff
@@ -0,0 +1,46 @@
...
+
+from django.test import TestCase
+
+from safe_transaction_service.analytics.tasks import get_transactions_per_safe_app_task
+from safe_transaction_service.history.tests.factories import MultisigTransactionFactory
+from safe_transaction_service.utils.redis import get_redis
+
+
+class TestTasks(TestCase):
+    def test_get_transactions_per_safe_apps(self):
```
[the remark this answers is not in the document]
moisses89: Done [here](https://github.com/safe-global/safe-transaction-service/pull/1177/commits/70df3051a5490272f58a2747beee9c9bbe410f05)

safe_transaction_service/analytics/tests/test_views_v2.py
```diff
@@ -0,0 +1,100 @@
...
+            reverse("v2:analytics:analytics-multisig-txs-by-origin")
+        )
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        origin_1 = {"url": "https://example1.com", "name": "SafeApp1"}
+        origin_2 = {"url": "https://example2.com", "name": "SafeApp2"}
+
+        MultisigTransactionFactory(origin=origin_1)
+        # Execute the periodic task
+        get_transactions_per_safe_app_task()
```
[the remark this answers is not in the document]
moisses89: Done [here](https://github.com/safe-global/safe-transaction-service/pull/1177/commits/70df3051a5490272f58a2747beee9c9bbe410f05)

safe_transaction_service/analytics/views_v2.py
```diff
@@ -0,0 +1,22 @@
...
+from safe_transaction_service.utils.redis import get_redis
+
+
+class AnalyticsMultisigTxsByOriginListView(ListAPIView):
+    pagination_class = None
+    renderer_classes = (JSONRenderer,)
+    serializer_class = None
+
+    def get(self, request, format=None):
+        redis_key = "analytics_transactions_per_safe_app"
```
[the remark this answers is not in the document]
moisses89: I moved redis_key to the service like a constant :) 
Done [here](https://github.com/safe-global/safe-transaction-service/pull/1177/commits/70df3051a5490272f58a2747beee9c9bbe410f05)

safe_transaction_service/history/services/index_service.py
Uxio0: Then we might remove the test, as there's no way from user input to get a "" on database, or at least we are making our best to prevent that

safe_transaction_service/history/migrations/0068_alter_multisigtransaction_origin.py
Uxio0: Ok, but then an object will be stored as an object instead of a stringfied json, right?

safe_transaction_service/analytics/views_v2.py
```diff
@@ -13,10 +13,5 @@ class AnalyticsMultisigTxsByOriginListView(ListAPIView):
...
 
     def get(self, request, format=None):
-        redis_key = "analytics_transactions_per_safe_app"
-        redis = get_redis()
-        analytic_result = redis.get(redis_key)
-        if analytic_result:
-            return Response(json.loads(analytic_result))
-        else:
-            return Response([])
+        analytics_service = AnalyticsService()
```
Uxio0, line 16: Add a function that's cached to get the Analytics service, so it's a singleton and you don't initialize the service every time you use it

safe_transaction_service/analytics/services/analytics_service.py
```diff
@@ -0,0 +1,16 @@
+import json
+from typing import List
+
+from safe_transaction_service.utils.redis import get_redis
+
+
```
Uxio0, line 6:
```
@cached
def get_analytics_service() -> "AnalyticsService":
  return AnalyticsService()
```

safe_transaction_service/history/migrations/0068_alter_multisigtransaction_origin.py
moisses89: This is the backward migration objects should be stored like `stringfield` json and strings should stored like strings.

safe_transaction_service/history/migrations/0068_alter_multisigtransaction_origin.py
```diff
@@ -0,0 +1,53 @@
...
+            transaction.save()
+
+
+def repair_backward_migration(apps, schema_editor):
+    MultisigTransaction = apps.get_model("history", "MultisigTransaction")
+    # Empty objects should be None
+    MultisigTransaction.objects.filter(origin__exact="{}").update(origin=None)
+    # Remove duplicated quotes example: '"hello"'
+    transactions = (
+        MultisigTransaction.objects.exclude(origin__icontains="{")
```
moisses89, line 28: @Uxio0  I'm not happy with this at all, let me explain step by step. 
### DB initial state is migration 0068, then the field is a jsonField
- applying the back migration
- Django migration applies `json.dumps` at each row in jsonField to convert it to `charField`
- at the end, the nex operation is `repair_backward_migration` to convert this `"{}"` in `None` and to convert `'"something"'` in `"something"`
My question is regarding the string conversion step, I'm looking for any origin that not contains `{` to get non object strings, but this couldn't include for example 'te { st' that is a string that should be fixed. Have any idea how to query it? my issue is that all origins are strings at this point.

Uxio0: Understood then!
- Maybe you can just have an special case for "{}": `MultisigTransaction.objects.filter(origin='"{}"').update(origin=None)`
- You can do json.loads() and check the type, if that's a string just store the result, and if not store the value you had previously (for objects)

safe_transaction_service/history/migrations/0068_alter_multisigtransaction_origin.py
Uxio0: Ok

safe_transaction_service/history/migrations/0068_alter_multisigtransaction_origin.py
moisses89: Done [here](https://github.com/safe-global/safe-transaction-service/pull/1177/commits/618d32d05e372f78c40b7136dadd43ba32f6cbc2) :smile:

safe_transaction_service/history/migrations/0068_alter_multisigtransaction_origin.py
```diff
@@ -22,16 +22,15 @@ def prepare_migration_charfield_to_jsonfield(apps, schema_editor):
 def repair_backward_migration(apps, schema_editor):
     MultisigTransaction = apps.get_model("history", "MultisigTransaction")
     # Empty objects should be None
-    MultisigTransaction.objects.filter(origin__exact="{}").update(origin=None)
+    MultisigTransaction.objects.filter(origin="{}").update(origin=None)
```
Uxio0, line 25: As this query might be slow, better do it together:

`MultisigTransaction.objects.filter(Q(origin="{}") | Q(origin='"{}"')).update(origin=None)`

moisses89: of course

safe_transaction_service/history/management/commands/setup_service.py
```diff
@@ -117,6 +117,12 @@ def create_task(self) -> Tuple[PeriodicTask, bool]:
         1,
         IntervalSchedule.HOURS,
     ),
+    CeleryTaskConfiguration(
+        "safe_transaction_service.analytics.tasks.get_transactions_per_safe_app_task",
+        "Run query to get number of transactions grouped by safe-app",
+        1,
+        IntervalSchedule.DAYS,
```
moisses89, line 124: This should be 1 week.

moisses89: Done [here](https://github.com/safe-global/safe-transaction-service/pull/1177/commits/6cf257018140f2f7f87f7c496964f9d0889be999)

safe_transaction_service/analytics/views_v2.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class AnalyticsMultisigTxsByOriginListView(ListAPIView):
+    pagination_class = None
+    swagger_schema = None
+    renderer_classes = (JSONRenderer,)
+
+    def get(self, request, format=None):
+        analytics_service = get_analytics_service()
+        return Response(analytics_service.get_safe_transactions_per_safe_app())
```
Uxio0, line 17: Maybe you could use a serializer so we get swagger documentation, but it works for me as it is

safe_transaction_service/history/serializers.py
```diff
@@ -234,7 +235,14 @@ def validate(self, data):
 
     def save(self, **kwargs):
         safe_tx_hash = self.validated_data["contract_transaction_hash"]
-        origin = self.validated_data["origin"]
+        # Origin field on db is a JsonField
+        if self.validated_data["origin"]:
```
Uxio0, line 239: I think you should do this on `def validate_origin` and already return the object (or string) there

Uxio0 (approved):

safe_transaction_service/analytics/views_v2.py
moisses89: I will do in a different PR.

safe_transaction_service/history/serializers.py
moisses89: I will include both changes in different PR.

moisses89:
- Add safe apps analytics V2
- Change field model origin to Jsonfield
- Add Environment variable to enable/disable analytics endpoints
-  Add migration tests

PR #1178, opened 2022-11-16
Remove message signatures when owner is removed

Uxio0:


safe_transaction_service/history/indexers/tx_processor.py
```diff
@@ -235,9 +237,10 @@ def swap_owner(
         MultisigConfirmation.objects.remove_unused_confirmations(
             contract_address, safe_status.nonce, owner
         )
+        safe_message_models.SafeMessageConfirmation.objects.filter(owner=owner).delete()
```
moisses89, line 240: If `SafeMessageConfirmation` have a owner `foreignkey` with delete on cascade shouldn't remove the `messageconfirmation` if the owner is removed?

fmrsabino: `SafeMessageConfirmation` does not have a foreign key to the owner right? It has to the `SafeMessage`. If the `SafeMessage` is removed then the associated confirmations should be removed too.

safe_transaction_service/history/indexers/tx_processor.py
```diff
@@ -235,9 +237,10 @@ def swap_owner(
...
 
     def store_new_safe_status(
-        self, safe_last_status: SafeStatus, internal_tx: InternalTx
+        self, safe_last_status: SafeLastStatus, internal_tx: InternalTx
```
fmrsabino, line 243: What is the reason to change the type here?

safe_transaction_service/history/tests/test_tx_processor.py
```diff
@@ -117,14 +123,23 @@ def test_tx_processor_with_factory(self):
         self.assertEqual(safe_status.threshold, threshold)
 
         threshold = 1
-        # Check deleting the owner did delete this pending confirmation
+        # Check deleting the owner did delete this pending confirmation for both signature and transaction
```
fmrsabino, line 126: nitpick: this test has around 274 lines. By adding functionality and editing the test it becomes more difficult to maintain the test.

My suggestion is to split the test into multiple ones where we test these conditions seperately. The setup might end up being the same but by adding more meaning to what is expected on each test it also makes it easier to debug.

safe_transaction_service/safe_messages/migrations/0002_alter_safemessageconfirmation_owner.py
```diff
@@ -0,0 +1,20 @@
+# Generated by Django 4.1.3 on 2022-11-16 14:35
+
+from django.db import migrations
+
+import gnosis.eth.django.models
+
+
+class Migration(migrations.Migration):
```
fmrsabino, line 8: What is the context for this migration. The model was not changed 🤔

safe_transaction_service/safe_messages/tests/factories.py
```diff
@@ -33,4 +33,4 @@ class Params:
             encode_defunct(text=o.safe_message.message)
         )["signature"].hex()
     )
-    signature_type = 3
+    signature_type = 2
```
fmrsabino, line 36: Why was this changed?

safe_transaction_service/safe_messages/serializers.py
```diff
@@ -66,6 +66,11 @@ def validate(self, attrs):
             raise ValidationError(
                 f"1 owner signature was expected, {len(safe_signatures)} received"
             )
+        for safe_signature in safe_signatures:
+            if not safe_signature.is_valid(ethereum_client, safe_address):
+                raise ValidationError(
+                    f"Signature={safe_signature.signature.hex()} for owner={safe_signature.owner} is not valid"
+                )
```
fmrsabino, line 73: What is the reason for this check? Was it missing?

Maybe `SafeSignature.parse_signature` should throw in these cases or?

safe_transaction_service/history/indexers/tx_processor.py
Uxio0: That's it, no foreign key to the owner

safe_transaction_service/history/indexers/tx_processor.py
Uxio0: It was a typo before and I detected it when checking the `tx_processor` for implementing this feature

safe_transaction_service/history/tests/test_tx_processor.py
Uxio0: @fmrsabino This test checks a complete flow for indexing a Safe, and I cannot think of how to split as it depends on the status of the previous steps. This test is completely necesary and has been really useful for detecting indexing issues, but I agree that maybe we could have different tests for checking all the `tx_processor` options and branches

safe_transaction_service/safe_messages/migrations/0002_alter_safemessageconfirmation_owner.py
Uxio0: I forgot to add the migration in the previous PR. Either way I will reset migrations before going to production when modifying the `description` -> `safeAppId`

safe_transaction_service/safe_messages/tests/factories.py
Uxio0: Signature is an EOA and not a ETH_SIGN

fmrsabino: But this is a factory? Or did you mean to change the default?

safe_transaction_service/safe_messages/serializers.py
Uxio0: > What is the reason for this check? Was it missing?
Yes

> Maybe SafeSignature.parse_signature should throw in these cases or?
No, as for EIP1271 and ApprovedHash signatures a network request is required to check, and maybe is not desired. So first we parse the signatures (only in structure) and then you can validate them

fmrsabino: > > Maybe SafeSignature.parse_signature should throw in these cases or?
> 
> No, as for EIP1271 and ApprovedHash signatures a network request is required to check, and maybe is not desired. So first we parse the signatures (only in structure) and then you can validate them

Ok in my opinion we should review these cases (if in the other cases they should throw a validation error or not). If the result of a "failed" parse can be ignored in some cases then I agree that we don't need to change it.

fmrsabino (approved):

safe_transaction_service/safe_messages/tests/test_views.py
```diff
@@ -303,6 +324,26 @@ def get_eip712_payload():
             },
         )
 
+        # Test not valid signature
```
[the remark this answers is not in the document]
Uxio0: I refactored the 2 tests using `subTest`

safe_transaction_service/safe_messages/tests/factories.py
Uxio0: No, is just a factory 😉

safe_transaction_service/safe_messages/serializers.py
Uxio0: But the problem is that parse can be valid but signature not, and that depends for example is a hash is approved on the contract, and you don't want to check the signatures when you parse them as blockchain calls are not really fast (we are using it in a lot of places as it's a core function from `safe-eth-py`)

PR #1187, opened 2022-11-22
Add confirmationsRequired for not executed txs

### LINKED ISSUE #1148: Report confirmationsRequired for pending txns in `multisig-transactions/{safe_tx}`
# What is needed?
Pending transactions do not currently report the number of confirmations required for execution. This would be incredibly helpful to determine how far along a given `safe_tx` is without doing a separate call, which is unneeded.

Since the model stores the `threshold`, this doesn't seem to be a terribly expensive query to make.

# Endpoint
`/api/v1/multisig-transactions/{safe_tx}`

Uxio0:
- Closes #1148
- Start returning `confirmationsRequired` for transactions that are not executed (using the last available Safe status). Previously `null` was returned.

### DIFF
--- a/safe_transaction_service/history/tests/test_models.py
+++ b/safe_transaction_service/history/tests/test_models.py
@@ -1325,6 +1325,54 @@ class TestMultisigTransactions(TestCase):
             MultisigTransaction.objects.not_indexed_metadata_contract_addresses()
         )
 
+    def test_with_confirmations_required(self):
+        # This should never be picked
+        SafeStatusFactory(nonce=0, threshold=4)
+
+        multisig_transaction = MultisigTransactionFactory()
+        self.assertIsNone(
+            MultisigTransaction.objects.with_confirmations_required()
+            .first()
+            .confirmations_required
+        )
+
+        # SafeStatus not matching the EthereumTx
+        safe_status = SafeStatusFactory(nonce=1, threshold=8)
+        self.assertIsNone(
+            MultisigTransaction.objects.with_confirmations_required()
+            .first()
+            .confirmations_required
+        )
+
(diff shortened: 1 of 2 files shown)


Uxio0:
@fmrsabino This is something pending for a long time and I realised there was a simple way to achieve that. Would it break something on the cgw?

safe_transaction_service/history/tests/test_models.py
```diff
@@ -1325,6 +1325,69 @@ def test_not_indexed_metadata_contract_addresses(self):
...
+            8,
+        )
+
+        # It will not be picked, as EthereumTx is not matching
+        SafeStatusFactory(nonce=2, threshold=15)
+        self.assertEqual(
+            MultisigTransaction.objects.with_confirmations_required()
+            .first()
+            .confirmations_required,
+            8,
```
moisses89, line 1363: This part of the test is a little bit confusing for me, could we change `assertEqual` for `assertNotEqual` 15.

Uxio0: I don't think that's a good idea. I want to check that `confirmations_required` are now `8` and inserting a new `SafeStatus` didn't change it (maybe to `4` again due to some bad logic implementation). If I take your suggestion I would be missing a test case,

fmrsabino:
> @fmrsabino This is something pending for a long time and I realised there was a simple way to achieve that. Would it break something on the cgw?

Should be good on the API level: https://github.com/safe-global/safe-client-gateway/blob/cd7aceb2035f1431b0a4221e3a0b0cabb010e0a1/src/common/models/backend/transactions.rs#L54

We consider this to be optional.

Uxio0:
Thanks @fmrsabino , then we are good to merge

moisses89 (approved):

PR #1195, opened 2022-11-28
Refactor ERC20 indexing

Uxio0:
Don't filter by address on the RPC, it's not feasible with a high number of Safes. Get every event and filter them locally.

This should fix ERC20/721 indexing issues on networks with big number of Safes, like Polygon


Uxio0:
> Is this related to an issue or is a performance improve ?

Yes, sorry, I updated the PR description

moisses89 (approved):
Looks good to me!

PR #1198, opened 2022-11-29
Don't index events previous to a Safe creation

Uxio0:
- Previously, as we inherited the logic from the relay, it was possible that an ERC20 token was sent to a Safe address before the Safe was created, so by default we were indexing one day before was indexed.
- When a Safe was created the service could be marked out of sync even if it's not
- We were spending resources for a very rare case
- We can always index manually if needed
- Set v4.12.1 for a hotfix


moisses89 (approved):
Agree, Should be a weird case.

mikhailxyz (approved):
Vamos! 🚀

PR #1200, opened 2022-11-30
reduce erc20 endpoint

Uxio0:
- Improve logs for EthereumIndexer
- Increase ERC20 indexing information cache time 15s -> 5m
- Set version v4.12.3


moisses89 (approved):

moisses89:
Maybe we should review if the new logs are useful for the future.

PR #1202, opened 2022-12-01
Refactor ERC20 indexer block status storage

Uxio0:
Until now, `erc20_block_number` indexed status was kept for every `SafeContract`. Updating this table was not very performant with a big number of Safes.

# Solution

- Updates psycopg back to v2.9.5, as it was downgraded just to test if db was having issues due to that.
- Remove `erc20_block_number` from `SafeContract` table. Instead, a new table `IndexingStatus` was created storing one erc20 `block_number` for every Safe.
- Add admin manager for `IndexingStatus`
- Update services and tests to use `IndexingStatus`
- Test migrations


safe_transaction_service/history/services/index_service.py
```diff
@@ -160,26 +153,19 @@ def is_service_synced(self) -> bool:
...
-            # Ignore less than 10% of contracts out of sync
-            if proportion_out_of_sync >= self.alert_out_of_sync_events_threshold:
-                logger.error(
-                    "%d Safe Contracts have ERC20/721 out of sync",
-                    out_of_sync_contracts,
-                )
-                synced = False
+        if (
+            self.erc20_indexer_storage.get_last_indexed_block_number()
+            < reference_block_number
```
moisses89, line 165: What do you think if instead check if we are in the last block, check if we are close to the last block.
I mean, log error if the last indexed_block is less than last_block - 500 for example.

Uxio0: @moisses89 We already do that, check how the `reference_block_number is calculated`

safe_transaction_service/history/indexers/erc20_events_indexer.py
```diff
@@ -173,3 +182,40 @@ def process_elements(
             return range(
                 result_erc20 + result_erc721
             )  # TODO Hack to prevent returning `TokenTransfer` and using too much RAM
+
+    def get_almost_updated_addresses(
+        self, current_block_number: int
```
moisses89, line 187: `current_block_number` is not used in this function.

safe_transaction_service/history/indexers/erc20_events_indexer.py
```diff
@@ -173,3 +182,40 @@ def process_elements(
...
+    ) -> QuerySet[MonitoredAddress]:
+        """
+
+        :param current_block_number:
+        :return: Monitored addresses to be processed
+        """
+
+        logger.debug("%s: Retrieving monitored addresses", self.__class__.__name__)
+
+        addresses = self.database_queryset.all()
```
moisses89, line 197: The database_queryset function is :
```
    def database_queryset(self):
        return SafeContract.objects.all()
```
then here database_queryset.all() makes sense?

safe_transaction_service/history/indexers/erc20_events_indexer.py
```diff
@@ -173,3 +182,40 @@ def process_elements(
...
+        return addresses
+
+    def get_not_updated_addresses(
+        self, current_block_number: int
+    ) -> QuerySet[MonitoredAddress]:
+        """
+        :param current_block_number:
+        :return: Monitored addresses to be processed
+        """
+        return []
```
moisses89, line 209: Return always an empty array?

moisses89 (changes requested):

safe_transaction_service/history/indexers/erc20_events_indexer.py
Uxio0: Yes, but's inheriting from the parent class and must define it 😉

safe_transaction_service/history/indexers/erc20_events_indexer.py
Uxio0: Yes, as `all()` forces reevaluating

safe_transaction_service/history/indexers/erc20_events_indexer.py
Uxio0: Yes, there are not `not updated addresses` anymore

moisses89: I could see that is ok, you are overwritting a method of EthereumIndexer :)

safe_transaction_service/history/models.py
```diff
@@ -107,6 +117,10 @@ def parse(tx_type: str):
             raise ValueError(f"{tx_type} is not a valid InternalTxType")
 
 
+class IndexingStatusType(Enum):
+    ERC20_721_EVENTS = 0
```
moisses89, line 121: Whis is the purpose of this Enum with only one option? Thinking to include `masterCopy` indexing in this table also?

Uxio0: Maybe in the future, yes 😉

safe_transaction_service/history/tests/test_erc20_events_indexer.py
```diff
@@ -20,25 +20,26 @@ def test_erc20_events_indexer(self):
...
         self.assertFalse(
             ERC20Transfer.objects.tokens_used_by_address(safe_contract.address)
         )
 
         self.assertEqual(erc20_events_indexer.start(), 1)
-        safe_contract.refresh_from_db()
+
+        # Erc20/721 last indexed block number is stored on IndexingStatus
+        self.assertGreater(
+            IndexingStatus.objects.get_erc20_721_indexing_status().block_number, 0
```
moisses89, line 38: What do you think to test `get_erc20_721_indexing_status()` with `indexingStatus` empty to return the value of `masterCopy` indexing block?

safe_transaction_service/history/tests/test_migrations.py
```diff
@@ -105,3 +130,89 @@ def test_migration_backward_0068(self):
...
+            erc20_block_number=4,
+            ethereum_tx=ethereum_tx,
+        )
+        SafeContract.objects.create(
+            address=Account.create().address,
+            erc20_block_number=15,
+            ethereum_tx=ethereum_tx,
+        )
+        new_state = self.migrator.apply_tested_migration(
+            ("history", "0069_indexingstatus_and_more"),
```
moisses89, line 159: The `_and_more` in the migration name is because remove `erc20_block_number` ?

safe_transaction_service/history/tests/test_migrations.py
```diff
@@ -105,3 +130,89 @@ def test_migration_backward_0068(self):
...
+        )
+        IndexingStatus = new_state.apps.get_model("history", "IndexingStatus")
+        self.assertEqual(IndexingStatus.objects.get().block_number, 4)
```
moisses89, line 162: What do you think about to include a case without `SafeContract` where the `block_number` should be equals to 0?

safe_transaction_service/history/tests/test_internal_tx_indexer.py
```diff
@@ -304,16 +305,15 @@ def test_tx_processor_using_internal_tx_indexer_with_existing_safe(self):
...
             InternalTxDecoded.objects.pending_for_safes()
         )
-        safe_contract: SafeContract = SafeContract.objects.first()
-        self.assertGreater(safe_contract.erc20_block_number, 0)
-        safe_contract.erc20_block_number = 0
-        safe_contract.save(update_fields=["erc20_block_number"])
+        self.assertEqual(
+            IndexingStatus.objects.get_erc20_721_indexing_status().block_number, 0
+        )
+        IndexingStatus.objects.set_erc20_721_indexing_status(0)
```
moisses89, line 311: `IndexingStatus.objects.get_erc20_721_indexing_status().block_number` if indexingStatus is not created then is writting the `min(master_copy_block_number)` and if this last value doesn't exist is writting 0, then why are you setting 0 once again?

safe_transaction_service/history/models.py
```diff
@@ -151,6 +165,49 @@ def bulk_create_from_generator(
                 return total
 
 
+class IndexingStatusManager(models.Manager):
+    def get_erc20_721_indexing_status(self) -> "IndexingStatus":
+        try:
+            return self.get(indexing_type=IndexingStatusType.ERC20_721_EVENTS.value)
+        except self.model.DoesNotExist:
+            # Start indexing on master copies minimum block number
+            min_master_copies_block_number = (
```
moisses89, line 174: I only want be sure that I'm understanding this correctly. This should be an initial state where still not ERC20/ERC721 transactions, that means we should start looking for them from the minimal `masterCopy` block indexed?

safe_transaction_service/history/views.py
```diff
@@ -167,7 +167,7 @@ class IndexingView(GenericAPIView):
     serializer_class = serializers.IndexingStatusSerializer
     pagination_class = None  # Don't show limit/offset in swagger
 
-    @method_decorator(cache_page(5 * 60))  # 5 minutes
+    @method_decorator(cache_page(15))  # 15 seconds
     def get(self, request):
```
moisses89, line 171: I'd be defensive on this, 15 seconds seems to low, maybe 2 or 3 minutes?

safe_transaction_service/history/tests/test_migrations.py
Uxio0: Yes, it was an automatic name from Django, I didn't rename it

safe_transaction_service/history/tests/test_migrations.py
Uxio0: This is already tested at the beginning of `test_migration_backward_0069`

safe_transaction_service/history/models.py
Uxio0: That's right, for people setting up the service from scratch. I realise I have to do it on the migration too to follow the same logic, or maybe I can just remove it from here as the migration will guarantee field always exists 👍🏻

safe_transaction_service/history/views.py
Uxio0: Why? In this case database queries are very fast and it gives us more information about what's going on on the service

safe_transaction_service/history/models.py
Uxio0: Done

PR #1209, opened 2022-12-05
Move origin validiation to validate function

moisses89:
Related with https://github.com/safe-global/safe-transaction-service/pull/1177#discussion_r1035941612

### DIFF
--- a/safe_transaction_service/history/serializers.py
+++ b/safe_transaction_service/history/serializers.py
@@ -231,18 +231,20 @@ class SafeMultisigTransactionSerializer(SafeMultisigTxSerializerV1):
                 f"Calculated owners={signature_owners}"
             )
 
+        # Origin field on db is a JsonField
+        if attrs["origin"]:
+            try:
+                attrs["origin"] = json.loads(attrs["origin"])
+            except ValueError:
+                pass
+        else:
+            attrs["origin"] = {}
+
         return attrs
 
     def save(self, **kwargs):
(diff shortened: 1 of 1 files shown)


Uxio0 (changes requested):
I would prefer a separate `validate_origin` function

moisses89:
> I would prefer a separate `validate_origin` function
I didn't in a separated function because is going to be called only in one place. I mean, is not being reusable.

Uxio0:
> I didn't in a separated function because is going to be called only in one place. I mean, is not being reusable.

I'm not taking about clean code 😉 , I'm talking about how Django REST serializers work: https://www.django-rest-framework.org/api-guide/serializers/#field-level-validation . We should only add validations in `validate` when they depend on another fields, else they should go in their own functions so validation is more clear

moisses89:
> > I didn't in a separated function because is going to be called only in one place. I mean, is not being reusable.
> 
> I'm not taking about clean code wink , I'm talking about how Django REST serializers work: https://www.django-rest-framework.org/api-guide/serializers/#field-level-validation . We should only add validations in `validate` (object level validation) when they depend on another fields, else they should go in their own functions so validation is more clear (field level validation)

Done [here](https://github.com/safe-global/safe-transaction-service/pull/1209/commits/5021503abb441218b6506d4a552f7ff61b2ad796)

Uxio0 (approved):

PR #1222, opened 2022-12-12
Don't use Binance API for price fetching

### LINKED ISSUE #1221: Error when using Binance API
**Describe the bug**
- This bug **is breaking the builds due to E2E tests, so a fix is urgent**
- Binance API is returning errors sometimes on the CI

```
Service unavailable from a restricted location according to 'b. Eligibility' in https://www.binance.com/en/terms. Please contact customer service if you believe you received this message in error.
```

**To Reproduce**
Run the price tests multiple times until it pops.

**Proposed solution**
- Get and configure an API key
- Stop using Binance API, there are already enough oracles.

Uxio0:
- Closes #1221

### DIFF
--- a/safe_transaction_service/tokens/clients/binance_client.py
+++ b/safe_transaction_service/tokens/clients/binance_client.py
@@ -7,7 +7,7 @@ from .exceptions import CannotGetPrice
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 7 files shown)


moisses89 (approved):
Looks good to me!

PR #1224, opened 2022-12-12
Don't run CI on PRs twice

Uxio0:
- For every push on a PR two jobs were launched on the CI

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -1,5 +1,10 @@
 name: Python CI
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+      - develop
+  pull_request:
 
 jobs:
   linting:


.github/workflows/python.yml
```diff
@@ -1,5 +1,12 @@
 name: Python CI
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
```
fmrsabino, line 6: We had this in the past on the config-service/cgw and the issue with this was that you'd not get CI feedback for pushed branches until you've opened a PR.

Not sure if you want this or not 😄

.github/workflows/python.yml
```diff
@@ -1,5 +1,12 @@
...
+  release:
+    types:
+      - released
```
fmrsabino, line 9: Why add this type?

.github/workflows/python.yml
Uxio0: I don't think that's the normal flow for this service as every branch usually is linked to a PR, only exceptions being master and develop (I forgot about develop and just added it). Thanks for the tip

.github/workflows/python.yml
Uxio0: I don't want to trigger the CI on other release events that are not `released` (like draft). We do the same in other repos (like config service) and it's working well

fmrsabino: I see. I asked because this is not mentioned: it was about not running the CI on PRs twice.

.github/workflows/python.yml
Uxio0: I get you @fmrsabino . Do you think that removing `pull_request` and only run CI on `push` would work? That way it will always run for PRs and for custom branches

.github/workflows/python.yml
Uxio0: You are right, I will mention it on the PR description

Uxio0: You are right, maybe release it's not needed for now, I will add it in a new PR if that's required

fmrsabino: Up to you :) I'm fine approving this as it is. Was just trying to understand the context.

.github/workflows/python.yml
fmrsabino: We had some issues with that in the past because if you remove `pull_request` then no action runs for forks – https://github.com/safe-global/safe-config-service/pull/404

Maybe there's a better approach to this.

For reference here are some other PRs that tried experimented with different setups:
- https://github.com/safe-global/safe-config-service/pull/5
- https://github.com/safe-global/safe-config-service/pull/25

Given this I think `pull_request` needs to stay. But the current solution that you propose also works (we just need to take into consideration that the CI would only run when opening a PR bus as long as we are fine with that then I think it should be good to go 👍 )

.github/workflows/python.yml
Uxio0: Removed for now

.github/workflows/python.yml
Uxio0: Ok, then, makes sense, thanks for the info. I've just updated the PR

fmrsabino (approved):

PR #1226, opened 2022-12-13
Fix block range auto adjust for ERC20 indexer

Uxio0:
- Block range for ERC20 was not adjusted, so it was only decreasing on exceptions but never increasing again


moisses89 (approved):
Looks good to me!

PR #1235, opened 2022-12-19
Add release event to github actions

Uxio0:
- If not no event will be triggered on release and docker image will not be generated


Uxio0:
Just realised trying to build a new image for the relay service

Uxio0:
I need this to release the hotfix

PR #1237, opened 2022-12-19
Prevent stuck Safes after a reorg

### LINKED ISSUE #1236: After a reorg Safes can get stuck
**Describe the bug**
After a reorg, Safes get stuck

**To Reproduce**
- Manually call `reprocess_addresses` on the `IndexService`
- Safe gets stuck and traces/events are not processed-

**Expected behavior**
Safe traces/events are processed after a reorg

**Additional context**
This is happening due to a double error:
- When recovering from a reorg, only `SafeStatus` is emptied, and `SafeLastStatus` is not: https://github.com/safe-global/safe-transaction-service/blob/master/safe_transaction_service/history/services/index_service.py#L322
- `SafeLastStatus` reports corrupted if `SafeStatus` table is empty: https://github.com/safe-global/safe-transaction-service/blob/master/safe_transaction_service/history/tasks.py#L349

Uxio0:
- Closes #1236

### DIFF
--- a/safe_transaction_service/history/services/index_service.py
+++ b/safe_transaction_service/history/services/index_service.py
@@ -17,6 +17,7 @@ from ..models import (
     ModuleTransaction,
     MultisigConfirmation,
     MultisigTransaction,
+    SafeLastStatus,
     SafeMasterCopy,
     SafeStatus,
 )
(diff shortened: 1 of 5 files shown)


safe_transaction_service/history/services/index_service.py
```diff
@@ -320,12 +321,17 @@ def _reprocess(self, addresses: List[str]):
...
         queryset = SafeStatus.objects.all()
         if addresses:
             queryset = queryset.filter(address__in=addresses)
         queryset.delete()
 
+        logger.info("Remove Safe Last statuses")
+        queryset = SafeLastStatus.objects.all()
+        if addresses:
+            queryset = queryset.filter(address__in=addresses)
+        queryset.delete()
```
moisses89, line 333: This is intended? here you can remove all lastStatus rows.

Uxio0: Same as for the other elements, right?

moisses89: yes :)

moisses89 (approved):

PR #1239, opened 2022-12-19
Fix request 1 block when block_process_limit is 1

### LINKED ISSUE #1238: block_process_limit = 1 query for two blocks instead of one
**Describe the bug**
When block_process_limit is 1, only one block should be requested to the node, but two blocks are requested.
The request parameter `to-block` is equal `from-block` + `block_process_limit` but to request one block should be `to-block` = `from-block` 

**Expected behavior**
Request only one block.

**Possible solution** 
`to-block` = `from-block` + `block_process_limit` - 1

moisses89:
Closes #1238 
Also included a fix to possible infinity loop on indexer that happen when `block_process_limit` is less than `blocks_to_reindex_again`.  
This shouldn't happen frequently because `block_process_limit` normally is increasing an reducing the value but could happen.

### DIFF
--- a/safe_transaction_service/history/indexers/ethereum_indexer.py
+++ b/safe_transaction_service/history/indexers/ethereum_indexer.py
@@ -164,6 +164,11 @@ class EthereumIndexer(ABC):
         :param current_block_number:
         :return: Top block number to process
         """
+        if self.block_process_limit == 1:
+            return min(
+                from_block_number,
+                current_block_number - self.confirmations,
+            )
         return min(
             from_block_number + self.block_process_limit,
             current_block_number - self.confirmations,
--- a/safe_transaction_service/history/indexers/ethereum_indexer.py
+++ b/safe_transaction_service/history/indexers/ethereum_indexer.py
@@ -145,8 +145,16 @@ class EthereumIndexer(ABC):
         if (from_block_number + self.block_process_limit) >= (
             current_block_number - self.confirmations
         ):
+            reindex_blocks = self.blocks_to_reindex_again
+            if self.blocks_to_reindex_again >= self.block_process_limit:
+                # The following formula ensure that get_to_block_number reach last block
+                reindex_blocks = (
+                    (self.block_process_limit + from_block_number)
+                    - current_block_number
+                ) - 1
+
             # Reindex again when it's almost synced to prevent reorg/missing elements issues
-            from_block_number = max(from_block_number - self.blocks_to_reindex_again, 0)
+            from_block_number = max(from_block_number - reindex_blocks, 0)
 
         if (current_block_number - common_minimum_block_number) <= self.confirmations:
             return  # We don't want problems with reorgs
@@ -164,13 +172,8 @@ class EthereumIndexer(ABC):
         :param current_block_number:
         :return: Top block number to process
         """
-        if self.block_process_limit == 1:
-            return min(
-                from_block_number,
-                current_block_number - self.confirmations,
-            )
         return min(
-            from_block_number + self.block_process_limit,
+            from_block_number + self.block_process_limit - 1,
             current_block_number - self.confirmations,
         )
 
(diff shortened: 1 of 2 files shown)


safe_transaction_service/history/indexers/ethereum_indexer.py
```diff
@@ -164,6 +164,11 @@ def get_to_block_number(
         :param current_block_number:
         :return: Top block number to process
         """
+        if self.block_process_limit == 1:
```
Uxio0, line 167: I would propose removing this and just:

```
 return min(
            from_block_number + self.block_process_limit - 1,
            current_block_number - self.confirmations,
)
```

Uxio0: Oh i get it, then we will never index the last block. Could you add a test for this so we don't break it in the future?

Uxio0 (changes requested):
This implementation will break the indexer, as indexer will not increase the `to_block_number` and will keep looping in the same block.

We should refactor the code where is needed so `get_to_block_number` can be:

```
 return min(
            from_block_number + self.block_process_limit - 1,
            current_block_number - self.confirmations,
)
```

I added a test that should pass

moisses89:
> This implementation will break the indexer, as indexer will not increase the `to_block_number` and will keep looping in the same block.
> 
> We should refactor the code where is needed so `get_to_block_number` can be:
> 
> ```
>  return min(
>             from_block_number + self.block_process_limit - 1,
>             current_block_number - self.confirmations,
> )
> ```
> 
> I added a test that should pass

I think that there is an issue in this part of code:
```
    def get_block_numbers_for_search(
        self, addresses: Sequence[str], current_block_number: Optional[int] = None
    ) -> Optional[Sequence[Tuple[int, int]]]:
        """
        :param addresses:
        :param current_block_number: To prevent fetching it again
        :return: Minimum common `from_block_number` and `to_block_number` for search of relevant `tx hashes`
        """
        current_block_number = (
            current_block_number or self.ethereum_client.current_block_number
        )

        common_minimum_block_number = self.get_minimum_block_number(addresses)
        if common_minimum_block_number is None:  # Empty queryset
            return None

        from_block_number = common_minimum_block_number + 1
        if (from_block_number + self.block_process_limit) >= (
            current_block_number - self.confirmations
        ):
            # Reindex again when it's almost synced to prevent reorg/missing elements issues
            from_block_number = max(from_block_number - self.blocks_to_reindex_again, 0)

        if (current_block_number - common_minimum_block_number) <= self.confirmations:
            return  # We don't want problems with reorgs

        to_block_number = self.get_to_block_number(
            from_block_number, current_block_number
        )
        return from_block_number, to_block_number
```
For example imagine that we define:
```
block_process_limit = 10
blocks_to_reindex = 10
current_block_number = 19
```
The first cycle go from block 1 to block 10.
Second cycle: 
- common_minimum_block_number = 10 
- from_block_number at this point is 11 because `common_minimum_block_number (10) + 1 `
- because from_block_number (11) + block_process_limit (10) > current_block_number (19) then 
  from_block_number is equals to 1 at this point.
- then get_to_block_number is 10 again because  `from_block_number(1) + block_process_limit (10)  -1 `
And again second cycle starts in 1 and ends in 10...
This is not related with `get_to_block_number` because would happen the same if  `blocks_to_reindex = 11` and with the old implementation. 
Then this happen always that `get_to_block_number` is equals to  `blocks_to_reindex`

safe_transaction_service/history/indexers/ethereum_indexer.py
```diff
@@ -145,8 +145,16 @@ def get_block_numbers_for_search(
         if (from_block_number + self.block_process_limit) >= (
             current_block_number - self.confirmations
         ):
+            reindex_blocks = self.blocks_to_reindex_again
+            if self.blocks_to_reindex_again >= self.block_process_limit:
```
Uxio0, line 149: Please add tests for this

Uxio0 (changes requested):

safe_transaction_service/history/indexers/ethereum_indexer.py
Uxio0: Also I'm a little worried that if `block_process_limit` is too tight we can miss reindexing some transactions, but I think there's no easy solution for that

moisses89:
Currently we are not having a lot of issues we the reindex blocks and I'm not feeling confident with this PR, but we should try to find a way to down to one block in the indexer.

PR #1241, opened 2022-12-21
Process `InternalTxDecoded` in batches

Uxio0:
- An iterator was used to process InternalDecodedTx. Either all were processed, or none was be (atomic update)
- Refactor method so we process InternalTxDecoded in batches, so even if they take a lot to process service can slowly process them all
- This issue was detected on production with the indexer stuck for days processing a Safe with more than 11k of InternalTxDecoded

### DIFF
--- a/safe_transaction_service/history/tasks.py
+++ b/safe_transaction_service/history/tasks.py
@@ -6,6 +6,8 @@ from functools import cache
 from typing import Any, Dict, Optional
 from urllib.parse import urlparse
 
+from django.conf import settings
+
 import requests
 from celery import app
 from celery.utils.log import get_task_logger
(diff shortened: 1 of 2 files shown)


safe_transaction_service/history/tasks.py
```diff
@@ -395,12 +397,19 @@ def process_decoded_internal_txs_for_safe_task(
...
-            internal_txs_decoded = InternalTxDecoded.objects.pending_for_safe(
+            internal_txs_decoded_queryset = InternalTxDecoded.objects.pending_for_safe(
                 safe_address
-            ).iterator()
-            number_processed = len(
-                tx_processor.process_decoded_transactions(internal_txs_decoded)
             )
+            number_processed = 0
+            internal_txs_decoded_paginator = Paginator(
+                internal_txs_decoded_queryset, 500
```
rmeissner, line 405: Should we make this configurable?

rmeissner (approved):

safe_transaction_service/history/tasks.py
```diff
@@ -395,12 +397,19 @@ def process_decoded_internal_txs_for_safe_task(
...
             # Use iterator for memory issues
-            internal_txs_decoded = InternalTxDecoded.objects.pending_for_safe(
+            internal_txs_decoded_queryset = InternalTxDecoded.objects.pending_for_safe(
                 safe_address
-            ).iterator()
-            number_processed = len(
-                tx_processor.process_decoded_transactions(internal_txs_decoded)
             )
+            number_processed = 0
+            internal_txs_decoded_paginator = Paginator(
```
fmrsabino, line 404: You can set a chunk size for the iterator. Wouldn't that be sufficient?

Uxio0: No, as it will try to process all the elements together

safe_transaction_service/history/tasks.py
Uxio0: Good call, I made it in the last commit

fmrsabino (approved):

PR #1242, opened 2022-12-21
Split tracing calls for TxProcessor and SafeService

Uxio0:
- Currently if tracing was enabled every RPC call from this service was performed to the tracing node, even if not required
- With this update, only tracing specific calls will be performed to the tracing node

### DIFF
--- a/safe_transaction_service/history/indexers/tx_processor.py
+++ b/safe_transaction_service/history/indexers/tx_processor.py
@@ -10,7 +10,7 @@ from hexbytes import HexBytes
 from packaging.version import Version
 from web3 import Web3
 
-from gnosis.eth import EthereumClient
+from gnosis.eth import EthereumClient, EthereumClientProvider
 from gnosis.eth.constants import NULL_ADDRESS
 from gnosis.eth.contracts import get_safe_V1_0_0_contract, get_safe_V1_3_0_contract
 from gnosis.safe import SafeTx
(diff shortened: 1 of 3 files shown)


safe_transaction_service/history/indexers/tx_processor.py
```diff
@@ -82,9 +83,14 @@ class SafeTxProcessor(TxProcessor):
     Processor for txs on Safe Contracts v0.0.1 - v1.0.0
     """
 
-    def __init__(self, ethereum_client: EthereumClient):
+    def __init__(
+        self,
+        ethereum_client: EthereumClient,
+        ethereum_tracing_client: Optional[EthereumClient],
```
rmeissner, line 89: Some docs documenting what happens if it is not set would be nice

safe_transaction_service/history/services/safe_service.py
```diff
@@ -64,9 +64,13 @@ def del_singleton(cls):
 
 
 class SafeService:
-    def __init__(self, ethereum_client: EthereumClient, tracing_enabled: bool):
+    def __init__(
+        self,
+        ethereum_client: EthereumClient,
+        ethereum_tracing_client: Optional[EthereumClient],
```
rmeissner, line 70: Same as above

safe_transaction_service/history/services/safe_service.py
```diff
@@ -220,10 +224,10 @@ def _get_next_internal_tx(self, internal_tx: InternalTx) -> Optional[InternalTx]
     def _get_parent_internal_tx(self, internal_tx: InternalTx) -> InternalTx:
         if parent_trace := internal_tx.get_parent():
             return parent_trace
-        if not self.tracing_enabled:
```
rmeissner, line 223: Was this field deprecated?

Uxio0: It's not required anymore, checking `if not self.ethereum_tracing_client` is enough

safe_transaction_service/history/indexers/tx_processor.py
Uxio0: Updated in a the new commit

PR #1270, opened 2023-01-16
Fix processing `ChangedMasterCopy` event

Uxio0:
`ChangedMasterCopy` event was supposed to generate an argument `_masterCopy` instead of `masterCopy`. Due to a typo, it was not.

### DIFF
--- a/safe_transaction_service/history/indexers/safe_events_indexer.py
+++ b/safe_transaction_service/history/indexers/safe_events_indexer.py
@@ -318,8 +318,8 @@ class SafeEventsIndexer(EventsIndexer):
             internal_tx_decoded = None
         elif event_name == "ChangedMasterCopy":
             internal_tx_decoded.function_name = "changeMasterCopy"
-            internal_tx.arguments = {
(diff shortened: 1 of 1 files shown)


safe_transaction_service/history/indexers/safe_events_indexer.py
```diff
@@ -318,8 +318,8 @@ def _process_decoded_element(
             internal_tx_decoded = None
         elif event_name == "ChangedMasterCopy":
             internal_tx_decoded.function_name = "changeMasterCopy"
-            internal_tx.arguments = {
-                "_masterCopy": args.get("singleton") or args.get("masterCopy")
+            internal_tx_decoded.arguments = {
+                "_masterCopy": args.get("masterCopy") or args.get("singleton")
```
moisses89, line 322: > ChangedMasterCopy event was supposed to generate an argument _masterCopy instead of masterCopy 

It is still masterCopy.

Uxio0: Yes but previously was not stored as `internal_tx` does not have an `arguments` field

PR #1275, opened 2023-01-18
Fix log message when reducing block range

Uxio0:
Due to updating the variable before logging, the log message `block_process_limit %d is bigger than block_process_limit_max %d, reducing` had the same value for `block_process_limit` and for `block_process_limit_max`


Uxio0:
You are right, done

PR #1282, opened 2023-01-20
Build ARM64 docker images

Uxio0:
Refactor docker image building:
  - Use buildx
  - Use github actions cache
  - Build image for AMD64 and ARM64 platforms
  - Remove deploy_docker.sh script

Fix typo on autodeploy, as it was called on tag pushes


.github/workflows/python.yml
```diff
@@ -96,33 +96,61 @@ jobs:
     needs:
       - linting
       - test-app
-    if: github.ref == 'refs/heads/master' || github.ref == 'refs/heads/develop' || startsWith(github.ref, 'refs/tags/')
+    if: github.ref == 'refs/heads/master' || github.ref == 'refs/heads/develop' || (github.event_name == 'release' && github.event.action == 'released')
     steps:
     - uses: actions/checkout@v3
+    - uses: docker/setup-qemu-action@v2
```
luarx, line 102: Should we specify Qemu VMs?
platforms -> Platforms to install (e.g. arm64,riscv64,arm ; default all)

https://github.com/docker/setup-qemu-action#customizing

Uxio0: You are right, maybe it's faster. Done

luarx (approved):

PR #1289, opened 2023-01-24
Add support for token lists

### LINKED ISSUE #1281: Periodically set trusted tokens using token list information
# What is needed?
- Tokens `trusted` flag must be set using token lists
- A task must be created to update them (once a day looks like a good compromise)

# Background
Currently `trusted` tokens are marked manually. Scraping trusted token lists it's a good way to automatically update them. We will hardcode the following lists:

| Network  | List |
| ------------- | ------------- |
| mainnet  | [Coingecko](https://tokenlists.org/token-list?url=https://tokens.coingecko.com/uniswap/all.json)  |
| Polygon  | [Official List](https://api-polygon-tokens.polygon.technology/tokenlists/polygonTokens.tokenlist.json) |
| Optimism | [Official List](https://static.optimism.io/optimism.tokenlist.json) |
| Gnosis Chain | [Honeyswap ](https://tokens.honeyswap.org/) |
| BSC | [PancakeSwap](https://tokens.pancakeswap.finance/pancakeswap-extended.json) |
| Aurora | [Official List](https://aurora.dev/tokens.json) |

# Endpoint
Endpoints will not be updated as it's only an internal change.

Uxio0:
- Closes #1281
- Add a new TokenList model holding token list url and description
- Add a task to automatic update `trusted` status for tokens from the token lists every day. Tokens not in any of the list will be marked as `not trusted`


safe_transaction_service/tokens/tasks.py
```diff
@@ -111,3 +113,33 @@ def fix_pool_tokens_task() -> Optional[int]:
...
+def update_token_info_from_token_list() -> int:
+    """
+    If there's at least one valid token list with at least 1 token, every token in the DB is marked as `not trusted`
+    and then every token on the is marked as `trusted`
+
+    :return: Number of tokens marked as `trusted`
+    """
+    tokens = []
+    for token_list in TokenList.objects.all():
+        tokens += token_list.get_tokens()
```
fmrsabino, line 129: If one list raises an exception we don't try the other lists right?

I think we should log it (as we are doing) and continue with the other lists.

safe_transaction_service/tokens/tasks.py
```diff
@@ -111,3 +113,33 @@ def fix_pool_tokens_task() -> Optional[int]:
...
+        return 0
+
+    # Make sure current chainId matches the one in the list
+    ethereum_network = get_ethereum_network()
+
+    token_addresses = [
+        fast_to_checksum_address(token["address"])
+        for token in tokens
+        if token["chainId"] == ethereum_network.value
+    ]
```
fmrsabino, line 141: I think it's better if we use a set here.

The reason being that the token lists can be potentially quite big. When we do a `filter(address__in=token_addresses)` I believe that the only way to get that query set would be that for every token _T_ we check if it's part of the of the token list with _N_ elements. So it'd iterate over a collection of _N_ (O_(T ** N)_) when we are just interested if it's part of the list or not (_O(T)_).

What do you think?

safe_transaction_service/tokens/tasks.py
```diff
@@ -111,3 +113,33 @@ def fix_pool_tokens_task() -> Optional[int]:
...
+
+    with transaction.atomic():
+        Token.objects.update(trusted=False)
+        return Token.objects.filter(address__in=token_addresses).update(trusted=True)
```
fmrsabino, line 145: Maybe instead of updating every `Token` we can update the ones where `address` is not in `token_addresses` to `False` and the other query set would update the other tokens to `True`. Maybe this can be done with query set exclusion.

This would result in less updates overall (items which are part of the list would get 1 update instead of 2).

fmrsabino (changes requested):

safe_transaction_service/tokens/tasks.py
Uxio0: Agree

Uxio0: Done

safe_transaction_service/tokens/tasks.py
Uxio0: > Maybe instead of updating every `Token` we can update the ones where `address` is not in `token_addresses` to `False` and the other query set would update the other tokens to `True`. Maybe this can be done with query set exclusion.

Then you will force a query for every token not on the list, which is not quite fast (usually the "not in" SQL queries are not the best). I wouldn't do this, I think updating elements on the list 2 times is more optimal

Even though I updated `Token.objects.update(trusted=False)` to `Token.objects.filter(trusted=True).update(trusted=False)` to speed up the query

fmrsabino: > > Maybe instead of updating every `Token` we can update the ones where `address` is not in `token_addresses` to `False` and the other query set would update the other tokens to `True`. Maybe this can be done with query set exclusion.
> 
> Then you will force a query for every token not on the list, which is not quite fast (usually the "not in" SQL queries are not the best). I wouldn't do this, I think updating elements on the list 2 times is more optimal
> 
> Even though I updated `Token.objects.update(trusted=False)` to `Token.objects.filter(trusted=True).update(trusted=False)` to speed up the query

What about filtering for tokens on the list (updating those) and then excluding that set and update those to `False`. It'd still use `NOT()` at the end though... 🤔

Anyways was mostly an alternative that I thought 👍

safe_transaction_service/tokens/tasks.py
Uxio0: I tested it with 100K addresses and query is instant, I thought about your solution but I think the current approach should work and be more optimal.

Also I took a look at:
- https://stackoverflow.com/questions/1009706/postgresql-max-number-of-parameters-in-in-clause
- https://stackoverflow.com/questions/4936731/maximum-length-of-an-sql-query

WDYT?

fmrsabino: Should be ok. I think that in the end it doesn't matter for how Django processes it. I took a look into:

https://github.com/django/django/blob/d8e1442ce2c56282785dd806e5c1147975e8c857/django/db/models/fields/related_lookups.py#L40-L62

And it seems that for this use-case it's serialised as a tuple so having a `list` or a `set` shouldn't matter.

But good to know! 😅

Uxio0: Thanks for bringing this up

fmrsabino (approved):

PR #1292, opened 2023-01-26
Delete older not trusted transactions

### LINKED ISSUE #1212: Not trusted transactions are stored forever on DB
# What is needed?
Not `trusted` transactions should be cleaned from database periodically.

# Background
- Everyone can propose transactions, no signatures are required.
- If transactions are not used, they will be kept forever on the database.

# Solution
- As usually if they are not signed/executed right away they will not be used in the future and also storing can be abused, I would propose to **run a daily task deleting not trusted transactions older than 30 days**.

Uxio0:
- Create a task to delete `not trusted` transactions older than one month
- Use `datetime.timedelta` instead of `timedelta` to prevent confusing `datetime` with `datetime.datetime`
- Closes #1212

### DIFF
--- a/safe_transaction_service/history/models.py
+++ b/safe_transaction_service/history/models.py
@@ -1197,6 +1197,9 @@ class MultisigTransactionQuerySet(models.QuerySet):
     def trusted(self):
         return self.filter(trusted=True)
 
+    def not_trusted(self):
+        return self.filter(trusted=False)
+
     def multisend(self):
         # TODO Use MultiSend.MULTISEND_ADDRESSES + MultiSend MULTISEND_CALL_ONLY_ADDRESSES
         return self.filter(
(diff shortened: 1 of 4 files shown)


safe_transaction_service/history/models.py
```diff
@@ -1197,6 +1197,9 @@ def without_confirmations(self):
     def trusted(self):
         return self.filter(trusted=True)
 
+    def not_trusted(self):
+        return self.filter(trusted=False)
```
moisses89, line 1201: should we include executed=False?

Uxio0: `executed` transactions should be always `trusted`: https://github.com/safe-global/safe-transaction-service/blob/master/safe_transaction_service/history/indexers/tx_processor.py#L602

Uxio0: `executed` transactions should always be  `trusted`: https://github.com/safe-global/safe-transaction-service/blob/master/safe_transaction_service/history/indexers/tx_processor.py#L602

hectorgomezv (approved):
LGTM 🙂

PR #1293, opened 2023-01-26
Check chainId to prevent misconfigurations

### LINKED ISSUE #1274: Check previously used network to prevent misconfiguring a node
# What is needed?
If network does not match the network already used to sync the service, prevent Celery workers from running.

# Background
If by mistake a node for a different network is configured for an already running service database might get corrupted by new indexed data from the new network.

Uxio0:
- Every time the service starts check the previous used chainId so there's no chance to have a misconfigured RPC
- Add a new DB model `Chain` to keep track of that
- Closes #1274


Uxio0:
@luarx do you think this solution works?

luarx:
I like it @Uxio0 !

moisses89 (approved):
Good check, looks good to me!

PR #1302, opened 2023-01-30
Don't send notifications for not trusted txs

Uxio0:


DmitryBespalov (approved):

moisses89 (approved):
Looks good to me!

Uxio0:
@fmrsabino @hectorgomezv do you think this can have be problematic for the client gateway? I can think on the lists not being refreshed

@DmitryBespalov Mobile team must also fix this on their side

hectorgomezv:
From my point of view, this change shouldn't be an issue on the Client Gateway side since an untrusted transaction shouldn't trigger a cache invalidation IMHO. In fact, we're explicitly sending a hardcoded `trusted=true` param for the  transaction queue retrieval in the Client Gateway new implementation 🤔

Uxio0:
Thanks for the input @hectorgomezv , good to know

PR #1303, opened 2023-01-30
hotfix

Uxio0:
- Don't send notifications for not trusted txs
- Set version v4.14.2


moisses89:
This hotfix was released https://github.com/safe-global/safe-transaction-service/releases/tag/v4.14.2

PR #1305, opened 2023-01-31
Decode base64/json ERC721 URIs

### LINKED ISSUE #1149: NFT image is not propagated to the service if encoded as data URI scheme
**Describe the bug**
If an NFT encodes the metadata as a data URI scheme which includes the image encoded as a data URI scheme as well, the image is not available in the interface.

**To Reproduce**
Steps to reproduce the behaviour:
1. Send a GET request to https://safe-transaction-goerli.safe.global/api/v1/safes/0xCDd6e79fBC90643dC59273500E3A408EfA24d57C/collectibles/?trusted=false&exclude_spam=false
2. The image URL points to a non-existing resource https://safe-transaction-assets.safe.global/tokens/logos/0xBe41D6b67D1796A7196f08f43165724E196C15F1.png)

**Expected behavior**dec
The image is parsed from the metadata and added to the service

 - Production
 - Goerli
 - OS: Symbian
 - Chrome

**Additional context**
Decoded token metadata from base64:
[decoded_metadata.txt](https://github.com/safe-global/safe-transaction-service/files/9927180/decoded_metadata.txt)

Data URI scheme RFC: https://datatracker.ietf.org/doc/html/rfc2397

Uxio0:
- Closes #1149


Uxio0:
Tests are breaking due to the Dappcon 2020 NFT uri temporary (I hope) error

moisses89 (approved):
Looks good, the image should be decoded by the client right?

Uxio0:
> Looks good, the image should be decoded by the client right?

Yes, I guess that should be valid html

Issue #1095, opened 2022-10-06
Log what oracle was used to fetch token price


Uxio0:
# What is needed?
Log what oracle was used to fetch token price

# Background
Currently we only see the token price but we don't know what oracle was used to fetch the price

PR #1316, opened 2023-02-06
Don't log price retrieval failures

Uxio0:
- Only log success, so logs get cleaner and more useful
- Closes https://github.com/safe-global/safe-transaction-service/issues/1095


moisses89 (approved):
This is closing also this issue? https://github.com/safe-global/safe-transaction-service/issues/1095

Uxio0:
> This is closing also this issue? #1095

Yes, I updated the issue to make it clear we are logging it, thanks for realizing

PR #1317, opened 2023-02-07
Refactor and fix reorg handling

Uxio0:
- Optimize reorg handling but using `get_blocks` to get 2 blocks at once instead of sequentially
- Increase reorg task frequency from 3 -> 1 minute
- Optimize db index for `confirmed` on `EthereumBlock`, as it's only needed for `not confirmed` blocks
- Fix bug on `ReorgService` where ERC20/721 indexing was not reset on a reorg
- Refactor `ReorgService` `reorg_models` to `reorg_functions`, easier to expand
- Fix `test_check_reorgs`, it was not correctly checking the result


safe_transaction_service/history/tasks.py
```diff
@@ -57,20 +57,18 @@
...
-                logger.warning(
-                    "Reorg found for block-number=%d", first_reorg_block_number
-                )
+            reorg_block_number = reorg_service.check_reorgs()
+            if reorg_block_number:
+                logger.warning("Reorg found for block-number=%d", reorg_block_number)
                 # Stopping running tasks is not possible with gevent
-                reorg_service.recover_from_reorg(first_reorg_block_number)
-                return first_reorg_block_number
+                reorg_service.recover_from_reorg(reorg_block_number)
```
moisses89, line 70: `recover_from_reorg` receive a parameter called `first_reorg_block_number`. Maybe we could keep `first_reorg_block_number` or `reorg_block_number` in both places.

safe_transaction_service/history/tests/test_reorg_service.py
```diff
@@ -23,26 +23,19 @@
...
 
-        def get_block_fn(number: int, full_transactions=False):
-            if number == block_number:
-                return block_child
-
-            if number == block_number + 1:
-                return block_parent
-
-        get_block_mock.side_effect = get_block_fn
+        get_blocks_mock.return_value = [block_child, block_parent]
```
moisses89, line 38: This is ok? get_blocks return a list of blocks, in this case we are querying 2 blocks, for example return block [1000, 1001] then the first one is the parent and the second one is the child. 
https://github.com/safe-global/safe-transaction-service/pull/1317/files#diff-b0f80c3ff20824ca54494683dd485a06d8cb04975c474db7aca1e4999ab9ef39R73

moisses89: Of course is ok! you can force a reorg :)

moisses89 (approved):
Looks good to me.

safe_transaction_service/history/tasks.py
Uxio0: You are right. Done, can you recheck?

PR #1344, opened 2023-02-22
Fix ERC20/721 indexing starting from 0

Uxio0:
- If service is setup from scratch, `ERC20/721 indexing block number` will start from `zero`, as there's no master copy in the DB when the migration runs
- Add a method in `setup_service` to update `ERC20/721 indexing block number` if it's lower than the oldest master copy deployed

### DIFF
--- a/safe_transaction_service/history/management/commands/setup_service.py
+++ b/safe_transaction_service/history/management/commands/setup_service.py
@@ -3,13 +3,14 @@ from typing import Dict, List, Sequence, Tuple
 
 from django.conf import settings
 from django.core.management.base import BaseCommand
+from django.db.models import Min
 
 from django_celery_beat.models import IntervalSchedule, PeriodicTask
 
 from gnosis.eth import EthereumClientProvider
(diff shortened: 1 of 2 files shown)


safe_transaction_service/history/management/commands/setup_service.py
```diff
@@ -645,3 +647,30 @@ def _setup_safe_proxy_factories(
                     "tx_block_number": initial_block_number,
                 },
             )
+
+    def _setup_erc20_indexing(self) -> bool:
+        """
+        Update ERC20/721 indexing status if `indexing block number` is less
+        than `Master copies` block deployments, as it sounds like a configuration error
+
+        :return: `True` if updated, `False` otherwise
```
moisses89, line 656: This is going to keep always ERC20 indexing behind MasterCopies indexing?

Uxio0: No, this is going to make sure that ERC20 indexing is never behind the first created master copy, as Safes cannot exist

moisses89 (approved):

Issue #1353, opened 2023-03-03
Duplicated webhook calls


Uxio0:
**Describe the bug**
Webhook calls are duplicated

**To Reproduce**
- Send a token to a Safe
- Webhook task will be called multiple times with the same information

**Expected behavior**
- One webhook call per unique event

**Additional context**
- Traces/events are always inserted in DB by ignoring errors (ignoring duplicates)
- That's because we have logics in place to reindex, sometimes it can find a missed transaction but usually is the same transactions all over again
- It's way more optimal to tell the database to insert everything and ignore already inserted transactions than checking
- When you insert data in that mode postresql does not return information about what it did, so all the inserted and not inserted elements will be treated as inserted
- Callback for sending webhooks is called again and a new webbook is sent. Not for notifications, as duplicated are handled and not sent (unless there's a reorg)

**Solution**
I'm working to find a solution without a big refactor, as it will increase a lot the performance of the tx service and also the client gateway (as for the current cgw invalidating caches in Redis is an expensive operation)

Uxio0:
Closed by https://github.com/safe-global/safe-transaction-service/issues/1353

PR #1354, opened 2023-03-03
Fix duplicated webhook calls

Uxio0:
- Related to #1353

Fix the issue by using a LRU cache that will ignore previously processed elements for every indexer, fixing webhooks bug and also increasing the performance of the service by reducing processing of events/traces and calls to PostgreSQL

- Add processed cache to ERC20/721 events indexer
- Remove not needed methods from ERC20 indexer
- Add processed cache to EventsIndexer
- Add processed cache to InternalTxIndexer


safe_transaction_service/history/indexers/erc20_events_indexer.py
```diff
@@ -47,12 +41,25 @@ def del_singleton(cls):
...
+
+    ERC20 Transfer Event: `Transfer(address indexed from, address indexed to, uint256 value)`
+    ERC721 Transfer Event: `Transfer(address indexed from, address indexed to, uint256 indexed tokenId)`
+
+    As `event topic` is the same both events can be indexed together, and then tell
+    apart based on the `indexed` part as `indexed` elements are stored in a different way in the
+    `ethereum tx receipt`.
+
+    Note: Some `ERC721` do not `index` the `tokenId`, so decimals are checked
+    in `_is_erc20` to tell apart `ERC20` from `ERC721` contracts (`ERC721` don't have decimals).
```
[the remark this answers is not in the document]
Uxio0: Yes, function is not used anymore so I must remove this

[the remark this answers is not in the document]
Uxio0: This is fixed

safe_transaction_service/history/tests/test_safe_events_indexer.py
```diff
@@ -602,3 +605,35 @@ def test_safe_events_indexer(self):
...
+        self.assertEqual(
+            len(self.safe_events_indexer.process_elements(safe_events_mock)), 0
+        )
+        self.assertEqual(
+            len(self.safe_events_indexer.process_elements(safe_events_mock)), 0
+        )
+
+        # Even if we empty the cache, events will not be reprocessed again
+        self.safe_events_indexer._processed_element_cache.clear()
+        self.assertEqual(
```
moisses89, line 637: Then we need the cache here?

safe_transaction_service/utils/utils.py
```diff
@@ -8,6 +8,24 @@
...
+
+    Dictionaries are guaranteed to be insertion sorted from Python 3.7 onwards
+    """
+
+    def __init__(self, *args, maxlen=0, **kwargs):
+        self._maxlen = maxlen
+        super().__init__(*args, **kwargs)
+
+    def __setitem__(self, key, value):
+        dict.__setitem__(self, key, value)
```
moisses89, line 23: This could be after the maxlen condition because here we are adding one element more than defined size for a while.

moisses89 (approved):

safe_transaction_service/history/tests/test_safe_events_indexer.py
```diff
@@ -602,3 +605,35 @@ def test_safe_events_indexer(self):
...
+
+    def test_mark_as_processed(self):
+        # SafeEventsIndexer does not use bulk saving into database,
+        # so mark_as_processed is just a optimization but not critical
+
+        # Create transaction in db so not fetching of transaction is needed
+        for safe_event in safe_events_mock:
+            tx_hash = safe_event["transactionHash"]
+            block_hash = safe_event["blockHash"]
+            if not EthereumTx.objects.filter(tx_hash=tx_hash).exists():
```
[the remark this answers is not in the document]
Uxio0: Some events belong to the same tx_hash, if you remove that line there's an error

safe_transaction_service/history/tests/test_safe_events_indexer.py
Uxio0: It saves processing an event and a not needed call to the database

safe_transaction_service/utils/utils.py
Uxio0: Good catch, but I don't think this is critical 😉

PR #1363, opened 2023-03-08
Refactor block range calculation for indexers

Uxio0:
- Store next block to index in database instead of last indexed block
- Allow to process only 1 block (until now 2 was the minimum)
- Allow to index the current block number if `confirmations=0`

### DIFF
--- a/safe_transaction_service/history/indexers/ethereum_indexer.py
+++ b/safe_transaction_service/history/indexers/ethereum_indexer.py
@@ -160,7 +160,6 @@ class EthereumIndexer(ABC):
             )
             from_block_number = max(from_block_number - blocks_to_reindex, 0)
 
-        print(from_block_number, to_block_number)
         return from_block_number, to_block_number
 
     def get_to_block_number(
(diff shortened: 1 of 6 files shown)


safe_transaction_service/history/indexers/ethereum_indexer.py
```diff
@@ -162,10 +168,11 @@ def get_to_block_number(
         """
         :param from_block_number:
         :param current_block_number:
+
         :return: Top block number to process
         """
         return min(
-            from_block_number + self.block_process_limit,
+            from_block_number + self.block_process_limit - 1,
```
moisses89, line 175: I don't get why minus 1?

Uxio0: No, minimum value for `block_process_limit` is `1`. `0` would be no limit but it cannot be used due to the RPC limits. So in this case imagine you want to process `1 block`, then `to and from block` will be the same

moisses89 (approved):

PR #1370, opened 2023-03-14
Remove not used ERC20Indexing endpoint

Uxio0:
- IndexingStatus endpoint provides ERC20 information

PR #1371, opened 2023-03-14
Fix Indexing Status endpoint

Uxio0:
- Prevent showing negative indexes on reorgs
- Fix issues introduced by https://github.com/safe-global/safe-transaction-service/pull/1363


safe_transaction_service/history/services/index_service.py
```diff
@@ -109,15 +109,27 @@ def block_get_or_create_from_block_hash(self, block_hash: int):
...
+            self.get_erc20_721_current_indexing_block_number() - 1, current_block_number
+        )
+
+        master_copies_block_number = (
+            self.get_master_copies_current_indexing_block_number()
+        )
+        master_copies_block_number = (
+            current_block_number
+            if master_copies_block_number is None
+            else min(master_copies_block_number - 1, current_block_number)
```
moisses89, line 131: Why check the min value? `master_copies_current_indexing` shouldn't be greater than `current_block` at least something wrong in the indexer. 
Also I would like to set only one time the value of `master_copies_block_number`

safe_transaction_service/history/views.py
```diff
@@ -146,21 +146,6 @@ def get(self, request, format=None):
             return Response(self._get_info(ethereum_client))
 
 
-class ERC20IndexingView(GenericAPIView):
```
moisses89, line 149: Can you rebase? and ping me to last review :) because this was removed here #1370

PR #1373, opened 2023-03-16
Ignore coingecko failing tests

Uxio0:
- Usually due to rate limits

### DIFF
--- a/safe_transaction_service/history/tests/utils.py
+++ b/safe_transaction_service/history/tests/utils.py
@@ -1,3 +1,4 @@
+import functools
 import os
 
 import pytest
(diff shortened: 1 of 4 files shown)


moisses89 (changes requested):

safe_transaction_service/history/services/index_service.py
```diff
@@ -119,17 +119,18 @@ def get_indexing_status(self) -> IndexingStatus:
 
         # Indexing points to the next block to be indexed, we need the previous ones
         erc20_block_number = min(
-            self.get_erc20_721_current_indexing_block_number() - 1, current_block_number
+            max(self.get_erc20_721_current_indexing_block_number() - 1, 0),
```
moisses89, line 122: This is not really related with the PR, could you add this change to the description?

moisses89 (approved):
Only mentions that I would like to have in the description the fix of erc20 indexing. 
Could we do at the same way here? https://github.com/safe-global/safe-transaction-service/blob/master/safe_transaction_service/history/tests/clients/test_ens_client.py#L29

safe_transaction_service/history/services/index_service.py
Uxio0: Done

Uxio0:
> Only mentions that I would like to have in the description the fix of erc20 indexing. 

Done!

> Could we do at the same way here? https://github.com/safe-global/safe-transaction-service/blob/master/safe_transaction_service/history/tests/clients/test_ens_client.py#L29

What do you mean?

PR #1392, opened 2023-03-30
Set to 0 block difficulty mock

### LINKED ISSUE #1059: Changes for "The Merge"
# What is needed?
Do certain changes after **The Merge**:
- Rename `latest` RPC calls to `unsafe` (or maybe `safe`)
 
# Background
https://blog.infura.io/post/what-you-need-to-know-to-get-ready-for-the-ethereum-merge-infura-developers-edition

moisses89:
### What was wrong? 👾
After ethereum merge difficulty is always 0.

Closes #1059 

### How was it fixed? 🎯
This PR update difficulty to the mock files to be more realistic.

### DIFF
--- a/safe_transaction_service/history/tests/mocks/mocks_internal_tx_indexer.py
+++ b/safe_transaction_service/history/tests/mocks/mocks_internal_tx_indexer.py
@@ -613,7 +613,7 @@ trace_transactions_result = [
 
 block_child = {
     "author": "0x0000000000000000000000000000000000000000",
-    "difficulty": 1,
+    "difficulty": 0,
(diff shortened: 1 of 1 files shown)


Uxio0:
I wouldn't do this. IIRC those mocks were taken from Rinkeby or other networks. We shouldn't modify real blocks as tests count on the blocks to be real. For example, if in the future we introduce a method to calculate the hash of a block it will break with our mocks, or if we try to use the mocks to check stuff in the chain, similar to what we do at `safe-eth-py` when comparing the expected traces with the ones returned from the nodes

PR #1410, opened 2023-04-13
Fix USD prices for tokens

### LINKED ISSUE #1407: Get token price on USD return ETH value on Gnosis Chain
**Describe the bug**
The endpoint GET /api/v1/tokens/0x177127622c4A00F3d409B75571e12cB3c8973d3c/prices/usd/ return the ETH value instead in USD on gnosis chain. 

**To Reproduce**
Steps to reproduce the behavior with COW token:
1. Do GET on https://safe-transaction-gnosis-chain.safe.global/api/v1/tokens/0x177127622c4A00F3d409B75571e12cB3c8973d3c/prices/usd/
    Return 
```
    {
     "fiatCode": "USD",
     "fiatPrice": "4.396770390106545e-05",
     "timestamp": "2023-04-12T13:38:27.098915Z"
   }
```
2. Check on cow swap and compare that the value is in ETH instead USD
image (4)

**Expected behavior**
Return USD price

** Aditional context ** 
The conversion is done using https://github.com/safe-global/safe-transaction-service/blob/master/safe_transaction_service/tokens/services/price_service.py#L419 native token price that in case of gnosis chain is XDA=0.99$ with the token price on ETH.

### LINKED ISSUE #1408: Get balances return 503 error for some safes
**Describe the bug**
Get the balances on transaction service for some safes returns a 503 error, the issue could be related with safes with long list of balances like for example more than 108 tokens. 

**To Reproduce**
1. Do GET /api/v1/safes/{safe-address}/balances/?exclude_spam=true
 Then the service response is:
 Status: 503
 message:  `{"exception":"Problem connecting to Ethereum network","trace":"NodeConnectionException"}`

**Expected behavior**
The service should return the list of tokens also for long lists.

 - production
 - Which chain? mainnet

Uxio0:
- `Token/ETH` value was retrieved in every network, but then the `Native token/USD` value was used to calculate the token `USD` value for every network.
- For example, for Gnosis Chain `Token/ETH` value was retrieved and then `DAI/USD` value was used to turn it into `USD`, instead of `ETH/USD`.
- Closes #1407

### DIFF
--- a/safe_transaction_service/tokens/services/price_service.py
+++ b/safe_transaction_service/tokens/services/price_service.py
@@ -3,13 +3,13 @@ from dataclasses import dataclass
 from datetime import datetime
 from enum import Enum
 from functools import cached_property
+from logging import getLogger
 from typing import Iterator, List, Optional, Sequence, Tuple
 
 from django.utils import timezone
 
 from cache_memoize import cache_memoize
 from cachetools import TTLCache, cachedmethod
-from celery.utils.log import get_task_logger
 from eth_typing import ChecksumAddress
 from redis import Redis
 
@@ -42,7 +42,7 @@ from safe_transaction_service.utils.redis import get_redis
 from ..clients import CannotGetPrice, CoingeckoClient, KrakenClient, KucoinClient
 from ..tasks import EthValueWithTimestamp, calculate_token_eth_price_task
 
-logger = get_task_logger(__name__)
+logger = getLogger(__name__)
 
 
 class FiatCode(Enum):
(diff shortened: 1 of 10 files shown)


safe_transaction_service/tokens/services/price_service.py
```diff
@@ -194,14 +197,26 @@ def get_kcs_usd_price(self) -> float:
         except CannotGetPrice:
             return self.coingecko_client.get_kcs_usd_price()
 
-    @cachedmethod(cache=operator.attrgetter("cache_eth_price"))
+    @cachedmethod(cache=operator.attrgetter("cache_ether_usd_price"))
     @cache_memoize(60 * 30, prefix="balances-get_eth_usd_price")  # 30 minutes
+    def get_ether_usd_price(self) -> float:
```
moisses89, line 202: Could you add a test for ether USD like we do for native coin?

Uxio0: This method is already tested on the native coin one, but sounds good, I will

Uxio0: Done

moisses89 (approved):
Let's test it on staging.

PR #1412, opened 2023-04-13
Fix USD prices

Uxio0:
- Refactor Coingecko Client
- Revert PriceService to use native USD coin


moisses89 (approved):

Uxio0:
It requires a new release of `safe-eth-py`

PR #1424, opened 2023-04-20
Add crontab schedule for periodic tasks

### LINKED ISSUE #1406: Celery scheduler inaccurate scheduling for periodic tasks
**Describe the bug**
Some tasks must run periodically, but they run in a irregular schedule. We see that a lot on the Polygon indexer. For example, "Reindexing task" should run every 110 minutes, but:

Celery scheduler issue

**Expected behavior**
Tasks must run as scheduled.

moisses89:
# Description
The purpose of this PR is try to improve the accuracy of our current beat scheduler using crontab definitions.
See https://docs.celeryq.dev/en/stable/userguide/periodic-tasks.html#crontab-schedules
# Issues related
Closes #1406 

# Other changes
Also because we can get more accuracy when the task is executed the `reindex_last_hours_task` was splitted in `reindex_last_hours_mastercopies_task` defeined every 2 hours at minute 0 and `reindex_last_hours_erc20_task` defined every 2 hours minute 30.

### DIFF
--- a/safe_transaction_service/history/management/commands/setup_service.py
+++ b/safe_transaction_service/history/management/commands/setup_service.py
@@ -5,7 +5,7 @@ from django.conf import settings
 from django.core.management.base import BaseCommand
 from django.db.models import Min
 
-from django_celery_beat.models import IntervalSchedule, PeriodicTask
+from django_celery_beat.models import CrontabSchedule, IntervalSchedule, PeriodicTask
 
 from gnosis.eth import EthereumClientProvider
 from gnosis.safe.addresses import MASTER_COPIES, PROXY_FACTORIES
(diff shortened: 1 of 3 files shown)


safe_transaction_service/history/management/commands/setup_service.py
```diff
@@ -5,37 +5,67 @@
...
     enabled: bool = True
 
     def create_task(self) -> Tuple[PeriodicTask, bool]:
-        interval_schedule, _ = IntervalSchedule.objects.get_or_create(
-            every=self.interval, period=self.period
-        )
-        periodic_task, created = PeriodicTask.objects.get_or_create(
-            task=self.name,
-            defaults={
+        if self.cron is None:
```
Uxio0, line 35: I think it's better to check what you are adding and not that the other attribute is not defined
```suggestion
        assert self.period or self.cron, "Task must define period or cron"
        if self.period:
```

safe_transaction_service/history/management/commands/setup_service.py
```diff
@@ -5,37 +5,67 @@
...
+            }
+
+        periodic_task, created = PeriodicTask.objects.get_or_create(
+            task=self.name,
+            defaults=defaults,
         )
         if not created:
             periodic_task.name = self.description
-            periodic_task.interval = interval_schedule
+            if self.cron is None:
```
Uxio0, line 64: Same that the previous comment, if you are checking for an interval check if the interval exists, don't asume there's an interval just because there's no cron

safe_transaction_service/history/tasks.py
```diff
@@ -270,7 +270,7 @@ def process_decoded_internal_txs_task(self) -> Optional[int]:
 
 
 @app.shared_task(bind=True, soft_time_limit=SOFT_TIMEOUT, time_limit=LOCK_TIMEOUT)
-def reindex_last_hours_task(self, hours: int = 2) -> Optional[int]:
+def reindex_last_hours_mastercopies_task(self, hours: int = 2) -> Optional[int]:
```
Uxio0, line 273: I think name is a little off (my fault(), as we are renaming it maybe `reindex_mastercopies_last_hours_task` would be better

safe_transaction_service/history/tasks.py
```diff
@@ -292,16 +292,32 @@ def reindex_last_hours_task(self, hours: int = 2) -> Optional[int]:
                         to_block_number,
                     )
                     reindex_master_copies_task.delay(from_block_number, to_block_number)
+
+
+@app.shared_task(bind=True, soft_time_limit=SOFT_TIMEOUT, time_limit=LOCK_TIMEOUT)
+def reindex_last_hours_erc20_task(self, hours: int = 2) -> Optional[int]:
```
Uxio0, line 298: `reindex_erc20_erc721_last_hours_task`

Uxio0 (changes requested):

safe_transaction_service/history/tasks.py
moisses89: Sounds good

Uxio0 (approved):

PR #1434, opened 2023-04-26
Allow multiple ABIs per topic for decoding

### LINKED ISSUE #1388: Support events with same topic but different indexed parameters on `EventIndexer`
# What is needed?
Support events with same topic but different indexed parameters on `EventIndexer`

# Background
Currently, `EventIndexer` matches an `Event` with a `topic` to decode logs, but for some `Events` structure is different due to `indexed` parameters even if topic is the same, e.g: 
- ERC20: `event Transfer(address indexed from, address indexed to, uint256 value);`. It stores `value` on `data` inside the ethereum `log` structure.
- ERC721: `event Transfer(address indexed from, address indexed to, uint256 indexed tokenId);`. It stores `value` on `topics` inside the ethereum `log` structure.

Service would need to be refactored so https://github.com/safe-global/safe-transaction-service/blob/master/safe_transaction_service/history/indexers/events_indexer.py#L99 returns `Dict[bytes, List[ContractEvent]]` and for decoding the `list` is iterated to find the matching event.

# Related issues
- https://github.com/safe-global/safe-contracts/pull/542

Uxio0:
- Support events with same topic but different indexed parameters on EventIndexer
- Closes #1388

### DIFF
--- a/safe_transaction_service/history/indexers/events_indexer.py
+++ b/safe_transaction_service/history/indexers/events_indexer.py
@@ -92,15 +92,23 @@ class EventsIndexer(EthereumIndexer):
     @abstractmethod
     def contract_events(self) -> List[ContractEvent]:
         """
-        :return: Web3 ContractEvent to listen to
+        :return: List of Web3.py `ContractEvent` to listen to
         """
 
     @cached_property
-    def events_to_listen(self) -> Dict[bytes, ContractEvent]:
-        return {
-            HexBytes(event_abi_to_log_topic(event.abi)).hex(): event
-            for event in self.contract_events
-        }
+    def events_to_listen(self) -> Dict[bytes, List[ContractEvent]]:
+        """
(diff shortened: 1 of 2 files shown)


moisses89 (commented):
Looks good to me, but only two questions:
the idea is try every decode way for each topic, how long could be the array for one topic? maybe 2, 3 ways, not 100? :) 
and my second question is if we could to add tests for test decoding using contracts 1.4.0.

Uxio0:
> Looks good to me, but only two questions: the idea is try every decode way for each topic, how long could be the array for one topic?

In this case 2, v1.3.0 version and v1.4.0 with changed indexed. In practice I doubt we change the indexed parameters again so it should be the maximum.

> second question is if we could to add tests for test decoding using contracts 1.4.0.

There's still no release for v1.4.0 so I will add them on the v1.4.0 PR

moisses89 (approved):
Looks good to me!

PR #1442, opened 2023-05-02
Make `check_index_problems` auto reindex Safes

Uxio0:
`check_index_problems` tries to find issues with Safes and detect the first problematic transaction, so it was easy to refactor it to try to reindex missing transactions and not having to do that manually after


moisses89 (approved):
Looks good to me! then this command is going to fix by default the indexing issues? 
If it works we could handle a task on one day of week to do it automatically, what do you think?

PR #1443, opened 2023-05-02
Return `ModuleTxs` with same `txHash` in all-transactions

### LINKED ISSUE #1372: Indexing issue in case of batching module txs in the nested safe
**Describe the bug**
Indexing issue in case of batching module txs in the nested safe

**To Reproduce**
1. 3 module txs created in the "parent safe" were confirmed and executed as a batched tx in the "owner safe"  
 https://safe-transaction-gnosis-chain.safe.global/api/v1/safes/0x00B0CbB97F5509492E33caCAcf0F4f88235FcFbA/module-transactions/
As a result, they are identical, only the "to" field is different for all 3 txs ( "transactionHash": "0x8f9acc2f904d673cbeeea27eb47da87dc9718f5e963ad0ff7952be630fa37536")

2. the txs are indexed as one tx with an incorrect value ( 30xdai instead of 90 xdai) - https://safe-transaction-gnosis-chain.safe.global/api/v1/safes/0x00B0CbB97F5509492E33caCAcf0F4f88235FcFbA/all-transactions/?executed=false&queued=true&trusted=true
image

**Expected behavior**
The txs should be returned as 3 module txs with a valid value

 - production
 - gnosis chain
 - Browser chrome/MM on mobile
 - Version "name": "Safe Transaction Service",
                  "version": "4.16.0",

Uxio0:
- `all-transactions` endpoint was returning only 1 module transactions when more than one `ModuleTransaction` had the same `txHash`
- More than one `ModuleTransaction` can be executed in a single transaction, so we refactor `getTransaction` method to be able to return multiple transactions for the same identifier
- Renames some variables and methods as `identifier` is more correct than `hash`
- Closes #1372

### DIFF
--- a/safe_transaction_service/history/services/transaction_service.py
+++ b/safe_transaction_service/history/services/transaction_service.py
@@ -8,6 +8,7 @@ from typing import Any, Dict, List, Optional, Sequence, Tuple, Union
 from django.db.models import Case, Exists, F, OuterRef, QuerySet, Subquery, Value, When
 from django.utils import timezone
 
+from eth_typing import HexStr
 from redis import Redis
 
 from gnosis.eth import EthereumClient, EthereumClientProvider
@@ -59,15 +60,15 @@ class TransactionService:
         copyreg.pickle(memoryview, lambda val: (memoryview, (bytes(val),)))
 
     #  Cache methods ---------------------------------
-    def get_cache_key(self, safe_address: str, tx_hash: str):
-        return f"tx-service:{safe_address}:{tx_hash}"
+    def get_cache_key(self, safe_address: str, tx_id: str):
+        return f"tx-service:{safe_address}:{tx_id}"
 
     def get_txs_from_cache(
-        self, safe_address: str, hashes_to_search: Sequence[str]
+        self, safe_address: str, ids_to_search: Sequence[str]
     ) -> List[Union[EthereumTx, MultisigTransaction, ModuleTransaction]]:
         keys_to_search = [
-            self.get_cache_key(safe_address, hash_to_search)
-            for hash_to_search in hashes_to_search
+            self.get_cache_key(safe_address, id_to_search)
+            for id_to_search in ids_to_search
         ]
         return [
             pickle.loads(data) if data else None
@@ -77,8 +78,8 @@ class TransactionService:
     def store_txs_in_cache(
         self,
         safe_address: str,
-        hashes_with_txs: Tuple[
-            str, Union[EthereumTx, MultisigTransaction, ModuleTransaction]
+        ids_with_txs: Tuple[
+            str, List[Union[EthereumTx, MultisigTransaction, ModuleTransaction]]
         ],
     ):
         """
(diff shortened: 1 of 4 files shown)


safe_transaction_service/history/services/transaction_service.py
```diff
@@ -276,57 +280,60 @@ def get_all_tx_hashes(
...
             for ethereum_tx in EthereumTx.objects.filter(
-                tx_hash__in=hashes_not_cached
+                tx_hash__in=id_not_cached
             ).select_related("block")
         }
 
         # We also need the in/out transfers for the MultisigTxs
-        all_hashes = hashes_not_cached + [
-            multisig_tx.ethereum_tx_id for multisig_tx in multisig_txs.values()
+        all_hashes = id_not_cached + [
```
hectorgomezv, line 333: Nitpick: Should this be renamed as well? (I mean, `all_ids/all_identifiers` instead of `all_hashes`?)

safe_transaction_service/history/services/transaction_service.py
```diff
@@ -276,57 +280,60 @@ def get_all_tx_hashes(
...
+                ids_to_search,
+                self.get_txs_from_cache(safe_address, ids_to_search),
             )
             if cached_tx
         }
-        hashes_not_cached = [
+        id_not_cached = [
             hash_to_search
-            for hash_to_search in hashes_to_search
+            for hash_to_search in ids_to_search
```
hectorgomezv, line 303: Another nitpick: do you think `hash_to_search` should be renamed to `id_to_search`?

safe_transaction_service/history/services/transaction_service.py
```diff
@@ -366,25 +373,36 @@ def get_all_txs_from_hashes(
...
+                    ]
                 return result
 
-            if result := plain_ethereum_txs.get(transaction_id):
+            if result := id_with_plain_ethereum_txs.get(transaction_id):
                 # If no Multisig or Module tx found, fallback to simple tx
-                result.transfers = transfer_dict[result.tx_hash]
+                for ethereum_tx in result:
+                    # Populate transfers
+                    ethereum_tx.transfers = transfer_dict[ethereum_tx.tx_hash]
```
hectorgomezv, line 405: I have one (maybe silly) doubt here: why are for loops needed for non-module transactions? According to the PR description it seems only module transactions identifiers could have several transactions associated.

hectorgomezv (changes requested):
Good job, thanks for fixing this! I've just some minor concerns 🙂

safe_transaction_service/history/services/transaction_service.py
```diff
@@ -276,57 +280,60 @@ def get_all_tx_hashes(
         # the same `execution_date` that the mined tx
         return queryset
 
-    def get_all_txs_from_hashes(
-        self, safe_address: str, hashes_to_search: Sequence[str]
+    def get_all_txs_from_identififers(
```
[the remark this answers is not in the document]
Uxio0: Done!

safe_transaction_service/history/services/transaction_service.py
Uxio0: Done!

safe_transaction_service/history/services/transaction_service.py
Uxio0: Done! Thank you for all the suggestions @hectorgomezv

safe_transaction_service/history/services/transaction_service.py
Uxio0: They are the only ones, but I refactored the dictionaries to all look the same and be `id -> List[Tx]`. I thought that would be more clear than having mixed types and custom logics, and would enable us to allow for example multiple transfers in the same hash or some future features

hectorgomezv: Got it, thanks for the explanation @Uxio0 🙂

hectorgomezv (approved):
🚀

safe_transaction_service/history/services/transaction_service.py
Uxio0: Thank you for the detailed review 😉

safe_transaction_service/history/views.py
```diff
@@ -243,7 +243,7 @@ def list(self, request, *args, **kwargs):
             return self.get_paginated_response([])
 
         all_tx_hashes = [element["safe_tx_hash"] for element in page]
```
moisses89, line 245: I haven't strong opinion about this, but maybe `all_safe_tx_hashes` or `all_tx_identifiers`?

Uxio0: Done and also added a comment

moisses89 (approved):

PR #1445, opened 2023-05-04
Fix ERC20 indexing for CELO

Uxio0:
- More context: https://github.com/celo-org/celo-blockchain/discussions/2079
- The added check should only be relevant for CELO and not interfere with other chains


Uxio0:
We need to wait for more context before merging this, as I would prefer not to do custom chain adjustments on the tx service

rmeissner (approved):

moisses89 (approved):

PR #1446, opened 2023-05-05
Fix ERC20 get_balances HTTP 413

Uxio0:
- Add configurable batch requests for token balances

### DIFF
--- a/safe_transaction_service/history/services/balance_service.py
+++ b/safe_transaction_service/history/services/balance_service.py
@@ -4,6 +4,7 @@ from dataclasses import dataclass
 from datetime import datetime
 from typing import List, Optional, Sequence
 
+from django.conf import settings
 from django.core.cache import cache as django_cache
 from django.db.models import Q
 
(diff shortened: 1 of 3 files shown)


safe_transaction_service/history/services/balance_service.py
```diff
@@ -210,9 +212,22 @@ def _get_balances(
...
+            # With a lot of addresses an HTTP 413 error will be raised
+            for erc20_addresses_chunk in chunks(
+                erc20_addresses, settings.TOKENS_ERC20_GET_BALANCES_BATCH
+            ):
+                balances = self.ethereum_client.erc20.get_balances(
+                    safe_address, erc20_addresses_chunk
+                )
+
+                # Skip ether transfer if already there
+                raw_balances.extend(balances[1:] if raw_balances else balances)
```
moisses89, line 225: Maybe we should consider a flag to ignore ether balances on get_balances if the flag is true, what do you think? In my opinion is weird that `erc20.get_balances` return also the ether balance.
We could do it in other PR.

moisses89 (approved):
Only a comment about `get_balances`. :)

safe_transaction_service/history/services/balance_service.py
Uxio0: Sure, I was thinking the same, but we have to do it in other PR as the logic is on `safe-eth-py`, I thought it was a lot to take care on this PR

PR #1456, opened 2023-05-08
Fix transfer_id validator

### LINKED ISSUE #1454: Status code 500 returned on /transfer/{id} endpoint when modifying the id
**Describe the bug**
On certain scenarios it is possible to get a status code 500 when an invalid transfer id is used for this endpoint.

**To Reproduce**
1. Via the /transfers endpoint get a valid transferId for a tx like this one: `e27e15ba8dea473d98c80a8e76b45d372c0f3c6f8c184177044c935c37eb419d7216 `
2. Edit it to `e27e15ba8dea473d98c80a6b45d372c0f3c6f8c184177044c935c37eb419d7216` and try to fetch that transaction via /transfer/{id} endpoint 

**Expected behavior**
Status code should be 404 and not 500

**Environment:**
 - Staging or production? Staging
 - Which chain? Polygon

### LINKED ISSUE #1455: Error message on status 400 for /module-tx/{txId} endpoint
**Describe the bug**
When an invalid moduleTxId is used for this endpoint, the message shown is: "module_transaction_id is too short". We should change it into a more generic "module_transaction_id is not valid" also aligning with the behaviour for /transfer/{transferId} endpoint. 

Note: IDs that are too long will also throw the too short error so the more generic the better.

**To Reproduce**
1. On /module-tx/{txId} endpoint input a valid ID and then delete the first char. Example failing ID for polygon: `8216c222bb58b72a870d5f0b378cc96eeaa9f2c936daf65eb6185adf19bde49581 `

**Expected behavior**
Change "module_transaction_id is too short" error message to a more generic "module_transaction_id is not valid". 

**Environment:**
 - Staging or production? Staging
 - Which chain? Any (polygon for the example on this ticket)

moisses89:
Closes #1455 
Closes #1454 

### What was wrong? 👾
Token transfer id should always finish with log_index value and internal_tx transfer_id should always finish with trace_address. 

### How was fixed? 
Improved `is_valid_unique_transfer_id` to check if `transfer_id` contains valid `log_index` when is token transfer or valid `trace_address` when is an internal_tx.

### DIFF
--- a/safe_transaction_service/history/helpers.py
+++ b/safe_transaction_service/history/helpers.py
@@ -56,12 +56,18 @@ class DelegateSignatureHelper:
 
 def is_valid_unique_transfer_id(unique_transfer_id: str) -> bool:
     """
-    Check if transfer_id starts with 'e' or 'i' followed by keccak256 and appended by optional digits separated by commas
+    Check if transfer_id starts with 'e' or 'i' followed by keccak256 and ended by digits or digits separated by commas
 
     :param unique_transfer_id:
-    :return: True in case valide unique_transfer_id or False in other case
+    :return: ``True`` for a valid ``unique_transfer_id``, ``False`` otherwise
     """
-    return re.match(r"^(e|i)[a-fA-F0-9]{64}(\d+(,\d+)*)?$", unique_transfer_id)
+    token_transfer_id_pattern = r"^(e)([a-fA-F0-9]{64})(\d+)"
+    internal_transfer_id_pattern = r"^(i)([a-fA-F0-9]{64})(\d+)(,\d+)*"
+
+    return bool(
+        re.fullmatch(token_transfer_id_pattern, unique_transfer_id)
+        or re.fullmatch(internal_transfer_id_pattern, unique_transfer_id)
+    )
 
 
 def add_tokens_to_transfers(transfers: TransferDict) -> TransferDict:
--- a/safe_transaction_service/history/tests/test_views.py
+++ b/safe_transaction_service/history/tests/test_views.py
@@ -480,7 +480,7 @@ class TestViews(SafeTestCaseMixin, APITestCase):
         self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
 
         no_exist_module_transaction_id = (
-            "ief060441f0101ab83d62066b962f97e3a582686e0720157407c965c5946c2f7a"
+            "ief060441f0101ab83d62066b962f97e3a582686e0720157407c965c5946c2f7a0"
         )
         url = reverse(
             "v1:history:module-transaction", args=(no_exist_module_transaction_id,)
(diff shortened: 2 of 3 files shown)


Uxio0 (changes requested):

safe_transaction_service/history/helpers.py
```diff
@@ -56,12 +56,27 @@ def calculate_all_possible_hashes(cls, delegate: ChecksumAddress) -> List[bytes]
...
-    Check if transfer_id starts with 'e' or 'i' followed by keccak256 and appended by optional digits separated by commas
+    Check if transfer_id starts with 'e' or 'i' followed by keccak256 and appended by digits or digits separated by commas
 
     :param unique_transfer_id:
-    :return: True in case valide unique_transfer_id or False in other case
+    :return: True in case valid unique_transfer_id or False in other case
     """
-    return re.match(r"^(e|i)[a-fA-F0-9]{64}(\d+(,\d+)*)?$", unique_transfer_id)
+    is_valid_hash = re.compile(r"[a-fA-F0-9]")
+    is_valid_transfer_id = re.compile(r"^\d+(,\d+)*$")
```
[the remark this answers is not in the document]
moisses89: Done!

safe_transaction_service/history/helpers.py
```diff
@@ -56,12 +56,20 @@ def calculate_all_possible_hashes(cls, delegate: ChecksumAddress) -> List[bytes]
...
-    return re.match(r"^(e|i)[a-fA-F0-9]{64}(\d+(,\d+)*)?$", unique_transfer_id)
+    token_transfer_id_pattern = r"^(e)([a-fA-F0-9]{64})(\d+)$"
+    internal_transfer_id_pattern = r"^(i)([a-fA-F0-9]{64})(\d+)(,\d+)*$"
+
+    if bool(re.match(token_transfer_id_pattern, unique_transfer_id)) or bool(
+        re.match(internal_transfer_id_pattern, unique_transfer_id)
+    ):
+        return True
+
+    return False
```
Uxio0, line 72:
```suggestion
    return bool(re.match(token_transfer_id_pattern, unique_transfer_id) or re.match(internal_transfer_id_pattern, unique_transfer_id))):
```

No need to cast to `bool` when using `if`, `if` does it automatically, but in this case using `return` is better

Btw, shouldn't we be using `fullmatch`? https://docs.python.org/3/library/re.html#re.Pattern.fullmatch

Uxio0 (changes requested):

safe_transaction_service/history/helpers.py
```diff
@@ -56,12 +56,18 @@ def calculate_all_possible_hashes(cls, delegate: ChecksumAddress) -> List[bytes]
 
 def is_valid_unique_transfer_id(unique_transfer_id: str) -> bool:
     """
-    Check if transfer_id starts with 'e' or 'i' followed by keccak256 and appended by optional digits separated by commas
+    Check if transfer_id starts with 'e' or 'i' followed by keccak256 and appended by digits or digits separated by commas
```
Uxio0, line 59: Ended instead of appended I think it's more clear

Uxio0 (approved):

safe_transaction_service/history/helpers.py
moisses89: Sounds good

PR #1465, opened 2023-05-15
Send events to queue

### LINKED ISSUE #1007: Refactor webhooks to use a queue system
# What is needed?
Don't send webhooks from the transaction service. Send them to a queue, so services can subscribe and get the needed webhooks.

# Background
- Current webhook implementation requires the tx service to POST to the configured servers.
- Some servers can take some time to answer, blocking the tx service (a quick solution could be reducing the `timeout`).
- Some third party services would like to integrate with our webhooks, but that could impact the performance of the tx service.

# Proposed solution
- Send webhooks to a queue (rabbitmq, kafka...)
- A new service (webhook service) reads the webhooks and sends them to 3rd party clients
- From that queue, CGW can also read the webhooks

# Solution
- The queue chosen is RabbitMq
- The queues are going to be created by the safe-events-service
- The publisher is going to publish to an exchange in `FanOut` mode 
- The publisher class should be a singleton class that keeps only one connection open per process
- One RabbitMq for all chains

# Schema
Queues-rabbit-mq(2)

moisses89:
Closes #1007 
- add class to manage asynchronous connection (`gevent`) with `rabbitmq`.
- add class to manage synchronous connection for testing purposes. 
- add mock class to don't connect when there is not `rabbitmq` host configured.
- add singleton provider class to keep only one connection opened.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -95,7 +95,7 @@ jobs:
         ETHEREUM_TRACING_NODE_URL: http://localhost:8545
         ETH_HASH_BACKEND: pysha3
         REDIS_URL: redis://localhost:6379/0
-        EVENTS_QUEUE_URL_HOST: amqp://guest:guest@localhost:5672/
+        EVENTS_QUEUE_URL: amqp://guest:guest@localhost:5672/
     - name: Send results to coveralls
       continue-on-error: true  # Ignore coveralls problems
       run: coveralls --service=github
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -95,7 +95,7 @@ jobs:
         ETHEREUM_TRACING_NODE_URL: http://localhost:8545
         ETH_HASH_BACKEND: pysha3
         REDIS_URL: redis://localhost:6379/0
-        EVENTS_QUEUE_HOST: localhost
+        EVENTS_QUEUE_URL_HOST: amqp://guest:guest@localhost:5672/
     - name: Send results to coveralls
       continue-on-error: true  # Ignore coveralls problems
       run: coveralls --service=github
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -53,6 +53,15 @@ jobs:
           --health-retries 5
         ports:
           - 5432:5432
+      rabbitmq:
+        image: rabbitmq:alpine
+        options: >-
+          --health-cmd "rabbitmqctl await_startup"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - "5672:5672"
     steps:
     - name: Setup and run ganache
       run: |
@@ -86,6 +95,7 @@ jobs:
         ETHEREUM_TRACING_NODE_URL: http://localhost:8545
         ETH_HASH_BACKEND: pysha3
         REDIS_URL: redis://localhost:6379/0
+        EVENTS_QUEUE_HOST: localhost
     - name: Send results to coveralls
       continue-on-error: true  # Ignore coveralls problems
       run: coveralls --service=github
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -496,10 +496,7 @@ ALERT_OUT_OF_SYNC_EVENTS_THRESHOLD = env.float(
 
 # Events
 # ------------------------------------------------------------------------------
-EVENTS_QUEUE_HOST = env("EVENTS_QUEUE_HOST", default=None)
-EVENTS_QUEUE_PORT = env("EVENTS_QUEUE_PORT", default=5672)
-EVENTS_QUEUE_USERNAME = env("EVENTS_QUEUE_USERNAME", default="guest")
-EVENTS_QUEUE_PASSWORD = env("EVENTS_QUEUE_PASSWORD", default="guest")
+EVENTS_QUEUE_URL_HOST = env("EVENTS_QUEUE_URL_HOST", default=None)
 EVENTS_QUEUE_ASYNC_CONNECTION = env("EVENTS_QUEUE_ASYNC_CONNECTION", default=False)
 EVENTS_QUEUE_RABBITMQ_EXCHANGE_NAME = env(
     "EVENTS_QUEUE_RABBITMQ_EXCHANGE_NAME", default="amq.fanout"
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -99,6 +99,7 @@ LOCAL_APPS = [
     "safe_transaction_service.notifications.apps.NotificationsConfig",
     "safe_transaction_service.safe_messages.apps.SafeMessagesConfig",
     "safe_transaction_service.tokens.apps.TokensConfig",
+    "safe_transaction_service.events.apps.EventsConfig",
 ]
 # https://docs.djangoproject.com/en/dev/ref/settings/#installed-apps
 INSTALLED_APPS = DJANGO_APPS + THIRD_PARTY_APPS + LOCAL_APPS
@@ -240,6 +241,10 @@ CELERY_ROUTES = (
         ),
         (
             "safe_transaction_service.history.tasks.send_webhook_task",
+            {"queue": "webhooks", "delivery_mode": "transient"},
+        ),
+        (
+            "safe_transaction_service.events.tasks.send_event_to_queue_task",
             {"queue": "webhooks", "delivery_mode": "transient"},
         ),
         (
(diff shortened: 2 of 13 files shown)


config/settings/test.py
```diff
@@ -47,3 +47,5 @@
         "level": "DEBUG",
     }
 }
+
+EVENTS_QUEUE_ASYNC_CONNECTION = False
```
moisses89, line 51: Needed to force to use blockingconnection when testing as we can test with geventconnection

safe_transaction_service/history/signals.py
```diff
@@ -154,19 +156,22 @@ def process_webhook(
...
-                )
-                send_webhook_task.apply_async(
-                    args=(address, payload), priority=2  # Almost lowest priority
-                )  # Almost the lowest priority
-                send_notification_task.apply_async(
-                    args=(address, payload),
-                    countdown=5,
-                    priority=2,  # Almost lowest priority
-                )
+                if mark_notification_as_processed(address, payload):
```
moisses89, line 159: @Uxio0 I move the check if payload is repeated here, at this way we could avoid repeated webhooks and repeated notifications, but maybe is in notifications because some reason that I missing.

safe_transaction_service/events/services/queue_service.py
```diff
@@ -0,0 +1,279 @@
...
+    EXCHANGE_NAME = settings.EVENTS_QUEUE_RABBITMQ_EXCHANGE_NAME
+
+    def __init__(self):
+        self._channel = None
+        self._connection: GeventConnection = None
+        self.unsent_events = []
+        credentials = PlainCredentials(
+            settings.EVENTS_QUEUE_USERNAME, settings.EVENTS_QUEUE_PASSWORD
+        )
+        self._connection_parameters = ConnectionParameters(
```
Uxio0, line 45: I think using URL parameters is way more flexible. I would only leave password out due to security reasons: https://pika.readthedocs.io/en/stable/modules/parameters.html#urlparameters

moisses89: Done!

safe_transaction_service/history/signals.py
Uxio0: I think it's a good idea but I would not change it on this PR, as it can break the current behaviour. But if you already refactored the tests don't worry and we will merge it like this

moisses89: I did not change the tests, but the tests still works. 
I'm concerned about repeated events in the queue.
If we don't want to change the code here, maybe I can call `is_duplicated` method before to call `send_notification` (because send_notification is setting the key) If I do this after `send_notification`  then the event will be send never.

moisses89: Anyway, we can handle it in a different PR. 
@Uxio0 I removed the commit that included the mentioned changes.

.github/workflows/python.yml
```diff
@@ -86,6 +95,7 @@ jobs:
         ETHEREUM_TRACING_NODE_URL: http://localhost:8545
         ETH_HASH_BACKEND: pysha3
         REDIS_URL: redis://localhost:6379/0
+        EVENTS_QUEUE_URL_HOST: amqp://guest:guest@localhost:5672/
```
Uxio0, line 98: EVENTS_QUEUE_URL_HOST -> EVENTS_QUEUE_URL

config/settings/base.py
```diff
@@ -489,6 +494,13 @@
     "ALERT_OUT_OF_SYNC_EVENTS_THRESHOLD", default=0.1
 )  # Percentage of Safes allowed to be out of sync without alerting. By default 10%
 
+# Events
+# ------------------------------------------------------------------------------
+EVENTS_QUEUE_URL_HOST = env("EVENTS_QUEUE_URL_HOST", default=None)
+EVENTS_QUEUE_ASYNC_CONNECTION = env("EVENTS_QUEUE_ASYNC_CONNECTION", default=False)
+EVENTS_QUEUE_RABBITMQ_EXCHANGE_NAME = env(
```
Uxio0, line 501: EVENTS_QUEUE_RABBITMQ_EXCHANGE_NAME -> EVENTS_QUEUE_EXCHANGE_NAME

run_tests.sh
```diff
@@ -4,9 +4,9 @@ set -euo pipefail
 
 export DJANGO_SETTINGS_MODULE=config.settings.test
 export DJANGO_DOT_ENV_FILE=.env.test
-docker compose -f docker-compose.yml -f docker-compose.dev.yml build --force-rm db redis ganache
-docker compose -f docker-compose.yml -f docker-compose.dev.yml up --no-start db redis ganache
-docker compose -f docker-compose.yml -f docker-compose.dev.yml start db redis ganache
+docker compose -f docker-compose.yml -f docker-compose.dev.yml build --force-rm db redis ganache rabbitmq
+docker compose -f docker-compose.yml -f docker-compose.dev.yml up --no-start db redis ganache rabbitmq
+docker compose -f docker-compose.yml -f docker-compose.dev.yml start db redis ganache rabbitmq
```
Uxio0, line 9: Maybe it's time to add healthcheck to docker compose now that's supported and remove the sleep: https://docs.docker.com/compose/compose-file/05-services/#healthcheck (we can do it in another PR)

safe_transaction_service/events/services/queue_service.py
```diff
@@ -0,0 +1,272 @@
...
+    def __new__(cls):
+        if not hasattr(cls, "instance"):
+            if settings.EVENTS_QUEUE_URL_HOST:
+                if settings.EVENTS_QUEUE_ASYNC_CONNECTION:
+                    cls.instance = AsyncQueueService()
+                else:
+                    cls.instance = SyncQueueService()
+            else:
+                # Mock send_event to not configured host us is not mandatory configure a queue for events
+                cls.instance = MockedQueueService()
```
Uxio0, line 26: Log a warning if a `Mocked` service is used

safe_transaction_service/events/services/queue_service.py
```diff
@@ -0,0 +1,272 @@
...
+        return cls.instance
+
+    @classmethod
+    def del_singleton(cls):
+        if hasattr(cls, "instance"):
+            del cls.instance
+
+
+class QueueService:
+    EXCHANGE_NAME = settings.EVENTS_QUEUE_RABBITMQ_EXCHANGE_NAME
```
Uxio0, line 36: Set this inside `__init__` (or pass it as a parameter in __init__ and send it on the provider)

safe_transaction_service/events/services/queue_service.py
```diff
@@ -0,0 +1,272 @@
...
+
+    def __init__(self):
+        self._channel = None
```
Uxio0, line 39: Type

safe_transaction_service/events/services/queue_service.py
```diff
@@ -0,0 +1,272 @@
...
+        self._connection: GeventConnection = None
+        self.unsent_events = []
```
Uxio0, line 41: Type

safe_transaction_service/events/services/queue_service.py
```diff
@@ -0,0 +1,272 @@
...
+        self._connection_parameters = URLParameters(settings.EVENTS_QUEUE_URL_HOST)
+
+    def send_event(
+        self, payload: Dict[str, Any], fail_retry: Optional[bool] = True
+    ) -> bool:
+        """
+        Send an event to rabbitMq exchange
+
+        :param event: Dict with the payload of the event
```
Uxio0, line 50: Params missing on the doc and not matching the function

safe_transaction_service/events/services/queue_service.py
```diff
@@ -0,0 +1,272 @@
...
+            return False
+
+        try:
+            event = json.dumps(payload)
+            self._channel.basic_publish(
+                exchange=self.EXCHANGE_NAME, routing_key="", body=event
+            )
+            return True
+        except pika.exceptions.ConnectionClosedByBroker:
+            logger.warning("Event can not be sended due there is no channel opened")
```
Uxio0, line 65:
```suggestion
            logger.warning("Event can not be sent due to there is no channel opened")
```

safe_transaction_service/events/services/queue_service.py
```diff
@@ -0,0 +1,272 @@
...
+            self._connection_parameters,
+            on_open_callback=self.on_connection_open,
+            on_open_error_callback=self.on_connection_open_error,
+            on_close_callback=self.on_connection_closed,
+        )
+
+    def on_connection_open(self, connection: GeventConnection):
+        """This method is called by pika once the connection to RabbitMQ has
+        been established. It passes the handle to the connection object.
+        :param GeventConnection connection: The connection
```
Uxio0, line 114:
```suggestion
        """
        This method is called by pika once the connection to RabbitMQ has
        been established. It passes the handle to the connection object.
        
        :param GeventConnection connection: The connection
```

safe_transaction_service/events/services/queue_service.py
```diff
@@ -0,0 +1,272 @@
...
+class SyncQueueService(QueueService):
+    """
+    Synchronous connection with test purpose as we cannot test using gevent connection
+    """
+
+    def __init__(self):
+        super().__init__()
+        self.connect()
+
+    def connect(self):
```
Uxio0, line 233: I think you should return the connection when connecting, even if you are assigning locally

safe_transaction_service/events/services/queue_service.py
```diff
@@ -0,0 +1,272 @@
...
+        )
+
+
+class MockedQueueService:
+    """
+    Mocked class to use in case that there is not rabbitMq queue to send events
+    """
+
+    def send_event(self, event: Dict[str, Any]):
+        logger.warning("MockedQueueService: Not sending event with payload %s", event)
```
Uxio0, line 272: I would make this debug, as people with this not configured should not get a warning

Uxio0 (approved):
Really nice job!

PR #1478, opened 2023-05-22
Bump safe-eth-py to v5.4.0

Uxio0:


safe_transaction_service/history/tests/test_commands.py
```diff
@@ -359,9 +359,9 @@ def test_setup_service_mainnet(self):
             last_proxy_factory.tx_block_number, last_proxy_factory_initial_block
         )
 
-        self.assertEqual(SafeMasterCopy.objects.count(), 8)
-        self.assertEqual(SafeMasterCopy.objects.l2().count(), 1)
-        self.assertEqual(ProxyFactory.objects.count(), 4)
+        self.assertEqual(SafeMasterCopy.objects.count(), 9)
```
moisses89, line 362: Could you add a comment why we expect 9? or maybe split by version.

Uxio0: I mean, we expect to support 9 master copies in mainnet, 2 l2s and 5 proxies. At least for mainnet I wanted to do this check

Uxio0: Added the comment

moisses89 (approved):

PR #1483, opened 2023-05-24
Fix duplicate notifications

Uxio0:
- Use md5 hash of the payload instead of the payload
- Increase cache expiration time from 5 minutes to 2 hours
- Optimize redis check, so only one query is needed (by using SETNX)
- Refactor classes and tests

### DIFF
--- a/safe_transaction_service/notifications/tasks.py
+++ b/safe_transaction_service/notifications/tasks.py
@@ -1,4 +1,3 @@
-import pickle
 from typing import Any, Dict, Optional, Tuple
 
 from celery import app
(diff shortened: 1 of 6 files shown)


safe_transaction_service/notifications/tasks.py
```diff
@@ -115,8 +90,8 @@ def send_notification_task(
         return 0, 0
 
     # Make sure notification has not been sent before
-    duplicate_notification = DuplicateNotification(address, payload)
-    if duplicate_notification.is_duplicated():
+    if not mark_notification_as_processed(address, payload):
+        # Notification was processed already
```
moisses89, line 94: In the line https://github.com/safe-global/safe-transaction-service/pull/1483/files#diff-13a79df51edabd3134f56efff3cb2876ed07f122ccbca0a755be8d919cf32f2cR87 we are sending notification, could be this notification duplicated?

safe_transaction_service/notifications/utils.py
```diff
@@ -1,10 +1,58 @@
...
+        """
+        return bool(self.redis.get(self.redis_key))
+
+    def set_duplicated(self) -> bool:
+        """
+        Stores key with an expiration time of 2 hours (if not set)
+
+        :return: ``True`` if key was not set before, ``False`` otherwise
+        """
+        return bool(self.redis.set(self.redis_key, 1, ex=60 * 60 * 2, nx=True))
```
moisses89, line 43: could be the expiration time configured as environment value where if not exists the default value is 2 hours? I'm concerned if this could bring us problems with redis memory or something similar because we are increasing from 5 minutes to 2 hours.

safe_transaction_service/notifications/tasks.py
Uxio0: You are right, we should check it as soon as posible, updating that, done

safe_transaction_service/notifications/utils.py
Uxio0: Good idea, done

safe_transaction_service/notifications/utils.py
```diff
@@ -1,10 +1,65 @@
...
+        :param address:
+        :param payload:
+        :return: Key built from ``address`` and MD5 hashing the ``payload``
+        """
+        payload_hash = json.dumps(payload, sort_keys=True)
+        # Use MD5 as it's fast and should be enough for this use case
+        hex_hash = hashlib.md5(payload_hash.encode()).hexdigest()
+        return f"notifications:{address}:{hex_hash}"
+
+    def is_duplicated(self) -> bool:
```
moisses89, line 36: Looks like the last changes this function is used only for testing, do you think that is still necessary?

moisses89 (approved):

safe_transaction_service/notifications/utils.py
Uxio0: Even if it's only for testing, the function can be useful in further iterations of this feature, so I wouldn't remove it

PR #1492, opened 2023-06-01
Don't mark txs as processed before processing

Uxio0:
- Txs/events were marked as processed in a optimistic way
- When txs/events were not processed due to an exception in the middle of the process, txs/events cannot be processed again as they were marked as processed
- Mark txs/events as processed only after they are inserted
- Improve logging for storing blocks, txs and receipts on database

### DIFF
--- a/safe_transaction_service/history/indexers/element_already_processed_checker.py
+++ b/safe_transaction_service/history/indexers/element_already_processed_checker.py
@@ -9,6 +9,10 @@ logger = getLogger(__name__)
 
 
 class ElementAlreadyProcessedChecker:
+    """
+    Keeps a cache of already processed transactions and events
+    """
+
     def __init__(self):
         self._processed_element_cache = FixedSizeDict(maxlen=40_000)  # Around 3MiB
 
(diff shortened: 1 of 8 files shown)


safe_transaction_service/history/indexers/internal_tx_indexer.py
```diff
@@ -367,7 +339,18 @@ def process_elements(
...
+
+        # Mark traces as processed
+        for tx_hash in list(tx_hash_with_traces.keys()):
+            block_hash = (
+                tx_hash_with_traces[tx_hash][0]["blockHash"]
+                if tx_hash_with_traces[tx_hash]
+                else None
+            )
+            self.element_already_processed_checker.mark_as_processed(
+                tx_hash, block_hash
```
moisses89, line 351: Could be the same transaction with different trace_address to be a problem here?

Uxio0: No, as all the traces are processed together for a `tx_hash`

safe_transaction_service/history/indexers/element_already_processed_checker.py
```diff
@@ -0,0 +1,51 @@
...
+        index = HexBytes(index)
+        return tx_hash + block_hash + index
+
+    def is_processed(
+        self, tx_hash: HexBytes, block_hash: Optional[HexBytes], index: int = 0
+    ) -> bool:
+        tx_id = self.get_key(tx_hash, block_hash, index)
+        return tx_id in self._processed_element_cache
+
+    def mark_as_processed(
```
moisses89, line 30: Could you add comment to this function?
"""
Mark element as processed if weren't marked before

:return: True if... False if...

moisses89, line 30: Could you add comment to this function?
"""
Mark element as processed if weren't marked before

:return: True if... False if...

Uxio0: Done, good suggestion

PR #1520, opened 2023-06-21
Add timeouts for contract/tokens tasks

Uxio0:
- Add timeouts for contract/token tasks
- Add timeout for token list GET HTTP request
- Set version v4.20.2

### DIFF
--- a/safe_transaction_service/contracts/tasks.py
+++ b/safe_transaction_service/contracts/tasks.py
@@ -23,6 +23,9 @@ from .models import Contract
 
 logger = get_task_logger(__name__)
 
+TASK_SOFT_TIME_LIMIT = 30  # 30 seconds
+TASK_TIME_LIMIT = 60  # 1 minute
+
 
 class ContractAction(Enum):
     CREATED = 0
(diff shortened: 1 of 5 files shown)


safe_transaction_service/contracts/tasks.py
```diff
@@ -23,14 +23,17 @@
...
+
 
 class ContractAction(Enum):
     CREATED = 0
     UPDATED = 1
     NOT_MODIFIED = 2
 
 
-@app.shared_task()
+@app.shared_task(soft_time_limit=TASK_SOFT_TIME_LIMIT, time_limit=TASK_TIME_LIMIT)
```
moisses89, line 36: Is necessary useful the soft time limit? https://docs.celeryq.dev/en/2.3-archived/userguide/workers.html#time-limits 
Without catch this exception maybe don't make sense, what do you think?

Uxio0: I prefer to always try to stop the process in a "polite" way before killing it. We will be able to see in the logs the soft limit errors

moisses89 (approved):

PR #1523, opened 2023-06-22
Fix concurrent HTTP connection issues

Uxio0:
- Add missing `timeout` for some http queries
- Increase connection pool for Coingecko, Kraken, Binance and more client APIs
- Fix typo and reduce number of retries for EthereumClient (safe-eth-py new version)

### DIFF
--- a/safe_transaction_service/tokens/clients/base_client.py
+++ b/safe_transaction_service/tokens/clients/base_client.py
@@ -1,0 +1,24 @@
+import requests
+
+
+class BaseHTTPClient:
+    def __init__(self, request_timeout: int = 10):
+        self.http_session = self._prepare_http_session()
+        self.request_timeout = request_timeout
+
+    def _prepare_http_session(self) -> requests.Session:
+        """
+        Prepare http session with custom pooling. See:
+        https://urllib3.readthedocs.io/en/stable/advanced-usage.html
+        https://docs.python-requests.org/en/v1.2.3/api/#requests.adapters.HTTPAdapter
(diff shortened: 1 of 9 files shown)


safe_transaction_service/tokens/clients/coingecko_client.py
```diff
@@ -36,8 +36,10 @@ class CoingeckoClient:
     }
     base_url = "https://api.coingecko.com/"
 
-    def __init__(self, network: Optional[EthereumNetwork] = None):
-        self.http_session = requests.Session()
+    def __init__(
+        self, network: Optional[EthereumNetwork] = None, request_timeout: int = 10
```
moisses89, line 40: Maybe should be `request_timeout Optional[int] = 10`, the same in other places

safe_transaction_service/tokens/clients/base_client.py
```diff
@@ -0,0 +1,24 @@
+import requests
+
+
+class BaseHTTPClient:
+    def __init__(self, request_timeout: int = 10):
```
moisses89, line 5: request_timeout is optional?

safe_transaction_service/tokens/clients/coingecko_client.py
Uxio0: Is not optional as it's always expecting an int

safe_transaction_service/tokens/clients/base_client.py
Uxio0: Yes, and 10 seconds by default

PR #1538, opened 2023-06-28
Reenable auto reindex of Safes when finding issues

Uxio0:
- It was disabled due to it taking a lot of resources
- When setting one Safe address to reindex instead of a lot reindexing is really fast


Uxio0:
@moisses89 please check this when you are back

PR #1572, opened 2023-07-18
Don't log every reindexed Safe

Uxio0:


safe_transaction_service/history/services/index_service.py
```diff
@@ -424,7 +424,7 @@ def _reindex(
                 )
                 element_number += len(elements)
 
-            logger.info("End reindexing addresses %s", addresses)
+            logger.info("End reindexing addresses %s", addresses_str)
```
moisses89, line 427: I agree to log only the 10 first addresses

moisses89 (approved):

PR #1574, opened 2023-07-20
Make locks not blocking on tasks

Uxio0:
- For some tasks we use Redis Locks to prevent 2 celery workers executing the same task
- Previously, we waited 1 second to get the Lock (in case it was taken) before failing
- During that second, every 0.1 second the Lock was checked again
- So every time a Lock was taken 10 queries were done to Redis, instead of 1
- With the new solution Lock will return inmediately if the Lock is taken

### DIFF
--- a/safe_transaction_service/utils/tasks.py
+++ b/safe_transaction_service/utils/tasks.py
@@ -56,7 +56,6 @@ def shutdown_worker():
 def only_one_running_task(
     task: CeleryTask,
     lock_name_suffix: Optional[str] = None,
-    blocking_timeout: int = 1,
     lock_timeout: Optional[int] = LOCK_TIMEOUT,
     gevent: bool = True,
 ):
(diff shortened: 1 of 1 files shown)


safe_transaction_service/utils/tasks.py
```diff
@@ -80,9 +79,7 @@ def only_one_running_task(
     lock_name = f"locks:tasks:{task.name}"
     if lock_name_suffix:
         lock_name += f":{lock_name_suffix}"
-    with redis.lock(
-        lock_name, blocking_timeout=blocking_timeout, timeout=lock_timeout
-    ) as lock:
+    with redis.lock(lock_name, blocking=False, timeout=lock_timeout) as lock:
```
moisses89, line 82: Remove from function documentation the `blocking_timeout` https://github.com/safe-global/safe-transaction-service/pull/1574/files#diff-c9cec6c9cc96fbb7b10fac3117122b25c4107e0d2f402d7135f5f4c5b80732b6R68

moisses89 (changes requested):

safe_transaction_service/utils/tasks.py
Uxio0: Done

moisses89 (approved):

PR #1575, opened 2023-07-20
    Cache get tx identifiers in all-transactions endpoint

Uxio0:
- Add `CACHE_ALL_TXS_VIEW` configuration parameter
- Add get count relevant txs for Safe, fast SQL query to know if cache is still valid


safe_transaction_service/history/models.py
```diff
@@ -1606,6 +1606,70 @@ class SafeContractManager(models.Manager):
...
+                    FROM "history_moduletransaction"
+                    WHERE "history_moduletransaction"."safe" = %s
+                ) subquery
+                """
+
+        with connection.cursor() as cursor:
+            hex_address = HexBytes(address)
+            cursor.execute(query, [hex_address] * 8)
+            return cursor.fetchone()[0]
+
```
moisses89, line 1672: Can you analyze this query with a safe with a lot of internalTx, multisigTx...?

safe_transaction_service/history/models.py
```diff
@@ -1606,6 +1606,70 @@ class SafeContractManager(models.Manager):
     def get_banned_safes(self) -> QuerySet[ChecksumAddress]:
         return self.filter(banned=True).values_list("address", flat=True)
 
+    def get_count_relevant_txs_for_safe(self, address: ChecksumAddress) -> int:
```
moisses89, line 1609: What do you think to change this method to an staticmethod due is not using anything from the instance class.

safe_transaction_service/history/services/transaction_service.py
```diff
@@ -109,6 +107,21 @@ def store_txs_in_cache(
...
+        as that method will group some transactions (for example, 3 ERC20 can be grouped in a ``MultisigTransaction``,
+        so it will be ``1`` element for ``get_all_tx_identifiers`` but ``4`` for this function.
+
+        This query should be pretty fast, and it's meant to be used for invalidating caches.
+
+        :param safe_address:
+        :return: number of relevant txs for a Safe
+        """
+
+        return SafeContract.objects.get_count_relevant_txs_for_safe(safe_address)
```
moisses89, line 123: SafeContract is not really used in `get_count_relevant_txs_for_safe`

safe_transaction_service/history/views.py
```diff
@@ -258,12 +267,61 @@ def list(self, request, *args, **kwargs):
...
+        cache_key = f"all-txs:{int(executed)}{int(queued)}{int(trusted)}:{limit}:{offset}:{safe}:{relevant_elements}"
+        lock_key = f"locks:{cache_key}"
+
+        if not cache_timeout:
+            # Cache disabled
+            return self.get_tx_identifiers(request, *args, **kwargs)
+
+        with redis.lock(
+            lock_key,
+            timeout=60,  # This prevents a service restart to leave a lock forever
```
moisses89, line 293: If we have more redis lock I would say to include to environment var. Maybe the same than tasks?

safe_transaction_service/history/views.py
```diff
@@ -231,10 +231,19 @@ def get_parameters(self) -> Tuple[bool, bool, bool]:
         )
         return executed, queued, trusted
 
-    def list(self, request, *args, **kwargs):
+    def get_tx_identifiers(self, request, *args, **kwargs) -> Optional[Response]:
```
moisses89, line 234: I think that `get_tx_identifiers` and `get_all_tx_identifiers` are too similar, coult work for you call this function `get_page_tx_indentifiers`?

moisses89 (commented):
Good work :)

safe_transaction_service/history/models.py
Uxio0: Done, with the new index less than one second for the biggest Safes in our database

safe_transaction_service/history/models.py
Uxio0: That's not [how Django Managers work](https://docs.djangoproject.com/en/4.2/topics/db/managers/)

safe_transaction_service/history/services/transaction_service.py
Uxio0: Sure, but queries belong to `models`, so I thought the best place to put the query is on `SafeContract`, as it would be weird to put it under `MultisigTransaction`, `ModuleTransaction` or any of the related classes

safe_transaction_service/history/views.py
Uxio0: Lock should have the same request than request timeout, as it's not expected that a request can lock more, but I need to refactor as Gunicorn configuration cannot be read by the app (as it's outside the package)

moisses89: related refactor https://github.com/safe-global/safe-transaction-service/pull/1609

safe_transaction_service/history/views.py
```diff
@@ -231,17 +234,44 @@ def get_parameters(self) -> Tuple[bool, bool, bool]:
...
+        This query will merge txs and events and will return the important
+        identifiers (``safeTxHash`` or ``txHash``) filtered
+
+        :param safe:
+        :param executed:
+        :param queued:
+        :param trusted:
+        :param ordering:
+        :param limit:
+        :param offset:
```
moisses89, line 260: I think that when somebody is going to read the code can be mistaken due he can think that this function is using ordering, limit and offset to do something, but in this case is only used to debug. 
Do you think that we could do this in another way? maybe printing this debug log, before and after `get_tx_identifiers` call? 
Other we could add a comment indicating that this three parameters are used only for debugging in this function.

Uxio0:
@moisses89 PR is ready now

PR #1590, opened 2023-07-26
Refactor all-transactions endpoint

Uxio0:
- Remove distinct from `ModuleTransaction`, as every `ModuleTransaction` is showed separated even if they share the same `ethereum txHash`
- Clean transaction_service, adding comments, renaming variables and refactoring typing
- Add logging to service and view
- Rename gevent variable for task locks (not very descriptive)

### DIFF
--- a/safe_transaction_service/history/services/transaction_service.py
+++ b/safe_transaction_service/history/services/transaction_service.py
@@ -36,6 +36,9 @@ from ..serializers import (
 logger = logging.getLogger(__name__)
 
 
+AnySafeTransaction = EthereumTx | MultisigTransaction | ModuleTransaction
+
+
 class TransactionServiceException(Exception):
     pass
 
@@ -65,7 +68,7 @@ class TransactionService:
 
     def get_txs_from_cache(
         self, safe_address: str, ids_to_search: Sequence[str]
-    ) -> List[Union[EthereumTx, MultisigTransaction, ModuleTransaction]]:
+    ) -> List[AnySafeTransaction]:
         keys_to_search = [
             self.get_cache_key(safe_address, id_to_search)
             for id_to_search in ids_to_search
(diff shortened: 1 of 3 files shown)


safe_transaction_service/history/views.py
```diff
@@ -251,7 +267,23 @@ def list(self, request, *args, **kwargs):
...
+            "%s: Got all txs from identifiers for Safe=%s executed=%s queued=%s trusted=%s",
+            self.__class__.__name__,
+            safe,
+            executed,
+            queued,
+            trusted,
+        )
         all_txs_serialized = transaction_service.serialize_all_txs(all_txs)
+        logger.debug(
+            "%s: All txs from identifiers for Safe=%s executed=%s queued=%s trusted=%s where serialized",
```
moisses89, line 280: where -> were

safe_transaction_service/history/services/transaction_service.py
```diff
@@ -369,36 +409,42 @@ def get_all_txs_from_identifiers(
                 }
             )
         }
+        logger.debug(
+            "Safe=%s Got %d tokens for transfers from tokens database",
```
moisses89, line 413: maybe better from database instead from tokens database.

safe_transaction_service/history/services/transaction_service.py
```diff
@@ -282,57 +288,87 @@ def get_all_tx_identifiers(
...
-                safe=safe_address, safe_tx_hash__in=id_not_cached
+                safe=safe_address, safe_tx_hash__in=ids_not_cached
             )
             .with_confirmations_required()
             .prefetch_related("confirmations")
             .select_related("ethereum_tx__block")
             .order_by("-nonce", "-created")
         }
+        logger.debug(
+            "Safe=%s Got %d Multisig txs from identifiers",
```
moisses89, line 337: Maybe better indicate that we are getting from database filtering by identifiers. The same for the other logs.

Uxio0: I'm not understanding this, we are already stating that we are getting multisig tx from identifiers

moisses89: I haven't strong opinion about it, but reading the logs without context I'd prefer indicate that the data is coming from database, because the other way is like the identifiers list include the transactions information and it is not true.

Uxio0: But I mean, Multisig Tx from identifiers can only be fetched from database (previously we logged the cache ones) so I don't think we have to be explicit about it

moisses89 (approved):

Issue #1566, opened 2023-07-13
Worker stops to consume tasks after "connection reset by peer"


moisses89:
**Describe the bug**
Celery worker stops to consume tasks after connectionResetError
```
2023-07-13 04:44:56,439 [CRITICAL] [???/???] Couldn't ack 2351, reason:ConnectionResetError(104, 'Connection reset by peer')
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/kombu/message.py", line 131, in ack_log_error
    self.ack(multiple=multiple)
  File "/usr/local/lib/python3.10/site-packages/kombu/message.py", line 126, in ack
    self.channel.basic_ack(self.delivery_tag, multiple=multiple)
  File "/usr/local/lib/python3.10/site-packages/amqp/channel.py", line 1407, in basic_ack
    return self.send_method(
  File "/usr/local/lib/python3.10/site-packages/amqp/abstract_channel.py", line 70, in send_method
    conn.frame_writer(1, self.channel_id, sig, args, content)
  File "/usr/local/lib/python3.10/site-packages/amqp/method_framing.py", line 186, in write_frame
    write(buffer_store.view[:offset])
  File "/usr/local/lib/python3.10/site-packages/amqp/transport.py", line 347, in write
    self._write(s)
  File "/usr/local/lib/python3.10/site-packages/gevent/_socketcommon.py", line 699, in sendall
    return _sendall(self, data_memory, flags)
  File "/usr/local/lib/python3.10/site-packages/gevent/_socketcommon.py", line 409, in _sendall
    timeleft = __send_chunk(socket, chunk, flags, timeleft, end)
  File "/usr/local/lib/python3.10/site-packages/gevent/_socketcommon.py", line 338, in __send_chunk
    data_sent += socket.send(chunk, flags)
  File "/usr/local/lib/python3.10/site-packages/gevent/_socketcommon.py", line 722, in send
    return self._sock.send(data, flags)
ConnectionResetError: [Errno 104] Connection reset by peer
```
**Expected behavior**
Connection errors can happen but the worker should handle to reconnect and continue consuming tasks.

Uxio0:
After updating RabbitMQ/Redis we haven't seen this issue again

alfredolopez80:
@Uxio0 can you give us more details about it?, how updating RabbitMQ, or commit reference?

Uxio0:
There's no commit, you must update RabbitMQ in your infrastructure

[...]

PR #1591, opened 2023-07-27
Add default celery configurations

moisses89:
# Description 
Several issues are appearing on `rabbitMq` connections, ("connection reset by peer", "missing heartbeat"...) 
One possible reason could be due the high number of messages that celery are sending to keep the heartbeat, mingle and gossip. 
The following links explains a little bit more about it:
https://www.cloudamqp.com/docs/celery.html#commandline-arguments  
https://github.com/celery/celery/issues/4980

Also we change some default parameter that can be related: 
- pool connection to 0  https://github.com/celery/celery/issues/4355
- connection max retries to 0 (unlimited re-connection retries) 
- broker channel error to True, that enable retries when there is an issue with the channel connection. 

# Issues related
#1566

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -212,8 +212,18 @@ INSTALLED_APPS += [
 CELERY_BROKER_URL = env("CELERY_BROKER_URL", default="django://")
 # https://docs.celeryproject.org/en/stable/userguide/optimizing.html#broker-connection-pools
 # https://docs.celeryq.dev/en/latest/userguide/optimizing.html#broker-connection-pools
-CELERY_BROKER_POOL_LIMIT = env(
-    "CELERY_BROKER_POOL_LIMIT", default=env("CELERYD_CONCURRENCY", default=1000)
+# Configured to 0 due to connection issues https://github.com/celery/celery/issues/4355
+CELERY_BROKER_POOL_LIMIT = env.int("CELERY_BROKER_POOL_LIMIT", default=0)
+# https://docs.celeryq.dev/en/stable/userguide/configuration.html#broker-heartbeat
+CELERY_BROKER_HEARTBEAT = env.int("CELERY_BROKER_HEARTBEAT", default=0)
+
+# https://docs.celeryq.dev/en/stable/userguide/configuration.html#std-setting-broker_connection_max_retries
+CELERY_BROKER_CONNECTION_MAX_RETRIES = env.int(
+    "CELERY_BROKER_CONNECTION_MAX_RETRIES", default=0
+)
+# https://docs.celeryq.dev/en/stable/userguide/configuration.html#broker-channel-error-retry
+CELERY_BROKER_CHANNEL_ERROR_RETRY = env.bool(
+    "CELERY_BROKER_CHANNEL_ERROR_RETRY", default=True
 )
 # http://docs.celeryproject.org/en/latest/userguide/configuration.html#std:setting-result_backend
 CELERY_RESULT_BACKEND = env("CELERY_RESULT_BACKEND", default="redis://")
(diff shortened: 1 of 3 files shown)


config/settings/base.py
```diff
@@ -213,7 +213,21 @@
 # https://docs.celeryproject.org/en/stable/userguide/optimizing.html#broker-connection-pools
 # https://docs.celeryq.dev/en/latest/userguide/optimizing.html#broker-connection-pools
 CELERY_BROKER_POOL_LIMIT = env(
-    "CELERY_BROKER_POOL_LIMIT", default=env("CELERYD_CONCURRENCY", default=1000)
+    "CELERY_BROKER_POOL_LIMIT", default=env("CELERYD_CONCURRENCY", default=0)
```
moisses89, line 216: @Uxio0 I changed this value to 0 to open and close connections every time it is needed, maybe could be better a value less than 1000. I'm not sure about to put this to 0, because connections with rabbitMq were designed to be durable.

Uxio0: This is not related to connections being durable, only to the number of connections to allow at the same time

config/settings/base.py
```diff
@@ -213,7 +213,21 @@
...
+CELERY_BROKER_HEARTBEAT = env.int(
+    "CELERY_BROKER_HEARTBEAT", default=env("CELERY_BROKER_HEARTBEAT", default=0)
+)
+# https://docs.celeryq.dev/en/stable/userguide/configuration.html#std-setting-broker_connection_max_retries
+CELERY_BROKER_CONNECTION_MAX_RETRIES = env.int(
+    "CELERY_BROKER_CONNECTION_MAX_RETRIES",
+    default=env("CELERY_BROKER_CONNECTION_MAX_RETRIES", default=0),
+)
+# https://docs.celeryq.dev/en/stable/userguide/configuration.html#broker-channel-error-retry
+CELERY_BROKER_CHANNEL_ERROR_RETRY = env.int(
```
Uxio0, line 228: env.bool

config/settings/base.py
```diff
@@ -212,8 +212,23 @@
...
+)
+# https://docs.celeryq.dev/en/stable/userguide/configuration.html#std-setting-broker_connection_max_retries
+CELERY_BROKER_CONNECTION_MAX_RETRIES = env.int(
+    "CELERY_BROKER_CONNECTION_MAX_RETRIES",
+    default=env("CELERY_BROKER_CONNECTION_MAX_RETRIES", default=0),
+)
+# https://docs.celeryq.dev/en/stable/userguide/configuration.html#broker-channel-error-retry
+CELERY_BROKER_CHANNEL_ERROR_RETRY = env.bool(
+    "CELERY_BROKER_CHANNEL_ERROR_RETRY",
+    default=env("CELERY_BROKER_CHANNEL_ERROR_RETRY", default=True),
```
Uxio0, line 231: Every environment variable is duplicated

docker/web/celery/scheduler/run.sh
```diff
@@ -13,4 +13,6 @@ fi
 sleep 10
 
 echo "==> $(date +%H:%M:%S) ==> Running Celery beat <=="
-exec celery -C -A config.celery_app beat -S django_celery_beat.schedulers:DatabaseScheduler --loglevel $log_level
+exec celery -C -A config.celery_app beat \
```
Uxio0, line 16: `without-heartbeat` and so on are not required for the scheduler?

docker/web/run_dev_web.sh
```diff
@@ -0,0 +1,3 @@
+pip install django_extensions
```
Uxio0, line 1: Django extensions should be already installed as they are on `requirements.txt`, why are you adding them here?

Uxio0 (changes requested):

docker/web/celery/scheduler/run.sh
moisses89: looks like is a flag accepted only for workers.
`safe-transaction-service-scheduler-1  | Error: No such option: --without-heartbeat`

config/settings/base.py
moisses89: fixed

Uxio0 (approved):

PR #1603, opened 2023-08-04
Optimize reorg handling

### LINKED ISSUE #1554: Refactor check_reorgs
# How `check_reorgs` works? 
The current solution works at general lines as follows:
- Indexer get every block that is including transactions with interactions with safe contracts and mark this blocks as not confirmed 
- `Check_reorgs` is executed at every minute and get the unconfirmed blocks to check if everything is ok until the current block minus a margin of blocks, this is to avoid to be in the last block and ensure that the  reorg happened. This margin is configurable by environment variable.
- for each unconfirmed block got from database, `check_reorgs` request to the node the information of database bblock and the next block, then checks if (Hash(block_database) is equals to hash(block_from_node) and equals to parentHash(next_block_from_node) then if everything is ok update the block database with confirmed field to`true` but if not some actions are done to update the database with the new block and transactions.

# What is needed?
Some times due node or connection issues `check_reorgs`  starts to be far from the current block. 
I could see that the node is responding slowly after several requests for different blocks, taking some times 1 minute for 1 request, causing that `check_reorgs` started to be far and far.  
I'm thinking that this issues could happen some times because the node is receiving a lot of requests at the same time, then I would like to propose to change this function to works at different way.
I think that better approach could be batch several blocks in one request to the node and check with database in a loop the results, this is going to reduce the number of requests to the node.

Uxio0:
- Until now, reorg handling was only checking "safe" blocks (with the required number of confirmations). That made the service really slow reacting to reorgs
- With this PR, all `not confirmed` blocks are checked, but only marking as confirmed the ones with the required number of confirmations
- `Previous block hash` check was removed, as it should be enough by checking the current returned block by the node
- Pagination is used to prevent having a lot of blocks of memory in the case of a service that went out of sync
- Pagination parameter was fixed to 250 by default as it's half of the value we use to index txs/blocks (using new configuration variable `ETH_REORG_BLOCKS_BATCH`)
- Closes #1554

### DIFF
--- a/safe_transaction_service/history/services/reorg_service.py
+++ b/safe_transaction_service/history/services/reorg_service.py
@@ -1,6 +1,7 @@
 import logging
 from typing import Callable, List, Optional
 
+from django.core.paginator import Paginator
 from django.db import transaction
 
 from hexbytes import HexBytes
@@ -24,7 +25,9 @@ class ReorgServiceProvider:
             from django.conf import settings
 
             cls.instance = ReorgService(
-                EthereumClientProvider(), settings.ETH_REORG_BLOCKS
+                EthereumClientProvider(),
+                settings.ETH_REORG_BLOCKS,
+                settings.ETH_REORG_BLOCKS_BATCH,
             )
         return cls.instance
 
@@ -40,6 +43,7 @@ class ReorgService:
         self,
         ethereum_client: EthereumClient,
         eth_reorg_blocks: int,
+        eth_reorg_blocks_batch: int,
         eth_reorg_rewind_blocks: Optional[int] = 10,
     ):
         """
@@ -50,6 +54,7 @@ class ReorgService:
         """
         self.ethereum_client = ethereum_client
         self.eth_reorg_blocks = eth_reorg_blocks
+        self.eth_reorg_blocks_batch = eth_reorg_blocks_batch
         self.eth_reorg_rewind_blocks = eth_reorg_rewind_blocks
 
         # List with functions for database models to recover from reorgs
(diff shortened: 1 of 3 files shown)


safe_transaction_service/history/services/reorg_service.py
```diff
@@ -77,37 +82,52 @@ def check_reorgs(self) -> Optional[int]:
...
-                    database_block.number,
-                    HexBytes(database_block.block_hash).hex(),
-                    HexBytes(blockchain_block["hash"]).hex(),
-                )
-                return database_block.number
+                if HexBytes(blockchain_block["hash"]) == HexBytes(
+                    database_block.block_hash
+                ):
+                    logger.debug(
+                        "Block with number=%d and hash=%s is matching blockchain one, setting as confirmed",
```
moisses89, line 116: this log message is not true at all, only setting as confirmed if the next condition is true `if database_block.number <= confirmation_block` 
Maybe we can include the message inside the if block?

Uxio0:
> There is not any problem removing from this `check_reorg` the old approach that was checking also the previous block hash with parent_hash?

It shouldn't. Node should not accept a block pointing to a wrong parent, so I think is a valid assumption that this should work and be as safe as the previous approach

safe_transaction_service/history/services/reorg_service.py
Uxio0: You are right

Uxio0: Done

Uxio0:
@moisses89 is there something missing for merging?

moisses89 (approved):

PR #1615, opened 2023-08-11
Catch ValueError for RPC Indexer errors

moisses89:
### What was wrong? 👾
Some RPC node return 200 ok result and some random messages indicating an issue like for example "Block range too wide" that should indicate us that the block_process_limit configured have higher values than the RPC node accepts. 
To avoid stuck the indexer in this cases, a strategy could be reduce the indexer `block_process_limit` to one if this error happens. 
This PR add `ValueError` exception to  catch and reduce the block_process_limit to 1 as is done with other exceptions.


hectorgomezv (approved):

Uxio0 (approved):

Uxio0:
Approved, I've also updated the PR name

PR #1629, opened 2023-08-18
Add signal to trigger message webhook

### LINKED ISSUE #1240: Invalidate Safe related data when a SafeMessage is created/updated
# What is needed?
Trigger the cache invalidation hook whenever a `safe_messages.models.SafeMessage` is modified/created.

# Background
The endpoint `GET /v1/chains/<chain_id>/safes/<safe_address>` on the Safe Client Gateway gets the latest modified message via the Safe Transaction Service endpoint: `GET /v1/safes/{}/messages/?ordering=-modified&limit=1`. This works quite well however when the Safe Client Gateway returns the response to the clients the resulting message is cached (for 1 hour).

Even if the Safe gets new messages (or any other message is updated), the Safe Info cache is not invalidated and the the current stored payload is returned.

# Related issues
- https://github.com/safe-global/web-core/issues/1403

moisses89:
# Description
The goal of this PR is send webhooks when a message is created, updated or confirmed. 
`process_webhook` was included in signals file of `safe_messages` app with some differences of the history signals method: 
- in this case is not necessary check if the notification is still relevant because the safe message have not `created` field.
- don't send notification to firebase. 

Two new types or webhook were included: 
- MESSAGE_CREATED
- MESSAGE_CONFIRMATION  

Also modify `build_webhook_payload`  to include payloads for message creation and confirmation following this specs https://github.com/safe-global/safe-transaction-service/issues/1240#issuecomment-1683585133 

# Issues related
Closes #1240

### DIFF
--- a/safe_transaction_service/history/services/webhooks.py
+++ b/safe_transaction_service/history/services/webhooks.py
@@ -18,6 +18,10 @@ from safe_transaction_service.history.models import (
     TokenTransfer,
     WebHookType,
 )
+from safe_transaction_service.safe_messages.models import (
+    SafeMessage,
+    SafeMessageConfirmation,
+)
 from safe_transaction_service.utils.ethereum import get_chain_id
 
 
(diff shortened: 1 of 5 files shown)


safe_transaction_service/history/services/webhooks.py
```diff
@@ -104,6 +108,22 @@ def build_webhook_payload(
...
+                "type": WebHookType.MESSAGE_CREATED.name,
+                "messageHash": instance.message_hash,
+            }
+        ]
+    elif sender == SafeMessageConfirmation:
+        payloads = [
+            {
+                "address": instance.safe_message.safe,  # This could make a db call
+                "type": WebHookType.MESSAGE_CONFIRMATION.name,
+                "messageHash": instance.safe_message.message_hash,
```
moisses89, line 124: checking if get address and `message_hash` throw foreign key is doing two queries.

moisses89: The second time is not being evaluated again.

rmeissner (approved):

Uxio0:
Really nice work @moisses89 . I don't like duplication of `process_webhook` but we can address it when we remove notifications and refactor that part

PR #1695, opened 2023-10-05
Add address that propose a transaction in transaction fields.

### LINKED ISSUE #1592: Get account that proposed a transaction
# What is needed?
When a transaction is proposed, it should be easy to find who proposed it in the same way that it is possible to know who executed it.

# Background
It is sort of possible right now getting the signatures of a transaction and seeing whoever signed the transaction first however this does not work for the case of delegates. If a delegate proposed a transaction the context is lost and this transaction in a way just appears.

# Endpoint
If applicable, description on the endpoint and the result you expect:

## URL
`GET /api/v1/safes/<safe-address>/multisig-transactions`
`GET /api/v1/safes/<safe-address>/all-transactions`

## Response
```
{
...current body,
 source: `account that proposed transaction` 
}
```

moisses89:
Closes #1592 
- Add proposer field to indicate the owner or delegator address that propose the transaction.


safe_transaction_service/history/serializers.py
```diff
@@ -256,6 +256,12 @@ def save(self, **kwargs):
                 and (user := request.user)
             ):
                 trusted = user.has_perm("history.create_trusted")
+        try:
```
Uxio0, line 259: No need to always perform this query. You can check `if sender in self.validated_data["safe_owners"]`, and then if not you can get the delegator from `SafeContractDelegate` as you are doing.

Also you don't need to capture the exception, as if `sender` is neither an `owner` nor a `delegate`, we have a bug in the `validate` function

safe_transaction_service/history/serializers.py
```diff
@@ -546,6 +553,7 @@ class SafeMultisigTransactionResponseSerializer(SafeMultisigTxSerializerV1):
     block_number = serializers.SerializerMethodField()
     transaction_hash = Sha3HashField(source="ethereum_tx_id")
     safe_tx_hash = Sha3HashField()
+    proposer = serializers.SerializerMethodField()
```
Uxio0, line 556: Why a `SerializerMethodField()`?

safe_transaction_service/history/tests/test_views.py
```diff
@@ -865,7 +879,7 @@ def test_get_multisig_transactions(self):
         self.assertEqual(response.data["count"], 0)
         self.assertEqual(response.data["count_unique_nonce"], 0)
 
-        multisig_tx = MultisigTransactionFactory(safe=safe_address)
+        multisig_tx = MultisigTransactionFactory(safe=safe_address, proposer=proposer)
```
Uxio0, line 882: How is this working if you didn't update `MultisigTransactionFactory`?

Uxio0 (changes requested):

safe_transaction_service/history/tests/test_views.py
moisses89: because proposer can be null, we haven't data about who was the proposer on old values.

safe_transaction_service/history/serializers.py
moisses89: Don't make sense to return the value.

PR #1719, opened 2023-10-25
Set version 4.27.0

Uxio0:
- Bump safe-eth-py to v6.0.0b4 and adjust tests
- Set version 4.27.0

### DIFF
--- a/safe_transaction_service/history/tests/test_commands.py
+++ b/safe_transaction_service/history/tests/test_commands.py
@@ -359,8 +359,8 @@ class TestCommands(TestCase):
             last_proxy_factory.tx_block_number, last_proxy_factory_initial_block
         )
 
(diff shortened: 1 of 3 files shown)


safe_transaction_service/history/tests/test_commands.py
```diff
@@ -360,7 +360,7 @@ def test_setup_service_mainnet(self):
         )
 
         # At May 2023 we support 9 Master Copies, 2 L2 Master Copies and 5 Proxy Factories
```
moisses89, line 362:
```suggestion
        # At May 2023 we support 10 Master Copies, 2 L2 Master Copies and 5 Proxy Factories
```

moisses89 (approved):

PR #1727, opened 2023-10-27
Bump safe-eth-py to 6.0.0b5

Uxio0:


Uxio0:
@JagoFigueroa this is already in staging, we need QA on estimating a transaction for a 1.2.0 Safe

JagoFigueroa:
Tested and verified! (apologies for the delay, did not have permissions to post comments but now this is fixed 😄)

PR #1735, opened 2023-10-31
Add batch-size option

### LINKED ISSUE #1711: Add option to configure batch size on check_index_problems
# What is needed?
Some provider nodes have some limitations as the batch request size, I would like to add an option to configure the `batch_size` 
https://github.com/safe-global/safe-transaction-service/blob/master/safe_transaction_service/history/management/commands/check_index_problems.py#L61

moisses89:
Closes #1711 
- Add batch-size option
- Add test for `check_index_problems` 
- Fix `_reindex` wasn't reaching the last block

### DIFF
--- a/safe_transaction_service/history/tests/test_commands.py
+++ b/safe_transaction_service/history/tests/test_commands.py
@@ -8,21 +8,30 @@ from django.core.management import CommandError, call_command
 from django.test import TestCase
 
 from django_celery_beat.models import PeriodicTask
+from eth_account import Account
 
 from gnosis.eth.ethereum_client import EthereumClient, EthereumNetwork
+from gnosis.safe.tests.safe_test_case import SafeTestCaseMixin
 
 from ..indexers import Erc20EventsIndexer, InternalTxIndexer, SafeEventsIndexer
-from ..models import IndexingStatus, ProxyFactory, SafeMasterCopy
+from ..models import (
+    IndexingStatus,
+    InternalTxDecoded,
+    ProxyFactory,
+    SafeLastStatus,
+    SafeMasterCopy,
+)
 from ..services import IndexServiceProvider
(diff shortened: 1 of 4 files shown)


safe_transaction_service/history/tests/test_commands.py
```diff
@@ -443,3 +452,45 @@ def test_check_chainid_matches(self, get_chain_id_mock: MagicMock):
...
+            SafeLastStatus.objects.filter(address=safe.address).update(nonce=1)
+            self.assertEqual(InternalTxDecoded.objects.count(), 0)
+            buf = StringIO()
+            call_command(command, stdout=buf)
+            self.assertIn(
+                f"Safe={safe.address} is corrupted, has some old transactions missing",
+                buf.getvalue(),
+            )
+            # Should be 1 InternalTxDecoded
+            self.assertEqual(InternalTxDecoded.objects.count(), 1)
```
moisses89, line 494: @Uxio0  Check index problems found an issue and trigger the reindex then InternalTxDecoded should have 1 transaction taken from blockchain but the model is empty. The transaction is captured if this reindex https://github.com/safe-global/safe-transaction-service/blob/master/safe_transaction_service/history/management/commands/check_index_problems.py#L132 start in the previous block where the issue was founded. 
What do you think?

Uxio0: So you mean that `reindex_master_copies` only works when setting one block less that the one where the tx is missing?

moisses89: It's right, and I'm concerning about the check index problems command can even jump the transactions missing because this.

Uxio0: Yes it should 👍🏻

Uxio0 (approved):

PR #1762, opened 2023-11-17
Remove price calculation

### LINKED ISSUE #1758: Remove price calculation from tx service
# What is needed?
Price calculation must be removed from tx service.

- Token price endpoint must be removed
- Balances usd endpoint should set usd prices always to 0 as a first step, and then it should be removed

moisses89:
Closes #1758


moisses89:
@Uxio0 this shouln´t be included on the next release until 30th November but could be in staging until that day.

safe_transaction_service/history/services/balance_service.py
```diff
@@ -268,50 +263,28 @@ def get_usd_balances(
...
-                    token_eth_value,
-                    token_eth_value_with_timestamp.timestamp,
-                    round(fiat_balance, 4),
-                    round(fiat_conversion, 4),
+                    0.0,
+                    datetime.utcfromtimestamp(0),
+                    0.0,
+                    0.0,
                     FiatCode.USD.name,
                 )
```
moisses89, line 289: @hectorgomezv just to inform you, the transaction service will return `eth_value` as 0, `timestamp` when was calculated as 1970-01... (epoch 0), `fiat_balance` 0 and `fiat_conversion`  as 0. 
This is affecting the endpoint  `/v1/safes/{address}/balances/usd/`

hectorgomezv: Thank you @moisses89! No problem at all from the CGW perspective, the service is not using that endpoint anymore.

Uxio0 (approved):

PR #1782, opened 2023-11-27
Fix kubernetes not detecting non numeric users

Uxio0:
- https://stackoverflow.com/a/77089706
- https://stackoverflow.com/a/66367783/724991


moisses89 (approved):

Uxio0:
@luarx this is merged now

PR #1790, opened 2023-12-04
Add trusted field to token serialiser

fmrsabino:
Exposes the `trusted` field from the `Token` model entity. This field can be used for logic related to the reputation of a given token.


Uxio0 (approved):

Uxio0:
Thanks for the PR @fmrsabino !

PR #1799, opened 2023-12-12
Ignore missing chain ids in token lists

Uxio0:


safe_transaction_service/tokens/tasks.py
```diff
@@ -82,7 +82,7 @@ def update_token_info_from_token_list_task() -> int:
     token_addresses = [
         fast_to_checksum_address(token["address"])
         for token in tokens
-        if token["chainId"] == ethereum_network.value
+        if token.get("chainId") == ethereum_network.value
```
fmrsabino, line 85: If there's no `chainId` then we should not include this token at all correct?

fmrsabino (approved):

moisses89 (approved):

safe_transaction_service/tokens/tasks.py
moisses89: Yes, I agree, shouldn't be included as we cannot ensure that the token address in the chain. The proposed approach is ok in my opinion because is not including it

fmrsabino: Makes sense! Thanks 👍

PR #1793, opened 2023-12-05
Create AbiService to prevent issues during ABI fetching

Uxio0:
- Closes #1792
- Implement a `ContractMetadataService` to fetch ABIs and contract metadata
- Make sure it's a singleton, so connection pools to the different clients (Etherscan, Sourcify...) are reused
- Refactor all tests and the hacky mocks for the clients. Instead, mock directly the service as it should be from the beginning

PR #1800, opened 2023-12-13
Fix 1271 signature validation for messages

Uxio0:
- Refactor signature validation to prevent code duplication

### DIFF
--- a/safe_transaction_service/safe_messages/tests/test_views.py
+++ b/safe_transaction_service/safe_messages/tests/test_views.py
@@ -269,13 +269,13 @@ class TestMessageViews(SafeTestCaseMixin, APITestCase):
             data=data,
         )
         self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
-        self.assertEqual(
+        self.assertDictEqual(
             response.data,
             {
-                "non_field_errors": [
+                "signature": [
                     ErrorDetail(
-                        string="1 owner signature was expected, 0 received",
-                        code="invalid",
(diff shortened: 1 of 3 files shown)


safe_transaction_service/safe_messages/tests/test_views.py
```diff
@@ -269,13 +269,13 @@ def test_safe_message_add_signature_view(self, get_owners_mock: MagicMock):
...
-        self.assertEqual(
+        self.assertDictEqual(
             response.data,
             {
-                "non_field_errors": [
+                "signature": [
                     ErrorDetail(
-                        string="1 owner signature was expected, 0 received",
-                        code="invalid",
+                        string="Ensure this field has at least 1 hexadecimal chars (not counting 0x).",
```
moisses89, line 277: Why the error message changed?

moisses89 (changes requested):
Bump safe-eth-py to v6.0.0b12 because have a fix related.

safe_transaction_service/safe_messages/tests/test_views.py
Uxio0: Because previously we were not requiring minimum length for the signature in the validator, and now we can discard the signature without further validation

Uxio0:
> Bump safe-eth-py to v6.0.0b12 because have a fix related.

The fix is not required for this to work, as it's only a error message typo, and #1793 must be merged before

moisses89 (approved):

PR #1802, opened 2023-12-14
Allow deleting multisig transactions not executed

### LINKED ISSUE #1757: Delete proposed not executed transactions
# What is needed?
Allow proposed transactions to be deleted by the proposer:
- A signature from the proposer will be required.
- Add a event for transaction deletion so they can be refreshed from client gateway caches (different issue, #1805)

A EIP712 will be signed, and epoch will be included to prevent replay attacks

# EIP712
```python
{
    "types": {
        "EIP712Domain": [
            {"name": "name", "type": "string"},
            {"name": "version", "type": "string"},
            {"name": "chainId", "type": "uint256"},
            {"name": "verifyingContract", "type": "address"},
        ],
        "DeleteRequest": [
            {"name": "safeTxHash", "type": "bytes32"},
            {"name": "totp", "type": "uint256"},
        ],
    },
    "primaryType": "DeleteRequest",
    "domain": {
        "name": "Safe Transaction Service",
        "version": "1.0",
        "chainId": chain_id,
        "verifyingContract": safe_address,
    },
    "message": {
        "safeTxHash": safe_tx_hash,
        "totp": totp,
    },
}
```

# Endpoint
`DELETE` `multisig-txs/<safe-tx-hash>/`

# Body
```json
{
  "timestamp": "<integer unix epoch>"
  "signature": "<0x hex string>"
}
```

Superseeds #1737

Uxio0:
- Closes #1757

### DIFF
--- a/safe_transaction_service/history/serializers.py
+++ b/safe_transaction_service/history/serializers.py
@@ -32,6 +32,7 @@ from safe_transaction_service.utils.serializers import get_safe_owners, get_safe
 from .exceptions import NodeConnectionException
 from .helpers import DelegateSignatureHelper, DeleteMultisigTxSignatureHelper
 from .models import (
+    MAX_SIGNATURE_LENGTH,
     EthereumTx,
     ModuleTransaction,
     MultisigConfirmation,
@@ -63,7 +64,7 @@ class GnosisBaseModelSerializer(serializers.ModelSerializer):
 #            Request Serializers
 # ================================================ #
 class SafeMultisigConfirmationSerializer(serializers.Serializer):
-    signature = HexadecimalField(min_length=65)  # Signatures must be at least 65 bytes
+    signature = HexadecimalField(min_length=65, max_length=MAX_SIGNATURE_LENGTH)
 
     def validate_signature(self, signature: bytes):
         safe_tx_hash = self.context["safe_tx_hash"]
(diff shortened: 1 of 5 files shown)


safe_transaction_service/history/serializers.py
```diff
@@ -472,6 +475,51 @@ def validate(self, attrs):
...
+            message_hash = DeleteMultisigTxSignatureHelper.calculate_hash(
+                safe_address, safe_tx_hash, chain_id, previous_totp=previous_totp
+            )
+            safe_signatures = SafeSignature.parse_signature(signature, message_hash)
+            if len(safe_signatures) != 1:
+                raise ValidationError(
+                    f"1 owner signature was expected, {len(safe_signatures)} received"
+                )
+            safe_signature = safe_signatures[0]
+            if safe_signature.signature_type != SafeSignatureType.EOA:
```
moisses89, line 515: Which is the reason to only support EOA? maybe we could leave a comment here.

safe_transaction_service/history/serializers.py
```diff
@@ -472,6 +475,51 @@ def validate(self, attrs):
         )
 
 
+class SafeMultisigTransactionDeleteSerializer(serializers.Serializer):
+    safe_tx_hash = Sha3HashField()
+    signature = HexadecimalField(min_length=65)
```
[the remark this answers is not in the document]
Uxio0: We are not storing it on database, but I agree we should add a limit

safe_transaction_service/history/serializers.py
Uxio0: If we want to suport more than EOA we would need to use the EIP712 domain of the Safe, so we would make it more difficult and I don't think there's a single person proposing transactions using different things from an EOA, I will add a comment

safe_transaction_service/history/serializers.py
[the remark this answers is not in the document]
Uxio0: Added limit for signatures in this serializer and in others

safe_transaction_service/history/serializers.py
Uxio0: Done

moisses89 (approved):

PR #1818, opened 2023-12-28
Refactor events publisher

### LINKED ISSUE #1814: Refactor sending events to RabbitMQ
# What is needed?
We should study how to send events to RabbitMQ in a more optimal way.

# Background
Sometimes when there're a lot of events they get stuck and never get sent or they are sent with delay.

Things that we could check:
- Using the right RabbitMQ library
- Maybe even using the right library gevent is not working propertly
- Sending the events directly to RabbitMQ instead of sending them to a Celery queue and then to Rabbit

moisses89:
Close #1814 
# General description
The focus of this pull request is to improve the communication between safe-transaction-service and the general-`RabbitMQ`, used as events broker.
The communication issues were:
- connection issues with `RabbitMQ` from celery task
- some events/message are lost when we have high volume of events to publish
- delayed events when we have high volume of events
# Connection issues with `RabbitMQ`
After some research, we could see that the connection issues were shared socket resources between `gevent greenlets ` from `celery` and `pika greenlet`, causing a loop of tries connecting and disconnecting. 
`ConcurrentObjectUseError('This socket is already used by another greenlet: <bound method Waiter.switch of <gevent._gevent_c_waiter.Waiter object at 0xffff9d30a9d0>>')`

# Some messages/events were lost
After create a test that send (delayed with celery) 10.000 of events with an incremental id and forcing to close the client connection from RabbitMQ management, I observed delayed event disconnection. During this delay several messages were sent without detecting the connection issue, more or less 500 events were lost. 
This could be handled forcing ACK to each message from RabbitMQ, but I don't focus in this solution because the previous point. 

# Delayed events when we have high volume of events
Publish events from a task is slower than publish directly, the reason can be that to add a task that publish an event is as follows: 
- publish to celery `RabbitMQ` the task to publish the event
- worker consume the task
- worker execute the task and publish to `general-rabbitmq` the event.
vs:
- publish the event to `general-rabbitMQ`

# Proposed solution on this PR
- use synchronous pika `BlockingConnection`  instead asynchronous pika `GeventConnection`
- enable ACK to be able to resend any message that wasn't confirmed reception from `RabbitMQ` 
- execute the publisher directly from signal instead execute it in a task
- because pika `BlockingConnection` cannot be shared between threads we must to create a pool of connections.

# Benchmark 
To do this benchmark we use asynchronous connection with ACK and synchronous connection with ACK.
The time calculation  is the difference of instant time of last message sent and acknowledge less time instant of first message. 
Asynchronous 100K messages takes 22 seconds
Synchronous 100K messages takes 27 seconds
It is important to mention and in asynchronous connection ACK is handled in other thread that means publish this quantity of messages takes around 800ms vs 27 seconds of `BlockingConnection`.


safe_transaction_service/events/services/queue_service.py
```diff
@@ -13,24 +13,44 @@
...
+        if self.queue_service_pool:
+            # If there are elements on the pool, take them
+            self.instance = self.queue_service_pool.pop()
+        else:
+            # If not, get a new client
+            self.instance = getQueueService()
+        return self.instance
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.queue_service_pool.insert(0, self.instance)
```
moisses89, line 53: @Uxio0 just mentions that this pool is going to increase the number of connections with general rabbitmq by pod. 
I could see during local tests that from indexer worker we can reach 29 connections simultaneously and because we are rotating every instance in the pool we are going to keep the connections opened. 
Do you think that this can be an issue? maybe we should limit the number of instances that the pool list can store and don´t store the new instances upper the limit if all the messages were sent.

Uxio0 (changes requested):

safe_transaction_service/events/services/queue_service.py
```diff
@@ -55,7 +75,10 @@ def send_event(
         if self._channel is None or not self._channel.is_open:
```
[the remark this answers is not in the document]
moisses89: When would you do?

safe_transaction_service/events/services/queue_service.py
```diff
@@ -248,7 +135,10 @@ def connect(self) -> BlockingConnection:
         try:
             self._connection = BlockingConnection(self._connection_parameters)
             self._channel = self.open_channel()
+            self._channel.confirm_delivery()
```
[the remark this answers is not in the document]
moisses89: yes, to avoid lose messages. 
This can avoid lose messages when disconnection happens on server side during sending.

safe_transaction_service/events/services/queue_service.py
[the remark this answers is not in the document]
Uxio0: Not sure, maybe when sending messages, but sending pending messages on connect sounds a little weird to me. As I don't have a strong opinion we can leave it as it is until we find a better solution

safe_transaction_service/events/services/queue_service.py
Uxio0: I don't like the structure of the service. Service should handle sending events and maintining a pool of RabbitMQ clients, having a pool of services look wrong to me. Also users of the service should not be aware that it's intended to use with a pool, that should be handled internally. Users should call `send_event` and all the magic should happen there

moisses89: I will handle to handle the pool of connections inside of the QueueService.

moisses89: Well, this [commit](https://github.com/safe-global/safe-transaction-service/pull/1818/commits/c2726e2a9bc105453a7e44f67996f05d0261fead) do a refactor separating the connection management from `QueueService` in a new class called `BrokerConnection`. 
At this way QueueService just handle the unsent events and the pool of `BrokerConnections`.

safe_transaction_service/events/services/queue_service.py
```diff
@@ -1,276 +1,151 @@
...
-            connection.
-        """
-        self._channel = None
-        logger.error(
-            "Connection closed with %s, reopening in 5 seconds: %s",
-            self._connection_parameters.host,
-            reason,
-        )
-        connection.ioloop.call_later(5, self.connect)
+@lru_cache
```
Uxio0, line 72: You can use @cache , as we don't need to support previous Python versions (in safe-eth-py we do)

safe_transaction_service/events/services/queue_service.py
```diff
@@ -1,276 +1,151 @@
...
-        Setup the exchange on RabbitMQ by invoking the Exchange.Declare RPC
-        command.
+        :return: number of messages sent
         """
-        logger.info("Declaring exchange %s", self.exchange_name)
+        if len(self.unsent_events):
+            broker_connection = self.get_connection()
+            sent_events = 0
+            logger.info("Sending %i not sent messages", len(self.unsent_events))
+            for unsent_message in list(self.unsent_events):
```
Uxio0, line 136: `list(self.unsent_events)`

This is to prevent race conditions?

Uxio0 (changes requested):

safe_transaction_service/events/services/queue_service.py
moisses89: It was by mistake, I don't know why... but you touch a very interesting topic because two or more threads can call this function and duplicate events, I'll handle it.

safe_transaction_service/events/services/queue_service.py
```diff
@@ -129,19 +130,24 @@ def send_unsent_events(self) -> int:
 
         :return: number of messages sent
         """
-        if len(self.unsent_events):
+        if self.unsent_events:
+            # Avoid race conditions
+            unsent_events = self.unsent_events
+            self.unsent_events = []
             broker_connection = self.get_connection()
```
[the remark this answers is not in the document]
moisses89: Very good catch!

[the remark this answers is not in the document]
moisses89: updated

Uxio0 (approved):

PR #1832, opened 2024-01-18
Validate safeAppId for SafeMessage creation

Uxio0:


safe_transaction_service/safe_messages/serializers.py
```diff
@@ -62,7 +69,9 @@ def get_valid_owner_from_signatures(
 
 class SafeMessageSerializer(SafeMessageSignatureParserMixin, serializers.Serializer):
     message = serializers.JSONField()
-    safe_app_id = serializers.IntegerField(allow_null=True, default=None)
+    safe_app_id = serializers.IntegerField(
+        allow_null=True, default=None, validators=[positive_integer_validator]
+    )
```
fmrsabino, line 74: Maybe we can use `min_value=0` instead of a custom validator.

Wdyt?

Uxio0: I think I'm still on holidays mode 🤦🏼‍♂️

Uxio0: Updated

safe_transaction_service/safe_messages/tests/test_views.py
```diff
@@ -175,13 +175,33 @@ def test_safe_messages_create_view(self, get_owners_mock: MagicMock):
...
                     format="json",
                     data=data,
                 )
                 self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+                self.assertEqual(
+                    response.data,
+                    {
+                        "safe_app_id": [
+                            ErrorDetail(
+                                string="A positive integer value must be greater than 0",
```
fmrsabino, line 191: This message should also be updated right?

fmrsabino (approved):

safe_transaction_service/safe_messages/tests/test_views.py
Uxio0: Yes, done too!

PR #1842, opened 2024-01-26
Only return trusted multisig txs by default

Uxio0:
- Currently all txs were returned in multisig-txs endpoint


moisses89 (commented):
I think that we should add in swagger doc that the default return value is returned for `Trusted = True`

moisses89 (approved):

PR #1854, opened 2024-02-06
Reduce number of blocks to reindex again

Uxio0:
As indexer it's running quite well, we shouldn't require to reindex so many blocks


Uxio0:
@liliya-soroka @JagoFigueroa we are reducing the number of blocks that we are reindexing (every time we index a block we also reindex some of the previous ones). I would like to check if it can be reduced safetly so we can increase the performance and reduce the requirements of the service, let me know if you see observe missing transactions on staging

Uxio0:
@moisses89 I'm merging this but I want you to be aware

falvaradorodriguez (approved):

PR #1865, opened 2024-02-08
Fix pagination issues for all-transactions

### LINKED ISSUE #1864: Pagination for all-transactions does not work well
https://safe-transaction-gnosis-chain.safe.global/api/v1/safes/0xeda046C9f8fd29636Bb621892718B9754F244b97/all-transactions/?limit=17&offset=0&executed=true&trusted=true does not return the number of elements set by `limit`. It looks like it's related to caching of the endpoint

Uxio0:
- Closes #1864


falvaradorodriguez (approved):

Uxio0:
I've just added a new test to make sure issue is fixed

PR #1882, opened 2024-02-20
Add limit connection configuration to queue_service

### LINKED ISSUE #1848: Add configurable limit to pika pool connections
# What is needed?
Our currently approach is open connections when the service need to publish, this means that when there are a lot of traffic around Safe (proposing, confirming, executing transactions) the number of connections will increase accordingly.
I think that we can add a limit number of connections and when there is not a connection available to reuse and the limit was reached, store the event in the `unsent_events` list to send later by other thread that gets connection.

moisses89:
Close #1848


safe_transaction_service/events/services/queue_service.py
```diff
@@ -76,36 +76,48 @@ def get_queue_service():
 class QueueService:
     def __init__(self):
         self._connection_pool: List[BrokerConnection] = []
+        self._total_connections: int = 0
```
Uxio0, line 79: No need for this variable, you can just use len of the connection pool, as it's quite fast: https://stackoverflow.com/a/1115329/724991

safe_transaction_service/events/services/queue_service.py
```diff
@@ -76,36 +76,48 @@ def get_queue_service():
...
     def send_event(self, payload: Dict[str, Any]) -> int:
         """
         Publish event using the `BrokerConnection`
 
         :param payload: Number of events published
         """
-        broker_connection = self.get_connection()
-
         event = json.dumps(payload)
+        if (broker_connection := self.get_connection()) is None:
```
Uxio0, line 116:
```suggestion
        if not (broker_connection := self.get_connection()):
```

safe_transaction_service/events/services/queue_service.py
```diff
@@ -76,36 +76,48 @@ def get_queue_service():
...
+            # No available pool connection then store to send later
+            self.unsent_events.append(event)
```
Uxio0, line 118: Log with warning level, as the service might not be well configured

safe_transaction_service/events/services/queue_service.py
```diff
@@ -76,36 +76,48 @@ def get_queue_service():
...
+            broker_connection = self._connection_pool.pop()
         else:
-            return BrokerConnection()
+            broker_connection = BrokerConnection()
 
-    def release_connection(self, broker_connection: BrokerConnection) -> None:
+        self._total_connections += 1
+        return broker_connection
+
+    def release_connection(self, broker_connection: Optional[BrokerConnection] = None):
```
Uxio0, line 98: `broker_connection` cannot be optional

safe_transaction_service/events/tests/test_queue_service.py
```diff
@@ -57,6 +57,17 @@ def test_send_unsent_messages(self):
             payload = f"not sent {i}"
             _, _, body = broker_connection.channel.basic_get(self.queue, auto_ack=True)
             self.assertEqual(json.loads(body), payload)
+        # Test unsent messages when pool limit is reached
```
Uxio0, line 60: I think you should split the tests, by default no connection pool limit and then 2 tests with connection pool limit set

config/settings/base.py
```diff
@@ -522,6 +522,9 @@
 # ------------------------------------------------------------------------------
 EVENTS_QUEUE_URL = env("EVENTS_QUEUE_URL", default=None)
 EVENTS_QUEUE_EXCHANGE_NAME = env("EVENTS_QUEUE_EXCHANGE_NAME", default="amq.fanout")
+EVENTS_QUEUE_POOL_CONNECTIONS_LIMIT = env.int(
+    "EVENTS_QUEUE_POOL_CONNECTIONS_LIMIT", default=50
```
Uxio0, line 526: We should maintain our current configuration, that's no limit. So I would set a `default=0` and special logic for that case

Uxio0 (changes requested):

safe_transaction_service/events/services/queue_service.py
Uxio0: Oh I get it! I misunderstood it

safe_transaction_service/events/services/queue_service.py
```diff
@@ -76,36 +76,56 @@ def get_queue_service():
...
         """
         :return: A `BrokerConnection` from the connection pool if there is one available, othwerwise
             returns a new BrokerConnection
         """
+        if (
+            settings.EVENTS_QUEUE_POOL_CONNECTIONS_LIMIT
+            and self._total_connections >= settings.EVENTS_QUEUE_POOL_CONNECTIONS_LIMIT
+        ):
+            logger.warning(
+                "No available pool connections, currently using: %i",
```
Uxio0, line 92:
```suggestion
                "Number of active connections reached the pool limit: %d",
```

`%d` is recommended as `%i` might be deprecated

safe_transaction_service/events/services/queue_service.py
```diff
@@ -76,36 +76,56 @@ def get_queue_service():
...
         """
         Publish event using the `BrokerConnection`
 
         :param payload: Number of events published
         """
-        broker_connection = self.get_connection()
-
         event = json.dumps(payload)
+        if not (broker_connection := self.get_connection()):
+            # No available pool connection then store to send later
```
Uxio0, line 125:
```suggestion
            # No available connections in the pool, store event to send it later
```

safe_transaction_service/events/services/queue_service.py
```diff
@@ -114,6 +134,7 @@ def send_event(self, payload: Dict[str, Any]) -> int:
         logger.warning("Unable to send the event due to a connection error")
         logger.debug("Adding %s to unsent messages", payload)
         self.unsent_events.append(event)
+        self.release_connection(None)
```
Uxio0, line 137:
```suggestion
        # As the message cannot be sent, we don't want to send the problematic connection back to the pool, only reduce the number of total connections
        self.release_connection(None)
```

safe_transaction_service/events/services/queue_service.py
```diff
@@ -125,7 +146,9 @@ def send_unsent_events(self) -> int:
         if not self.unsent_events:
             return 0
 
-        broker_connection = self.get_connection()
+        if not (broker_connection := self.get_connection()):
+            # No available pool connection
```
Uxio0, line 150:
```suggestion
            # Connection not available in the pool
```

safe_transaction_service/events/tests/test_queue_service.py
```diff
@@ -58,9 +58,26 @@ def test_send_unsent_messages(self):
...
+            queue_service.release_connection(connection_1)
+            self.assertEqual(len(queue_service.unsent_events), 1)
+            self.assertEqual(queue_service.send_event(payload), 2)
+            self.assertEqual(len(queue_service.unsent_events), 0)
+
     def test_send_event_to_queue(self):
         payload = {"event": "test_event", "type": "event type"}
-        queue_service = get_queue_service()
+        queue_service = QueueService()
         # Clean previous pool connections
```
Uxio0, line 81:
```suggestion
        # Clean previous connection pool
```

safe_transaction_service/events/tests/test_queue_service.py
```diff
@@ -72,15 +89,29 @@ def test_send_event_to_queue(self):
         self.assertEqual(json.loads(body), payload)
 
     def test_get_connection(self):
-        queue_service = get_queue_service()
+        queue_service = QueueService()
         # Clean previous pool connections
```
Uxio0, line 93:
```suggestion
        # Clean previous connection pool
```

safe_transaction_service/events/tests/test_queue_service.py
```diff
@@ -72,15 +89,29 @@ def test_send_event_to_queue(self):
...
         self.assertEqual(len(queue_service._connection_pool), 1)
+        self.assertEqual(queue_service._total_connections, 1)
         queue_service.release_connection(connection_2)
         self.assertEqual(len(queue_service._connection_pool), 2)
+        self.assertEqual(queue_service._total_connections, 0)
+        with self.settings(EVENTS_QUEUE_POOL_CONNECTIONS_LIMIT=1):
+            connection_1 = queue_service.get_connection()
+            self.assertEqual(len(queue_service._connection_pool), 1)
+            self.assertEqual(queue_service._total_connections, 1)
+            # We should reach the limit connections
```
Uxio0, line 113:
```suggestion
            # We should reach the connection limit of the pool
```

safe_transaction_service/events/services/queue_service.py
```diff
@@ -76,36 +76,56 @@ def get_queue_service():
...
+    def release_connection(self, broker_connection: Optional[BrokerConnection]):
         """
         Return the `BrokerConnection` to the pool
 
         :param broker_connection:
         :return:
         """
-        return self._connection_pool.insert(0, broker_connection)
+        self._total_connections -= 1
+        # We don't add to the pool broken connections
```
Uxio0, line 113:
```suggestion
        # Don't add broken connections to the pool
```

Uxio0 (approved):
Nice job!

Issue #1890, opened 2024-02-27
Replace `unique_together` by `UniqueConstraint`


Uxio0:
`unique_together` will be deprecated in Django v5: https://docs.djangoproject.com/en/5.0/ref/models/options/#unique-together

PR #1892, opened 2024-02-27
Update unique_together by UniqueConstraint

falvaradorodriguez:
Closes https://github.com/safe-global/safe-transaction-service/issues/1890


safe_transaction_service/history/models.py
```diff
@@ -497,7 +497,11 @@ class Meta:
             Index(fields=["_from", "timestamp"]),
             Index(fields=["to", "timestamp"]),
         ]
-        unique_together = (("ethereum_tx", "log_index"),)
+        constraints = [
+            models.UniqueConstraint(
+                fields=["ethereum_tx", "log_index"], name="unique_token_transfer_index"
```
Uxio0, line 502: Would this work if you don't set `name`? I would prefer to use the default django naming if posible

falvaradorodriguez: It seems that the 'name' parameter is mandatory:
Screenshot 2024-02-27 at 17 36 26

Uxio0: Thanks Felipe! Makes sense then. I don't understand why they set an optional field for mandatory argument, but... 🤷🏻

Uxio0 (approved):

PR #1908, opened 2024-03-07
Use alias instead of annotate to optimize queries

Uxio0:
- https://docs.djangoproject.com/en/5.0/ref/models/querysets/#django.db.models.query.QuerySet.alias


moisses89 (approved):
Nice!

falvaradorodriguez (approved):

PR #1916, opened 2024-03-12
Update Celery configuration

Uxio0:
- Stop storing results, as we are not using them.
- Remove `MAX_MEMORY_PER_CHILD` and `MAX_TASKS_PER_CHILD`, as they are not valid for `gevent` pool.


config/settings/base.py
```diff
@@ -239,7 +239,7 @@
     "CELERY_BROKER_CONNECTION_RETRY_ON_STARTUP", default=True
 )
 # http://docs.celeryproject.org/en/latest/userguide/configuration.html#std:setting-result_backend
```
moisses89, line 241: Could you update the doc link? 
https://docs.celeryq.dev/en/latest/userguide/configuration.html#database-backend-settings

moisses89 (approved):

PR #1937, opened 2024-03-21
Limit Transfers

### LINKED ISSUE #1930: Safes with a lot of transfers take a lot to load
## Problem
`all-transactions` endpoint becomes so slow when there are a lot of transfers in the database

## Solution
- [ ] As a quick fix, let's limit the number of processed transfers to `1K` in `all-transactions` endpoint
- [ ] Update swagger to make it clear

Uxio0:
- Prevent issues with Safes receiving a lot of transfers every second
- Closes #1930

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -562,3 +562,8 @@ SWAGGER_SETTINGS = {
 # Shell Plus
 # ------------------------------------------------------------------------------
 SHELL_PLUS_PRINT_SQL_TRUNCATE = env.int("SHELL_PLUS_PRINT_SQL_TRUNCATE", default=10_000)
+
+# Endpoints
+TX_SERVICE_ALL_TXS_ENDPOINT_LIMIT_TRANSFERS = env.int(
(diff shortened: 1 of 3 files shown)


config/settings/base.py
```diff
@@ -562,3 +562,8 @@
 # Shell Plus
 # ------------------------------------------------------------------------------
 SHELL_PLUS_PRINT_SQL_TRUNCATE = env.int("SHELL_PLUS_PRINT_SQL_TRUNCATE", default=10_000)
+
+# Endpoints
+TX_SERVICE_ALL_TXS_ENDPOINT_LIMIT_TRANSFERS = env.int(
+    "TX_SERVICE_ALL_TXS_ENDPOINT_LIMIT_TRANSFERS", default=1_000
```
moisses89, line 568: No strong opinion here bu maybe would be better default 0 and add logic to don't limit by default.

moisses89 (approved):

config/settings/base.py
Uxio0: Thinking on our providers, I think it's better to do it as the default. Also adding logic for not limiting would imply modifying the code way more

PR #1957, opened 2024-04-01
Delete delegates from safe and owner when owner is removed

### LINKED ISSUE #1850: Delete owner’s delegated keys when the owner is removed
A feature request from the Yearn team:

If an owner who an added a delegate key has been removed, that delegate is no longer valid. So it would be good to remove all delegates by that owner when the owner is removed/replaced.

falvaradorodriguez:
Closes #1850

The delegate should only be removed if it is specifically from the safe and the owner. If it is a general delegate for a user, it should not be deleted.

### DIFF
--- a/safe_transaction_service/history/models.py
+++ b/safe_transaction_service/history/models.py
@@ -1809,11 +1809,15 @@ class SafeContractDelegateManager(models.Manager):
     def remove_delegates_for_owner_in_safe(
         self, safe_address: ChecksumAddress, owner_address: ChecksumAddress
     ) -> int:
-        return (
-            self.filter(Q(safe_contract_id=safe_address))
-            .filter(delegator=owner_address)
-            .delete()[0]
-        )
+        """
+        This method deletes delegated users only if the safe address and the owner address match.
+        Used when an owner is removed from the Safe.
+
+        :return: number of delegated users deleted
+        """
+        return self.filter(
+            safe_contract_id=safe_address, delegator=owner_address
+        ).delete()[0]
 
 
 class SafeContractDelegate(models.Model):
--- a/safe_transaction_service/history/models.py
+++ b/safe_transaction_service/history/models.py
@@ -1806,6 +1806,15 @@ class SafeContractDelegateManager(models.Manager):
             .distinct()
         )
 
+    def remove_delegates_for_owner_in_safe(
+        self, safe_address: ChecksumAddress, owner_address: ChecksumAddress
+    ) -> int:
+        return (
+            self.filter(Q(safe_contract_id=safe_address))
+            .filter(delegator=owner_address)
+            .delete()[0]
+        )
+
 
 class SafeContractDelegate(models.Model):
     """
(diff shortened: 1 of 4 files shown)


safe_transaction_service/history/models.py
```diff
@@ -1806,6 +1806,15 @@ def get_delegates_for_safe_and_owners(
             .distinct()
         )
 
+    def remove_delegates_for_owner_in_safe(
+        self, safe_address: ChecksumAddress, owner_address: ChecksumAddress
+    ) -> int:
+        return (
+            self.filter(Q(safe_contract_id=safe_address))
+            .filter(delegator=owner_address)
```
Uxio0, line 1814: No need for `Q`, also you can do the 2 queries together

```suggestion
            self.filter(safe_contract_id=safe_address, delegator=owner_address)
```

safe_transaction_service/history/models.py
```diff
@@ -1806,6 +1806,15 @@ def get_delegates_for_safe_and_owners(
             .distinct()
         )
 
+    def remove_delegates_for_owner_in_safe(
+        self, safe_address: ChecksumAddress, owner_address: ChecksumAddress
+    ) -> int:
```
Uxio0, line 1811: Add documentation for what you are returning, as that `.delete()[0]` is not very intuitive. Also it's good to explain that this method will only delete a delegate if both delegate and Safe match

safe_transaction_service/history/tests/test_tx_processor.py
```diff
@@ -101,6 +103,17 @@ def test_tx_processor_with_factory(self):
         self.assertEqual(safe_status.nonce, 1)
         self.assertEqual(safe_status.threshold, threshold)
 
+        safe_contract = SafeContract.objects.get(address=safe_address)
+        safe_contract_delegate = SafeContractDelegateFactory(
+            delegator=owner, safe_contract=safe_contract
```
Uxio0, line 108: I think you can use `safe_contract=safe_address` or `safe_contract_id=safe_address` so you don't have to retrieve the object to call the factory

Uxio0 (changes requested):
Nice job, just some small comments

safe_transaction_service/history/models.py
falvaradorodriguez: Fast copy/paste, sorry for that 😓

safe_transaction_service/history/tests/test_tx_processor.py
Uxio0: Factory should be creating the Safe if you set `safe_contract__address=address`, but it shouldn't be the case. Also I'm not an expert on `factory_boy`, so it looks good this way then 😃

falvaradorodriguez: You are right, I had tried `safe_contract=safe_address` (causes an error because it requires an instance) but not `safe_contract_id=safe_address`. This way it works as you indicated in the first comment (Lesson learned).

Uxio0: Lesson learned also for me, thanks!

Uxio0 (approved):

PR #1965, opened 2024-04-03
Clean cached receipts on reorg recover

moisses89:


moisses89:
The cache is cleaned at this point https://github.com/safe-global/safe-transaction-service/pull/1965/files#diff-b0f80c3ff20824ca54494683dd485a06d8cb04975c474db7aca1e4999ab9ef39R144 This PR is not necessary.

PR #1969, opened 2024-04-04
Fix count relevant transactions

moisses89:
# Bug description
Count of relevant transactions was wrongly calculated with outgoing ether transfers.
The count value is part of the cache key of all-transactions. Issues in this value was causing cache invalidation issues for that endpoint. 

# Query performance
The Ether transfer query was separated in `Incoming transfers` and `Outgoing transfers` to keep using the index   [history_internal_transfer_idx](https://github.com/safe-global/safe-transaction-service/blob/43f0812bb36de91c5f7a967e20a60e275dc2c98b/safe_transaction_service/history/models.py#L1006).
A partial index was added to `_from` field.
At this way the database is not scanning tables and just using Index Only for both queries. 
```
 ->  Aggregate  (cost=84.26..84.27 rows=1 width=8) (actual time=0.176..0.177 rows=1 loops=1)
               ->  Index Only Scan using history_internal_transfer_idx on history_internaltx  (cost=0.28..81.77 rows=999 width=0) (actual time=0.022..0.143 rows=998 loops=1)
                     Index Cond: ("to" = '\xf5951f6fb664819ddba656ab228fbb8e8f5a1c33'::bytea)
                     Heap Fetches: 0
 ->  Aggregate  (cost=10.32..10.33 rows=1 width=8) (actual time=0.030..0.031 rows=1 loops=1)
               ->  Index Only Scan using history_internal_transfer_from on history_internaltx history_internaltx_1  (cost=0.28..10.07 rows=102 width=0) (actual time=0.015..0.027 rows=102 loops=1)
                     Index Cond: (_from = '\xf5951f6fb664819ddba656ab228fbb8e8f5a1c33'::bytea)
                     Heap Fetches: 0

```


moisses89:
@Uxio0 we could add another conditional index for `_from` query if we need it.

Uxio0 (approved):

Uxio0:
> @Uxio0 we could add another conditional index for `_from` query if we need it.

I prefer that `Bitmap Heap Scan` to go to `Index Only Scan` if possible

moisses89:
> > @Uxio0 we could add another conditional index for `_from` query if we need it.
> 
> I prefer that `Bitmap Heap Scan` to go to `Index Only Scan` if possible

@Uxio0 looks like OR is not possible because is doing a bitmap OR and using different index. 
```
         ->  Aggregate  (cost=166.10..166.11 rows=1 width=8) (actual time=0.322..0.323 rows=1 loops=1)
               ->  Bitmap Heap Scan on history_internaltx  (cost=49.37..163.41 rows=1076 width=0) (actual time=0.075..0.290 rows=1100 loops=1)
                     Recheck Cond: (("to" = '\xf5951f6fb664819ddba656ab228fbb8e8f5a1c33'::bytea) OR (_from = '\xf5951f6fb664819ddba656ab228fbb8e8f5a1c33'::bytea))
                     Filter: ((value > '0'::numeric) AND (call_type = 0))
                     Rows Removed by Filter: 2
                     Heap Blocks: exact=26
                     ->  BitmapOr  (cost=49.37..49.37 rows=1102 width=0) (actual time=0.064..0.064 rows=0 loops=1)
                           ->  Bitmap Index Scan on history_int_to_e72886_idx  (cost=0.00..43.78 rows=1000 width=0) (actual time=0.059..0.059 rows=1000 loops=1)
                                 Index Cond: ("to" = '\xf5951f6fb664819ddba656ab228fbb8e8f5a1c33'::bytea)
                           ->  Bitmap Index Scan on history_internaltx__from_36470cfd  (cost=0.00..5.05 rows=102 width=0) (actual time=0.005..0.005 rows=102 loops=1)
                                 Index Cond: (_from = '\xf5951f6fb664819ddba656ab228fbb8e8f5a1c33'::bytea)

```

JagoFigueroa:
Hola! On friday I was transacting on staging with the fix in place and after 40 txs or so I was not able to reproduce it whereas before it it was quite easy to do.

Looking good to me 😄

PR #1984, opened 2024-04-09
4337: Fix `valid_after` and `valid_until` parsing

Uxio0:
- Fix `valid_after` and `valid_until` parsing
-  Fix `valid_until` check

### DIFF
--- a/safe_transaction_service/account_abstraction/tests/test_views.py
+++ b/safe_transaction_service/account_abstraction/tests/test_views.py
@@ -441,7 +441,7 @@ class TestAccountAbstractionViews(SafeTestCaseMixin, APITestCase):
             },
         )
 
-        # Fix `valid_until`, set it in the future
+        # Set valid_until in the future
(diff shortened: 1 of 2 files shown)


safe_transaction_service/account_abstraction/tests/test_views.py
```diff
@@ -438,3 +440,20 @@ def test_safe_operation_valid_until_create_view(
                 ]
             },
         )
+
+        # Fix `valid_until`, set it in the future
```
moisses89, line 444: Could you update the test doc? https://github.com/safe-global/safe-transaction-service/pull/1984/files#diff-3c6ffe3225ec603b1d607b99ecc7b1c05fe58b8828f77f8be255334c91b3c16bR381 
Because now is testing also the success case.

moisses89, line 444: Do you think this comment is necessary?
```suggestion
        # Set valid_until in the future
```

moisses89 (approved):

Issue #1989, opened 2024-04-12
Refactor delegate endpoint signatures to use EIP712


Uxio0:
Currently we require to sign a hash, which is really wrong, and we should refactor the delegate signature to use an EIP712, similar to what we do for the delete transactions endpoint.

mjadach-iv:
Duplicate of https://github.com/safe-global/safe-transaction-service/issues/1857#issuecomment-1934416322

JagoFigueroa:
Looks good to me after checking with Felipe. I have left the checks for the v1 endpoints to the mobile team as a staging build of the app is needed to verify that delegation is working as expected.

Cheers

PR #2001, opened 2024-04-19
Refactor delegate endpoint signatures to use EIP712

falvaradorodriguez:
Closes https://github.com/safe-global/safe-transaction-service/issues/1989

Points of interest:

- Added endpoints in v2 to manage delegates. The delete and add endpoint has an optional "safe" field to manage the case of adding/deleting delegates for a safe (previously there were two delete endpoints).

- A valid signature can be obtained with the following python script:

```python
import environ

from eth_account import Account
from gnosis.eth.eip712 import eip712_encode_hash
from gnosis.eth import EthereumClient

env = environ.Env()

ethereum_client = EthereumClient(env.str("RPC_URL"))
account_a = Account.from_key(env.str("ACCOUNT_A_PRIVATE_KEY"))
account_b = Account.from_key(env.str("ACCOUNT_B_PRIVATE_KEY"))

types = {'EIP712Domain': [{'name': 'name', 'type': 'string'},
                          {'name': 'version', 'type': 'string'},
                          {'name': 'chainId', 'type': 'uint256'}],
         'Delegate': [{"name": "delegateAddress", "type": "bytes32"},
                      {"name": "totp", "type": "uint256"}]}

message = {"delegateAddress": account_b.address,
           "totp": get_totp()}

payload = {'types': types,
           'primaryType': 'Delegate',
           'domain': {"name": "Safe Transaction Service",
                      "version": "1.0",
                      "chainId": ethereum_client.get_chain_id()},
           'message': message}

eip712_hash = eip712_encode_hash(payload)
signature = account_a.signHash(eip712_hash)
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -33,5 +33,5 @@ psycogreen==1.0.2
 psycopg2==2.9.9
 redis==5.0.3
 requests==2.31.0
-safe-eth-py[django]==6.0.0b25
+safe-eth-py[django]==6.0.0b26
 web3==6.17.0


safe_transaction_service/history/serializers.py
```diff
@@ -439,15 +398,56 @@ def validate(self, attrs):
...
+                raise ValidationError(
+                    "More than one signatures detected, just one is expected"
+                )
+            safe_signature = safe_signatures[0]
+            if safe_signature.signature_type not in (
+                SafeSignatureType.EOA,
+                SafeSignatureType.ETH_SIGN,
             ):
-                return attrs
+                raise ValidationError("Only EOA and ETH_SIGN signatures are supported")
```
moisses89, line 427: I couldn´t find this restriction in the previous version of serializer. Is this new restriction expected? 
If for example there is a Safe which one owner is other Safe, and this Safe delegated using contract signature EIPT1271 with the previous version wouldn´t be able to remove the delegate with the new version of the endpoint?

moisses89 (changes requested):

safe_transaction_service/history/serializers.py
```diff
@@ -468,28 +468,22 @@ def save(self, **kwargs):
...
-                if self.check_delegate_signature(
-                    ethereum_client, signature, operation_hash, signer
-                ):
-                    return attrs
+        delegate = self.context["request"].parser_context["kwargs"]["delegate_address"]
+        delegator = attrs["delegator"]
+        self.validate_safe_address_and_delegator(safe_address, delegator)
+        if self.validate_delegator_signature(
+            delegate, signature, delegator
+        ) or self.validate_delegator_signature(delegate, signature, delegate):
```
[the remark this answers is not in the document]
moisses89: Thanks, I didn't know

safe_transaction_service/history/serializers.py
Uxio0: As we didn't have the limit before, we shouldn't enforce now, good catch @moisses89

safe_transaction_service/history/tests/test_views_v2.py
```diff
@@ -145,3 +156,316 @@ def test_safe_collectibles_paginated(self):
         )
         self.assertEqual(response.data["results"][1]["address"], erc721_full.address)
         self.assertEqual(int(response.data["results"][1]["id"]), erc721_full.token_id)
+
+    def test_delegates_post(self):
+        url = reverse("v2:history:delegates")
+        safe_address = Account.create().address
+        delegate = Account.create()
+        delegator = Account.create()
+        label = "Saul Goodman"
```
moisses89, line 165: :smile:

safe_transaction_service/history/tests/test_views_v2.py
```diff
@@ -145,3 +156,316 @@ def test_safe_collectibles_paginated(self):
...
+            "label": "Kim Wexler",
+            "delegate": delegate.address,
+            "delegator": delegator.address,
+            "signature": delegator.signHash(hash_to_sign)["signature"].hex(),
+        }
+        response = self.client.post(url, format="json", data=data)
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(SafeContractDelegate.objects.count(), 2)
+
+        # Test not internal server error on contract signature
```
moisses89, line 241: What means this test? that we don't return a 500 for a specific case?

safe_transaction_service/history/tests/test_views_v2.py
```diff
@@ -145,3 +156,316 @@ def test_safe_collectibles_paginated(self):
...
+            response.data["non_field_errors"][0],
+        )
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+    def test_delegates_get(self):
+        url = reverse("v2:history:delegates")
+        response = self.client.get(url, format="json")
+        self.assertEqual(response.data[0], "At least one query param must be provided")
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
```
moisses89, line 258: Could you include a test for not found delegate address?

moisses89 (changes requested):

safe_transaction_service/history/tests/test_views_v2.py
falvaradorodriguez: I have just updated the commentary to avoid confusion.

safe_transaction_service/history/tests/test_views_v2.py
falvaradorodriguez: Added.

safe_transaction_service/history/tests/test_helpers.py
```diff
@@ -0,0 +1,43 @@
...
+    def test_calculate_hash(self):
+        # Mock calculate_totp
+        DelegateSignatureHelperV2.calculate_totp = MagicMock(
+            side_effect=lambda previous: 123456 if not previous else 654321
+        )
+
+        delegate_address = "0x1234567890123456789012345678901234567890"
+        chain_id = 1
+
+        expected_hash_previous_false = HexBytes(
```
moisses89, line 23: It takes some time to me to understand what means the false, previous_false did me to think that was a previous wrong hash. 
I think that could help add a comment to this line as `# Hash calculated when totp previous is false` the same for the previous true.

safe_transaction_service/history/serializers.py
```diff
@@ -982,9 +970,10 @@ class AllTransactionsSchemaSerializer(serializers.Serializer):
 # Deprecated ---------------------------------------------------------------
 
 
-class SafeDelegateDeleteSerializer(serializers.Serializer):
+class SafeDelegateDeleteDeprecatedSerializer(serializers.Serializer):
```
moisses89, line 973: Other deprecated classes were keeping the class name but with a comment that the class is deprecated, are you aware about that this class is renamed including the Deprecated tag?

moisses89 (approved):
Just some minor comments

safe_transaction_service/history/serializers.py
falvaradorodriguez: Updated the name, it had been forgotten. As for the reasons, it was in [this PR](https://github.com/safe-global/safe-transaction-service/pull/409), but I don't know why.

safe_transaction_service/history/tests/test_helpers.py
falvaradorodriguez: Added.

PR #2007, opened 2024-04-25
Add safe deployments endpoint

### LINKED ISSUE #1996: Return Safe deployments addresses
# What is needed?
Return Safe deployments addresses

# Background
Add an endpoint returning Safe contract addresses from https://github.com/safe-global/safe-deployments

# Related issues
- Superseeds #1470

# Endpoint
## URL
`GET /v1/about/safe-contracts/`

## Response
Suggested, but not final:
```
{
  V1.3.1: {
    multiSend: "0xA238CBeb142c10Ef7Ad8442C6D1f9E89e07e7761",
    multisendCallOnly: "0x40A2aCCbd92BCA938b02010E17A5b8929b49130D",
  }
}
```

moisses89:
# Description
Add a new endpoint to return the Safe deployments from https://github.com/safe-global/safe-eth-py/blob/main/gnosis/safe/safe_deployments.py 
The new endpoint can be queried by version or/and contract name providing it by query parameter.
This PR also includes the swagger documentation for this new endpoint. 

Closes #1996

### DIFF
--- a/safe_transaction_service/history/views.py
+++ b/safe_transaction_service/history/views.py
@@ -176,7 +176,7 @@ class SingletonsView(ListAPIView):
         return SafeMasterCopy.objects.relevant()
 
 
-class SafeDeployments(ListAPIView):
+class SafeDeploymentsView(ListAPIView):
     """
     Returns a list of safe deployments by version.
     """
--- a/safe_transaction_service/history/views.py
+++ b/safe_transaction_service/history/views.py
@@ -29,6 +29,7 @@ from gnosis.eth import EthereumClient, EthereumClientProvider, EthereumNetwork
 from gnosis.eth.constants import NULL_ADDRESS
 from gnosis.eth.utils import fast_is_checksum_address
 from gnosis.safe import CannotEstimateGas
+from gnosis.safe.safe_deployments import safe_deployments
 
 from safe_transaction_service import __version__
 from safe_transaction_service.utils.ethereum import get_chain_id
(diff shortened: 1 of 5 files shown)


safe_transaction_service/history/views.py
```diff
@@ -175,6 +176,75 @@ def get_queryset(self):
...
+        responses={404: "Provided version does not exist"},
+        manual_parameters=[
+            _schema_version_param,
+            _schema_contract_param,
+        ],
+    )
+    @method_decorator(cache_page(60))  # 60 seconds
+    def get(self, request):
+        """
+        Get current indexing status for ERC20/721 events
```
Uxio0, line 212: I don't think so 😉

safe_transaction_service/history/views.py
```diff
@@ -175,6 +176,75 @@ def get_queryset(self):
...
+        """
+        filter_version = self.request.query_params.get("version")
+        filter_contract = self.request.query_params.get("contract")
+
+        if filter_version and filter_version not in safe_deployments.keys():
+            return Response(status=status.HTTP_404_NOT_FOUND)
+
+        versions = [filter_version] if filter_version else list(safe_deployments.keys())
+        chain_id = get_chain_id()
```
Uxio0, line 221:
```suggestion
        chain_id = str(get_chain_id())
```

So you don't have to parse it to `str` in every loop. Btw, `chain_id` in safe-deployments should be an integer, not a string

safe_transaction_service/history/views.py
```diff
@@ -175,6 +176,75 @@ def get_queryset(self):
...
+        data_response = []
+        for version in versions:
+            contracts = []
+            if not filter_contract:
```
Uxio0, line 225:
```suggestion
            if filter_contract:
```

Just a personal preference. I think if you are checking the attribute then you should do something about it now and not on the `else` branch. This could prevent issues if we are adding an `elif` in the future

safe_transaction_service/history/views.py
```diff
@@ -175,6 +176,75 @@ def get_queryset(self):
...
+                for contract_name, addresses in safe_deployments[version].items():
+                    contracts.append(
+                        {
+                            "contract_name": contract_name,
+                            "address": addresses.get(str(chain_id)),
+                        }
+                    )
+            else:
+                # Filter by contract
```
Uxio0, line 234:
```suggestion
                # Filter by contract name
```

safe_transaction_service/history/views.py
```diff
@@ -175,6 +176,75 @@ def get_queryset(self):
...
+                    contracts.append(
+                        {
+                            "contract_name": filter_contract,
+                            "address": addresses.get(str(chain_id)),
+                        }
+                    )
+
+            data_response.append({"version": version, "contracts": contracts})
+
+        return Response(status=status.HTTP_200_OK, data=data_response)
```
Uxio0, line 245: Use the serializer you defined

safe_transaction_service/history/views.py
```diff
@@ -175,6 +176,75 @@ def get_queryset(self):
...
+            _schema_version_param,
+            _schema_contract_param,
+        ],
+    )
+    @method_decorator(cache_page(60))  # 60 seconds
+    def get(self, request):
+        """
+        Get current indexing status for ERC20/721 events
+        """
+        filter_version = self.request.query_params.get("version")
```
Uxio0, line 214: I understand you don't use filtersets due to not requiring a database call, right?

Uxio0 (changes requested):

safe_transaction_service/history/views.py
moisses89: There is not database related with this endpoint.

safe_transaction_service/history/views.py
moisses89: Fixed here https://github.com/safe-global/safe-deployments/pull/461

Uxio0: Can you then enable raising exceptions?

safe_transaction_service/history/views.py
Uxio0: Sounds good

Uxio0 (approved):

Issue #2035, opened 2024-05-13
4337: Fix issues with reverted transactions


Uxio0:
# What is needed?
Some transactions executed by Pimlico are reverted. We need to:
- Figure it out what's happening.
- Check if they are increasing nonce or not, in case they are increasing it we should index them.

# Related PRs
- https://github.com/safe-global/safe-transaction-service/pull/2040

rmeissner:
Is the state of this issue correct? (It is open and a "Draft" on our GitHub project)

Uxio0:
This was already fixed, but forgot to close it

PR #2041, opened 2024-05-13
4337: Add endpoint to return UserOperations

Uxio0:
- Sometimes SafeOperation is reverted and the only existing information is the UserOperation
- Related to #2035


falvaradorodriguez (approved):
Reviewed here https://github.com/safe-global/safe-transaction-service/pull/2042

PR #2044, opened 2024-05-14
4337: Add `safeOperation` inside of `userOperation` serializer

Uxio0:
- `user-operations` endpoints should return `safeOperation` if there's one for the `userOperation`
- Related to #2035

### DIFF
--- a/safe_transaction_service/history/serializers.py
+++ b/safe_transaction_service/history/serializers.py
@@ -757,7 +757,9 @@ class SafeCreationInfoResponseSerializer(serializers.Serializer):
     master_copy = EthereumAddressField(allow_null=True)
     setup_data = HexadecimalField(allow_null=True)
     data_decoded = serializers.SerializerMethodField()
-    safe_operation = aa_serializers.SafeOperationResponseSerializer(allow_null=True)
+    user_operation = aa_serializers.UserOperationWithSafeOperationResponseSerializer(
+        allow_null=True
+    )
 
     def get_data_decoded(self, obj: SafeCreationInfo) -> Dict[str, Any]:
         return get_data_decoded_from_data(obj.setup_data or b"")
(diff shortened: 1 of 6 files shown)


safe_transaction_service/history/serializers.py
```diff
@@ -757,7 +757,9 @@ class SafeCreationInfoResponseSerializer(serializers.Serializer):
     master_copy = EthereumAddressField(allow_null=True)
     setup_data = HexadecimalField(allow_null=True)
     data_decoded = serializers.SerializerMethodField()
-    safe_operation = aa_serializers.SafeOperationResponseSerializer(allow_null=True)
+    user_operation = aa_serializers.UserOperationWithSafeOperationResponseSerializer(
```
moisses89, line 760: This change is not indicated in the PR description, the reason behind this is that during SafeCreation the is not a SafeOperation?

moisses89:
For me is a little bit confusing have UserOperationWithSafeOperation and SafeOperationWithUserOperation. 
What is the relation between them?

safe_transaction_service/history/serializers.py
Uxio0: If UserOperation is reverted there's still UserOperation and noce is increased, but there's no SafeOperation as it gets reverted

Uxio0: So, a SafeOperation is always linked to a UserOperation, but the opposite is not true

Uxio0:
> For me is a little bit confusing have UserOperationWithSafeOperation and SafeOperationWithUserOperation. What is the relation between them?

You can query a list of UserOperations (and you will get the related safeOperation if it exist) or you can query a list of SafeOperations (and you will always get the related UserOperation).

In the future we might be able to remove one of the enpoints, but currently while we figure it out how to add support for 4337 I think all the endpoints are good there

moisses89 (approved):

PR #2047, opened 2024-05-15
Add logic to define Swagger tags

### LINKED ISSUE #2036: Organize swagger methods with tags
## Reference

- https://drf-yasg.readthedocs.io/en/stable/custom_spec.html?highlight=operation_keys#swaggerautoschema

Uxio0:
- Closes #2036

Currently tags are generated using the first path in the url, like `v1` or `v2`.
We are now defining some logic to generate `tags`:
- If they are explicitly defined in the view, we keep that (`self.overrides`).
- If the `operation_id` contains any of the words defined, we override the tag.
- Otherwise, just iterate the `operation_keys` and return

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -572,6 +572,7 @@ SWAGGER_SETTINGS = {
     "SECURITY_DEFINITIONS": {
         "api_key": {"type": "apiKey", "in": "header", "name": "Authorization"}
     },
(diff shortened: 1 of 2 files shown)


falvaradorodriguez:
Looks really nice now! 
Screenshot 2024-05-15 at 15 59 35

falvaradorodriguez (approved):

PR #2048, opened 2024-05-15
Remove deprecated endpoints

Uxio0:
- `master-copies/` -> `singletons/`
- `transactions/` -> `multisig-transactions/`

### DIFF
--- a/safe_transaction_service/history/tests/test_views.py
+++ b/safe_transaction_service/history/tests/test_views.py
@@ -3517,10 +3517,6 @@ class TestViews(SafeTestCaseMixin, APITestCase):
         url = reverse("v1:history:singletons")
         return self._test_singletons_view(url)
 
-    def test_master_copies_view(self):
-        url = reverse("v1:history:master-copies")
-        return self._test_singletons_view(url)
-
     def test_modules_view(self):
         invalid_address = "0x2A"
         response = self.client.get(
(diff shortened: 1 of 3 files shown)


Uxio0:
@hectorgomezv @iamacook @fmrsabino Could you confirm this does not affect cgw?

iamacook:
> @hectorgomezv @iamacook @fmrsabino Could you confirm this does not affect cgw?

We have migrated to `singletons`. However, a `/transactions/` endpoint is being used [for deletion](https://github.com/safe-global/safe-client-gateway/blob/replace-imitation-data/src/datasources/transaction-api/transaction-api.service.ts#L622). Should that also be migrated (it's not marked as deprecated on Swagger)? If so, I'm assuming it has no breaking changes?

Uxio0:
@iamacook This was not intended. I've been trying to deprecate `transactions` -> `multisig-transactions` for multiple years now, but as we use the same view when we implemented the deletion it was posible to use the old path too. You should migrate asap to `multisig-transactions`

iamacook:
> @iamacook This was not intended. I've been trying to deprecate `transactions` -> `multisig-transactions` for multiple years now, but as we use the same view when we implemented the deletion it was posible to use the old path too. You should migrate asap to `multisig-transactions`

We weren't aware of this. I've opened a [PR migrating it](https://github.com/safe-global/safe-client-gateway/pull/1549) and will keep you informed.

Uxio0:
> We weren't aware of this. I've opened a [PR migrating it](https://github.com/safe-global/safe-client-gateway/pull/1549) and will keep you informed.

We will then wait for the next release and don't it today

falvaradorodriguez (approved):

PR #2065, opened 2024-06-03
4337: Add endpoint to add signatures to an existing SafeOperation

### LINKED ISSUE #2060: 4337: Add endpoint for adding more signatures to a safe-operation
# What is needed?
- Add more signatures to a `safe-operation`
- Similar flow to what we do with `multisig-transactions`

# Endpoint
## URL
`POST /api/v1/safe-operations/<safe-operation-hash>/confirmations/`

## Response
`201`

Uxio0:
- Closes #2060

### DIFF
--- a/safe_transaction_service/account_abstraction/models.py
+++ b/safe_transaction_service/account_abstraction/models.py
@@ -5,6 +5,7 @@ from typing import Optional
 from django.db import models
 from django.db.models import Index
 
+from eth_abi.packed import encode_packed
 from hexbytes import HexBytes
 from model_utils.models import TimeStampedModel
 
@@ -138,8 +139,35 @@ class SafeOperation(TimeStampedModel):
     def __str__(self) -> str:
         return f"{HexBytes(self.hash).hex()} SafeOperation for user-operation={HexBytes(self.user_operation_id).hex()}"
 
+    def build_signature_prefix(self) -> bytes:
+        """
+        `UserOperation` `signature` is used to store data for the `SafeOperation`
+        First 6 bytes of the signature are `SafeOperation` `valid_after` and next 6 are `valid_until`
+        Real user signature is after those 12 bytes
+
+        :return: 12 bytes of `valid` encoded
+        """
+        return encode_packed(
+            ["uint48"] * 2,
+            [
+                int(valid_date.timestamp()) if valid_date else 0
+                for valid_date in (
+                    self.valid_after,
+                    self.valid_until,
+                )
+            ],
+        )
+
     def build_signature(self) -> bytes:
-        return b"".join(
+        """
+        `UserOperation` `signature` is used to store data for the `SafeOperation`
+        First 6 bytes of the signature are `SafeOperation` `valid_after` and next 6 are `valid_until`
+        Real user signature is after those 12 bytes
+
+        :return: 12 bytes of `valid` encoded following by the signature built from the confirmations
+        """
+        valid_encoded = self.build_signature_prefix()
+        confirmations_signature = b"".join(
             [
                 HexBytes(signature)
                 for _, signature in sorted(
@@ -148,6 +176,7 @@ class SafeOperation(TimeStampedModel):
                 )
             ]
         )
+        return valid_encoded + confirmations_signature
 
 
 class SafeOperationConfirmation(TimeStampedModel):
--- a/safe_transaction_service/account_abstraction/models.py
+++ b/safe_transaction_service/account_abstraction/models.py
@@ -5,7 +5,6 @@ from typing import Optional
 from django.db import models
 from django.db.models import Index
 
-from eth_abi.packed import encode_packed
 from hexbytes import HexBytes
 from model_utils.models import TimeStampedModel
 
@@ -84,23 +83,9 @@ class UserOperation(models.Model):
             else None
         )
 
-        if self.signature:
-            signature = HexBytes(self.signature)
-        elif self.safe_operation:
-            # `UserOperation` `signature` is used to store data for the `SafeOperation`
-            # First 6 bytes of the signature are `SafeOperation` `valid_after` and next 6 are `valid_until`
-            # Real user signature is after those 12 bytes
-            # If there's no signature it should be built from `SafeOperation` fields
-            valid_after, valid_until = [
-                int(valid_date.timestamp()) if valid_date else 0
-                for valid_date in (
-                    self.safe_operation.valid_after,
-                    self.safe_operation.valid_until,
-                )
-            ]
-            signature = encode_packed(["uint48"] * 2, [valid_after, valid_until])
-        else:
-            signature = b""
+        if not self.signature:
+            raise ValueError("Signature should not be empty")
+        signature = HexBytes(self.signature)
 
         return UserOperationClass(
             HexBytes(self.hash),
--- a/safe_transaction_service/account_abstraction/models.py
+++ b/safe_transaction_service/account_abstraction/models.py
@@ -5,6 +5,7 @@ from typing import Optional
 from django.db import models
 from django.db.models import Index
 
+from eth_abi.packed import encode_packed
 from hexbytes import HexBytes
 from model_utils.models import TimeStampedModel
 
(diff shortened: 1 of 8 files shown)


safe_transaction_service/account_abstraction/serializers.py
```diff
@@ -93,6 +78,39 @@ def _validate_signature(
             safe_signatures.append(safe_signature)
         return safe_signatures
 
+
+class SafeOperationSerializer(
+    SafeOperationSignatureValidatorMixin, serializers.Serializer
+):
+    nonce = serializers.IntegerField(min_value=0)
```
moisses89, line 85: nonce is uint256, maybe would be better set as decimal and the number of max digits. 
https://www.erc4337.io/docs/understanding-ERC-4337/user-operation

safe_transaction_service/account_abstraction/serializers.py
```diff
@@ -93,6 +78,39 @@ def _validate_signature(
...
+    init_code = eth_serializers.HexadecimalField(allow_null=True)
+    call_data = eth_serializers.HexadecimalField(allow_null=True)
+    call_gas_limit = serializers.IntegerField(min_value=0)
```
moisses89, line 88: also this field is uint256.  Every integer field is uint256 type in the ERC4337 specification, do you think that we could have an overflow?

safe_transaction_service/account_abstraction/serializers.py
fmrsabino: Do you mean in this case to use `Uint256Field`? In that case I also agree and is in line with the other representations that we have for `nonce`.

safe_transaction_service/account_abstraction/serializers.py
Uxio0: In python `int` type is unbounded

moisses89: Ok, but other clients could have issues thinking that is int and receiving bigger numbers.

safe_transaction_service/account_abstraction/serializers.py
moisses89: Yes, Uint256Field is a decimal field with the max_digits configured to fit to an unsigned 256 bits number.

safe_transaction_service/account_abstraction/models.py
```diff
@@ -83,6 +84,19 @@ def to_user_operation(self, add_tx_metadata: bool = False) -> UserOperationClass
             else None
         )
 
+        if self.signature:
+            signature = HexBytes(self.signature)
+        else:
+            # Signature stores `valid_after` and `valid_until`, we can build it from SafeOperation fields
```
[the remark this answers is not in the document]
Uxio0: Improved docs for this

safe_transaction_service/account_abstraction/serializers.py
Uxio0: But you are pointing out the serializer, not the model. Serializer is alright with an `IntegerField`

safe_transaction_service/account_abstraction/serializers.py
Uxio0: Bigint is now native to js and other languages, so I expect js to take care of that. Just in case, @dasanra , WDYT?

safe_transaction_service/account_abstraction/models.py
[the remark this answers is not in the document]
nlordell: > (@nlordell would be able to provide more context)

Correct.

This is just how the spec was defined and how we chose to implement validity ranges for Safes.

safe_transaction_service/account_abstraction/serializers.py
fmrsabino: I think this can be more of an issue with the default parsing that is provided. JSON doesn't set any limits on `Number` but some frameworks might map a `Number` to lower bounded integer values causing an overflow. This is not the case for some implementations though. For example, JS has no issues with it:

```javascript
JSON.parse("{\"number\": 115792089237316195423570985008687907853269984665640564039457584007913129639936}")
{ number: 1.157920892373162e+77 }
```

I would say that the most friendly option towards the consumers of the API might actually be representing this as a `String` which moves the boundary validation to the consumers.

safe_transaction_service/account_abstraction/serializers.py
fmrsabino: > But you are pointing out the serializer, not the model. Serializer is alright with an `IntegerField`

I would say that this is in line with the rest of the project 👍 

Maybe in the future we can discuss if `nonce` should be represented as string (same reasons as https://github.com/safe-global/safe-transaction-service/pull/2065#discussion_r1629487900). This would be a breaking change though.

safe_transaction_service/account_abstraction/serializers.py
fmrsabino: Some extra information that might help deciding between one or the other:

> In fact, JSON allows number literals that are arbitrarily long; they just cannot be parsed to full precision in JavaScript. If you are communicating with another program in a language that supports longer integers (such as 64-bit integers), and you want to transmit the BigInt as a JSON number instead of a JSON string, see [Lossless number serialization](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON#using_json_numbers).

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#use_within_json

Uxio0: Maybe we should deal with this in another issue, as the current existing endpoints are using Integers

safe_transaction_service/account_abstraction/models.py
```diff
@@ -83,6 +84,24 @@ def to_user_operation(self, add_tx_metadata: bool = False) -> UserOperationClass
...
+            valid_after, valid_until = [
+                int(valid_date.timestamp()) if valid_date else 0
+                for valid_date in (
+                    self.safe_operation.valid_after,
+                    self.safe_operation.valid_until,
+                )
+            ]
+            signature = encode_packed(["uint48"] * 2, [valid_after, valid_until])
+        else:
+            signature = b""
```
fmrsabino, line 103: As discussed, signatures should not be empty and probably we can throw here instead 👍

fmrsabino (changes requested):

safe_transaction_service/account_abstraction/serializers.py
nlordell: For some reason I'm subscribed to this issue and stumbled on this comment.

We had similar issues at CoW and came to the conclusion: big integers should be serialized to JSON strings over the wire.

Just to dump what we found during our investigations at the time (and how I remember it): JSON standard only specifies how numeric literals are [parsed](https://www.json.org), but does not impose any restrictions on the actual representation of those values. The actual underlying representation of numbers is left up to the parser. Some do exclusively IEEE 754 doubles, some do a combination of doubles and 64-bit integers (`serde` in Rust this this AFAIR), etc. Notably the `JSON.parse` method part of the ECMA script standard represents all numerical values with doubles by default.

Note that there are, in the ECMA script standard, ways to correctly parse these numbers ([source](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#the_reviver_parameter)), but for compatibility purposes, strings are just easier to use for clients and typically suffer less from compatibility issues in ecosystems with simpler JSON parsers.

Anyway, unsolicited drive-by comment, Sorry :see_no_evil:.

fmrsabino: > Note that there are, in the ECMA script standard, ways to correctly parse these numbers ([source](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#the_reviver_parameter)), but for compatibility purposes, strings are just easier to use for clients and typically suffer less from compatibility issues in ecosystems with simpler JSON parsers.

This has been my experience as well working with multiple parsers on different stack. Each parser sets constraints or fallback logic differently when it comes to trying to map the JSON value to a native type representation.

Given the parser's responsibility in all of this I'm of the opinion that a `String` is the value is most "compatible" and client-friendly format for big numbers.

Uxio0: Ok, agree with going for CharField. But I would prefer to do it in a different PR and do the same for the existing 4337 endpoints

fmrsabino (approved):
👏

PR #2068, opened 2024-06-04
4337: Fix exception when a receipt is not found

Uxio0:
- We were counting on the UserOperation receipt always existing for processed UserOperations.
- Sometimes, somehow, it doesn't happen, and indexing gets stuck.
- Now we handle the exception and ignore the UserOperation. It can always be indexed later.


fmrsabino (approved):
👏 🚀

PR #2079, opened 2024-06-11
Optimize load of addresses in indexer

Uxio0:
Make indexers use `sets` instead of db instances

- Previously, `MonitoredAddress` instances were used.
- Currently, only `address` is retrieved from database, and transformed into a set
- This approach is:
    - More optimal both on CPU/RAM, as it does not have to use a `list` and then convert to a `set`
    - Easier to cache (in a following PR)
    - Easiert to extend and implement new indexers
- Tests and indexers were refactored
- Typing was checked and fixed

### DIFF
--- a/safe_transaction_service/history/indexers/erc20_events_indexer.py
+++ b/safe_transaction_service/history/indexers/erc20_events_indexer.py
@@ -3,6 +3,7 @@ from logging import getLogger
 from typing import Iterator, List, Optional, Sequence
 
 from django.db.models import QuerySet
+from django.db.models.query import EmptyQuerySet
 
 from eth_typing import ChecksumAddress
 from web3.contract.contract import ContractEvent
@@ -15,7 +16,6 @@ from ..models import (
     ERC20Transfer,
     ERC721Transfer,
     IndexingStatus,
-    MonitoredAddress,
     SafeContract,
     TokenTransfer,
 )
@@ -72,11 +72,11 @@ class Erc20EventsIndexer(EventsIndexer):
 
     @property
     def database_queryset(self) -> QuerySet:
-        return SafeContract.objects.all()
+        return SafeContract.objects
 
     def _do_node_query(
         self,
-        addresses: List[ChecksumAddress],
+        addresses: set[ChecksumAddress],
         from_block_number: int,
         to_block_number: int,
     ) -> List[LogReceipt]:
@@ -89,7 +89,8 @@ class Erc20EventsIndexer(EventsIndexer):
         :return:
         """
 
-        # If not too much addresses it's alright to filter in the RPC server
+        # If not too many addresses are provided it's alright to do the filtering in the RPC server
+        # Otherwise, get all the ERC20/721 events and filter them here
         parameter_addresses = (
             None if len(addresses) > self.query_chunk_size else addresses
         )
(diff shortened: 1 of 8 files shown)


safe_transaction_service/history/indexers/ethereum_indexer.py
```diff
@@ -474,40 +483,24 @@ def start(self) -> Tuple[int, int]:
...
+            current_block_number
+        ):
             logger.info(
-                "%s: Processing %d not updated addresses total",
+                "%s: Processing not updated addresses",
                 self.__class__.__name__,
-                len(not_updated_addresses),
             )
 
-            # Not updated addresses are sorted by tx_block_number
```
moisses89, line 487: This is not True now, are you aware of this?

safe_transaction_service/history/indexers/erc20_events_indexer.py
```diff
@@ -71,12 +71,12 @@ def database_field(self):
         return "erc20_block_number"
 
     @property
-    def database_queryset(self):
-        return SafeContract.objects.all()
+    def database_queryset(self) -> QuerySet:
+        return SafeContract.objects
```
moisses89, line 75: I think that this could break this https://github.com/safe-global/safe-transaction-service/pull/2079/files#diff-23397faacb994eb17a67a778b0607b0e5739593676738489ae13c9c3433ddfe4R198
Maybe better set the all inside of the queryset.

safe_transaction_service/history/indexers/internal_tx_indexer.py
```diff
@@ -178,7 +184,7 @@ def _find_relevant_elements_using_trace_filter(
                 to_traces = self.ethereum_client.tracing.trace_filter(
                     from_block=from_block_number,
                     to_block=to_block_number,
-                    to_address=addresses,
+                    to_address=list(addresses),
```
moisses89, line 187: Finally we use a list here :)

safe_transaction_service/history/tasks.py
```diff
@@ -130,12 +130,11 @@ def index_erc20_events_out_of_sync_task(
...
-        almost_updated_address.address
-        for almost_updated_address in erc20_events_indexer.get_almost_updated_addresses(
-            current_block_number
-        )[:number_of_addresses]
-    ]
+    addresses = (set(addresses) if addresses else None) or set(
+        list(erc20_events_indexer.get_almost_updated_addresses(current_block_number))[
+            :number_of_addresses
+        ]
+    )
```
moisses89, line 137:
```suggestion
    addresses = set(addresses) if addresses else  set(
        list(erc20_events_indexer.get_almost_updated_addresses(current_block_number))[
            :number_of_addresses
        ]
    )
```

moisses89 (changes requested):

safe_transaction_service/history/indexers/erc20_events_indexer.py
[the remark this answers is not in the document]
Uxio0: Agree

safe_transaction_service/history/indexers/ethereum_indexer.py
Uxio0: I'm aware, it shouldn't be relevant now, but thanks for reminding me :)

safe_transaction_service/history/indexers/erc20_events_indexer.py
Uxio0: Agree, you already flagged it in another comment

safe_transaction_service/history/indexers/internal_tx_indexer.py
Uxio0: `set` should work, but better to use a `list` as it's the expected parameter

moisses89 (approved):

JagoFigueroa:
Hola! I have been transacting in Sepolia (L1) and Polygon (L2) and no issues so far. Will keep an eye but looking good to me 😄

PR #2083, opened 2024-06-14
Cache db addresses query for ERC20/721 indexer

Uxio0:
- The number of addresses in database can get really high (for example in Optimism).
- It's not required to go to DB every time, as it can be easily cached.
- A new `created` field was added to `SafeContract`, so only new created addresses will be retrieved from database.


safe_transaction_service/history/indexers/erc20_events_indexer.py
```diff
@@ -210,7 +217,32 @@ def get_almost_updated_addresses(
...
+            addresses = set()
+            last_checked = None
+
+        for created, address in query.values_list("created", "address").order_by(
+            "created"
+        ):
+            addresses.add(address)
+
+        try:
+            last_checked = created
```
fmrsabino, line 239: We always try to set `last_checked` here. Would it make sense to remove the respective assignments in line 227 and 231? We can also set the initial value of `last_checked` to `None` in line 220.

Uxio0: I thought about it too, but I feel it's more clear as it is right now, as it's clear that if cache is in place last_checked is the last element in cache, and otherwise is None

safe_transaction_service/history/indexers/erc20_events_indexer.py
```diff
@@ -210,7 +217,32 @@ def get_almost_updated_addresses(
...
+        last_checked: Optional[datetime.datetime]
+        if self.addresses_cache:
+            # Only search for the new addresses
+            query = self.database_queryset.filter(
+                created__gte=self.addresses_cache.last_checked
+            )
+            addresses = self.addresses_cache.addresses
+            last_checked = self.addresses_cache.last_checked
+        else:
+            query = self.database_queryset.all()
```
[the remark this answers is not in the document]
Uxio0: This is not an issue at all, and I think it makes it more clear

moisses89 (approved):

JagoFigueroa:
Looking good to me! 😄

PR #2089, opened 2024-06-19
Use address type for EIP712 payload

### LINKED ISSUE #2084: Incorrect type for /v2/delegate endpoint delegateAddress
**Describe the bug**
In the signTypedData form for the Safe Transaction Service, the following type is used as `AddDelegate`:

```
        "AddDelegate": [
            {"name": "delegateAddress", "type": "bytes32"},
            {"name": "totp", "type": "uint256"},
        ],
```

However, the correct type for delegateAddress would be address. This way a normal provided address needs to be left padded.

fmrsabino:
### What was wrong? 👾

Closes #2084

### How was it fixed? 🎯

The Delegate EIP712 payload was using `bytes32` as the type for the delegate address (which is an Ethereum Address).

While this works, it represents the wrong input type and size constraints that we should accept for that field (an address is 20 bytes so any input bigger than that should not be valid).

### Considerations

Signatures made before this change will not be considered valid as the resulting has would be different. The new payload would need to be signed again by the required signers again in those scenarios.

### DIFF
--- a/safe_transaction_service/history/helpers.py
+++ b/safe_transaction_service/history/helpers.py
@@ -106,7 +106,7 @@ class DelegateSignatureHelperV2(TemporarySignatureHelper):
                     {"name": "chainId", "type": "uint256"},
                 ],
                 "Delegate": [
-                    {"name": "delegateAddress", "type": "bytes32"},
+                    {"name": "delegateAddress", "type": "address"},
                     {"name": "totp", "type": "uint256"},
                 ],
             },
--- a/safe_transaction_service/history/tests/test_helpers.py
+++ b/safe_transaction_service/history/tests/test_helpers.py
@@ -22,7 +22,7 @@ class TestDelegateSignatureHelperV2(SafeTestCaseMixin, TestCase):
 
         # Hash calculated when totp previous is false
         expected_hash_previous_totp_false = HexBytes(
-            "0xc095ec37d1798b39b8cf9306a3d6788f6118f46a0d18fcfac037c8306bdbf397"
+            "0x40ef28feb993bf127265260d48ab1a427d5b852aa8b38f511fcd368bfc9cbfdf"
         )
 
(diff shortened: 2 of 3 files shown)


Uxio0 (approved):

falvaradorodriguez (approved):

safe_transaction_service/history/helpers.py
```diff
@@ -106,7 +106,7 @@ def calculate_hash(
                     {"name": "chainId", "type": "uint256"},
                 ],
                 "Delegate": [
-                    {"name": "delegateAddress", "type": "bytes32"},
+                    {"name": "delegateAddress", "type": "address"},
```
falvaradorodriguez, line 109: This will be a breaking change, no?

falvaradorodriguez (changes requested):

safe_transaction_service/history/helpers.py
fmrsabino: Kind of yes. I've mentioned this in the PR description.

falvaradorodriguez: We should make the same change here: https://github.com/safe-global/safe-eth-py/blob/d780c51039ebcc2593f80dbc4e8f1140aadfe019/gnosis/safe/api/transaction_service_api/transaction_service_messages.py#L18

falvaradorodriguez (approved):
Based on the description, the PR is ok.

falvaradorodriguez (approved):

safe_transaction_service/history/helpers.py
fmrsabino: > We should make the same change here: https://github.com/safe-global/safe-eth-py/blob/d780c51039ebcc2593f80dbc4e8f1140aadfe019/gnosis/safe/api/transaction_service_api/transaction_service_messages.py#L18

Added it here. https://github.com/safe-global/safe-eth-py/pull/1176

Thanks @falvaradorodriguez !

PR #2105, opened 2024-06-28
Add ENS Subgraph configuration

fmrsabino:
- Adds the possibility to add an ENS Subgraph configuration.
- The Subgraph requires an API key which can be obtained via The Graph interface (https://thegraph.com/docs/en/querying/querying-the-graph/)
- Adds three environment variables that should be used if The Subgraph support needs to be enabled:
  * `ENS_SUBGRAPH_URL` – the base url to query the subgraph from
  * `ENS_SUBGRAPH_API_KEY` – the API key that was obtained from The Graph
  * `ENS_SUBGRAPH_ID` – the subgraph id. An example of an ENS Subgraph can be found in https://thegraph.com/explorer/subgraphs/5XqPmWe6gjyrJtFn9cLy237i4cWw2j9HcUJEXsP5qGtH?view=Query&chain=arbitrum-one. To query this graph, you should use `5XqPmWe6gjyrJtFn9cLy237i4cWw2j9HcUJEXsP5qGtH` as the subgraph id.
- The ENS Subgraph configuration is only required if querying on `Mainnet`. 
- Support was added to `Sepolia` and `Holesky`. These two networks do not require an API key but should not be considered for production usage (rate limits might apply).
- If the current chain is `Mainnet` and the Subgraph configuration was not provided, a fallback for Mainnet is used. However, such fallback might be under stricter rate limits and should only be used in a testing environment. A warning is logged if the fallback option is used.

### DIFF
--- a/safe_transaction_service/history/services/collectibles_service.py
+++ b/safe_transaction_service/history/services/collectibles_service.py
@@ -17,7 +17,7 @@ from cachetools import TTLCache, cachedmethod
 from eth_typing import ChecksumAddress
 from redis import Redis
 
-from gnosis.eth import EthereumClient, EthereumClientProvider
+from gnosis.eth import EthereumClient, EthereumClientProvider, EthereumNetwork
 from gnosis.eth.clients import EnsClient
 
 from safe_transaction_service.tokens.constants import (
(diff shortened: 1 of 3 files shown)


safe_transaction_service/history/services/collectibles_service.py
```diff
@@ -161,7 +161,41 @@ def __init__(self, ethereum_client: EthereumClient, redis: Redis):
...
+                "https://api.studio.thegraph.com/query/49574/ensholesky/version/latest",
+            )
+        elif base_url and api_key and subgraph_id:
+            config = EnsClient.SubgraphConfig(
+                network=EthereumNetwork.MAINNET,
+                base_url=base_url,
+                api_key=api_key,
+                subgraph_id=subgraph_id,
+            )
+        else:
```
moisses89, line 186: I don't think that we need this sequence of ifs.
```suggestion
        if  base_url: 
            config = EnsClient.Config(self.ethereum_network, base_ur, api_key, subgraph_idl)
        else: 
            logger.warning()
            # Default fallback
```

moisses89 (changes requested):

safe_transaction_service/history/services/collectibles_service.py
fmrsabino: Not sure if I follow. If we only check that the `base_url` is not `None` then `api_key` and `subgraph_id` might be `None` and we should not use the `EnsClient.SubgraphConfig` 🤔

moisses89: Understood but would you find other way to avoid chose between networks in the endpoint?

fmrsabino: Restructured the control flow in [41d316c](https://github.com/safe-global/safe-transaction-service/pull/2105/commits/41d316c640eb4fdb29d83c8a1f67a142d6930cde)

Given that the chain id is no longer required I think the current state is now clearer.

PR #2112, opened 2024-07-02
    Optimize Module Transaction fetching

Uxio0:
- Modify ModuleTransaction model:
  - Create an index for `safe` and `created`, including `internal_tx_id` for JOIN
  - Remove `safe` index, as the previous index will be used
  - Refactor unique id calculation inside of model
- Update ModuleTransaction query for all-transactions to use `created` instead of `internal_tx__created` field. It has the same value and can use indexes
- Rename ModuleTransaction view to follow the same convention


Uxio0:
@JagoFigueroa for QA some endpoints need to be checked, all with ModuleTransactions:
- All-transactions
- List of module transactions for a Safe
- Getting a module transaction by unique id

safe_transaction_service/history/services/transaction_service.py
```diff
@@ -210,7 +210,7 @@ def get_all_tx_identifiers(
         module_tx_ids = (
             ModuleTransaction.objects.filter(safe=safe_address)
             .annotate(
-                execution_date=F("internal_tx__timestamp"),
+                execution_date=F("created"),
```
moisses89, line 213: `created` field is not the same than the block timestamp, it is a auto created field and would be the date when the transaction was indexed. 
Could we have issues because there are inserted transactions out of order? I mean recovered transactions because a reindex.

Uxio0: It's the same due to our implementation: https://github.com/safe-global/safe-transaction-service/blob/c9f762bbc53a5afd4895992e059b6034d5559916/safe_transaction_service/history/indexers/tx_processor.py#L535

moisses89: Ok I see but is a little bit confusing that we are using AutoCreatedField. https://github.com/safe-global/safe-transaction-service/blob/4631ecd691e32b4b1cf1309c44b0eabd4a8a1e27/safe_transaction_service/history/models.py#L1512
Maybe we should thinking about add the created field to the model and remove this dependency?

Uxio0: I agree and that makes sense, but we are doing the same for other models, so right now I would leave it that way

moisses89 (approved):

JagoFigueroa:
Looking good to me! 😄

Uxio0:
Thanks!

PR #2116, opened 2024-07-02
Allow cross chain Delegate registration

### LINKED ISSUE #2081: Delegates signing structure should allow signing all chains
# What is needed?
Currently, for adding delegates a `EIP712` structure with the `chainId` must be signed. But if the `chainId: 0`, that signature should be valid for every network (currently `0` is taken as a valid `chainId`).

# Background
Instead of the need to sign X signatures for X transaction services, only 1 signature would be valid for every tx service.

This was requested by @DmitryBespalov , and I think it makes a lot of sense so we can deprecate the old `delegate` endpoints

fmrsabino:
### What was wrong? 👾

Closes #2081

### How was it fixed? 🎯

- Adds support for signatures of EIP-712 payloads which might not include the Chain ID.
- Signatures of payloads which included the Chain ID are still valid.

### DIFF
--- a/safe_transaction_service/history/serializers.py
+++ b/safe_transaction_service/history/serializers.py
@@ -1,8 +1,9 @@
+import itertools
 import json
+import logging
 from enum import Enum
 from typing import Any, Dict, List, Optional
 
-from django.conf import settings
 from django.http import Http404
 
 from drf_yasg.utils import swagger_serializer_method
@@ -48,6 +49,8 @@ from .models import (
     TransferDict,
 )
 from .services.safe_service import SafeCreationInfo
+
+logger = logging.getLogger(__name__)
 
 
 def get_data_decoded_from_data(data: bytes, address: Optional[ChecksumAddress] = None):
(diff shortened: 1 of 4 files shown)


safe_transaction_service/history/tests/test_views_v2.py
```diff
@@ -461,7 +513,7 @@ def test_delegate_delete(self):
             )
             self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
             self.assertIn(
-                "Signature is not valid",
```
fmrsabino, line 464: This error message is no longer emitted. Since a signature can be valid with our without a Chain ID, we do not return early with this error.

safe_transaction_service/history/serializers.py
```diff
@@ -406,14 +410,18 @@ def validate_delegator_signature(
     ) -> bool:
         ethereum_client = EthereumClientProvider()
         chain_id = ethereum_client.get_chain_id()
+
         # Accept a message with the current topt and the previous totp (to prevent replay attacks)
-        for previous_totp in (True, False):
+        for previous_totp, chain_id in list(
+            itertools.product((True, False), (chain_id, None))
+        ):
```
fmrsabino, line 417: Since we now consider the `chain_id` to be an optional part of the payload to be signed, we do the cross product between the `chain_id` and the `previous_totp`. So we validate the signature against:
- Chain ID, True (previous totp)
- Chain ID, False
- No Chain ID, True
- No Chain ID, False

safe_transaction_service/history/serializers.py
```diff
@@ -406,14 +410,18 @@ def validate_delegator_signature(
...
             message_hash = DelegateSignatureHelperV2.calculate_hash(
-                delegate, chain_id, previous_totp=previous_totp
+                delegate, chain_id=chain_id, previous_totp=previous_totp
             )
             safe_signatures = SafeSignature.parse_signature(signature, message_hash)
             if not safe_signatures:
-                raise ValidationError("Signature is not valid")
```
falvaradorodriguez, line 416: Why this change? IMO, If the request doesn't have signatures will be right to raise an exception

`SafeSignature.parse_signature` would return a safe_signature in both cases, with our without `chain_id`

falvaradorodriguez (changes requested):

safe_transaction_service/history/serializers.py
fmrsabino: > Why this change? IMO, If the request doesn't have signatures will be right to raise an exception
> 
> `SafeSignature.parse_signature` would return a safe_signature in both cases, with our without `chain_id`

Good point. I mistakenly assumed that `parse_signature` would not return any signature if the signature was not for the `message_has`. That validation happens further down in the control flow. Will fix 👍

fmrsabino: Changed in [cc92b7a](https://github.com/safe-global/safe-transaction-service/pull/2116/commits/cc92b7a1fc3848b5b36193d53c91dbbe4e7bc295)

falvaradorodriguez (approved):
🚀

PR #2119, opened 2024-07-03
4337: Allow confirmations for not initialized Safes

Uxio0:
- Previously, if a Safe was not deployed no more confirmations were allowed.
- After this PR the owners encoded in the `initCode` will be checked if the Safe is not deployed.

### DIFF
--- a/safe_transaction_service/account_abstraction/serializers.py
+++ b/safe_transaction_service/account_abstraction/serializers.py
@@ -32,20 +32,26 @@ from .models import UserOperation as UserOperationModel
 # ================================================ #
 class SafeOperationSignatureValidatorMixin:
     """
-    Mixin class to validate SafeOperation signatures. _get_owners must be overridden to define
+    Mixin class to validate SafeOperation signatures. `_get_owners` can be overridden to define
     the valid owners to sign
     """
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.ethereum_client = EthereumClientProvider()
+        self._deployment_owners: List[ChecksumAddress] = []
 
     def _get_owners(self, safe_address: ChecksumAddress) -> List[ChecksumAddress]:
         """
         :param safe_address:
-        :return: Valid owners to sign the SafeOperation
+        :return:  `init_code` decoded owners if Safe is not deployed or current blockchain owners if Safe is deployed
         """
-        raise NotImplementedError("Must be implemented in subclass")
+        try:
+            return get_safe_owners(safe_address)
+        except ValidationError as exc:
+            if self._deployment_owners:
+                return self._deployment_owners
+            raise exc
 
     def _validate_signature(
         self,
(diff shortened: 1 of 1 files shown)


Uxio0:
@JagoFigueroa for QA contact with @yagopv , as this was reported by him

safe_transaction_service/account_abstraction/serializers.py
```diff
@@ -32,20 +32,26 @@
...
 
     def _get_owners(self, safe_address: ChecksumAddress) -> List[ChecksumAddress]:
         """
         :param safe_address:
-        :return: Valid owners to sign the SafeOperation
+        :return:  `init_code` decoded owners if Safe is not deployed or current blockchain owners if Safe is deployed
         """
-        raise NotImplementedError("Must be implemented in subclass")
+        try:
+            return get_safe_owners(safe_address)
```
moisses89, line 50: I'm thinking that a better approach would be check if there is `init_code`, as we validate [here](https://github.com/safe-global/safe-transaction-service/blob/229d4bf6a533e1819cb84ccf2c5d26c7a67c5f20/safe_transaction_service/account_abstraction/serializers.py#L128) When `init_code` is missing means not deployed address.
At this way we will reduce the number of calls to the RPC node.

safe_transaction_service/account_abstraction/serializers.py
```diff
@@ -363,6 +352,13 @@ def validate(self, attrs):
                 f"SafeOperation with hash={safe_operation_hash_hex} does not exist"
             )
 
+        # Parse valid owners from init code
+        if user_operation_model.init_code:
+            decoded_init_code = decode_init_code(
+                bytes(user_operation_model.init_code), self.ethereum_client
+            )
+            self._deployment_owners = decoded_init_code.owners
```
moisses89, line 360: Thinking in the future what do you think about to have something like a `dataclass` for the `init_code` and refactor this, I could see that we are doing the same here [safe_transaction_service/account_abstraction/serializers.py:146](https://github.com/safe-global/safe-transaction-service/blob/229d4bf6a533e1819cb84ccf2c5d26c7a67c5f20/safe_transaction_service/account_abstraction/serializers.py#L146)

safe_transaction_service/account_abstraction/serializers.py
```diff
@@ -32,20 +32,26 @@
 # ================================================ #
 class SafeOperationSignatureValidatorMixin:
     """
-    Mixin class to validate SafeOperation signatures. _get_owners must be overridden to define
+    Mixin class to validate SafeOperation signatures. `_get_owners` can be overridden to define
```
moisses89, line 35:
```suggestion
    Mixin class to validate SafeOperation signatures. 
```
A developer knows that he can override a function class :)

moisses89 (approved):
I wrote some suggestions, let me know what do you think.

safe_transaction_service/account_abstraction/serializers.py
Uxio0: That's not true. You can send `init_code` with the address deployed, that's why I did it this way

safe_transaction_service/account_abstraction/serializers.py
Uxio0: What do you mean? `Decode` already returns a `dataclass`

safe_transaction_service/account_abstraction/serializers.py
Uxio0: Haha not always 😉

safe_transaction_service/account_abstraction/serializers.py
moisses89: You are right, then makes sense to have this field `_deployment_owners` int he `SafeOperationSignatureValidatorMixin`? 
For me doesn't look part of this class, I would say that would be better use directly from the decode init code Dataclass.

Uxio0: It has to be there, as the default `_get_owners` function is using it, so it needs to be declared

PR #2120, opened 2024-07-03
Optimize out of order tx processing

Uxio0:
- When processing transactions, if there is a reindex a old transaction can appear between the already processed ones
- Previously, the solution for that was removing status for the Safe and then process all the transactions again
- This was not optimal at all for Safes with a high number of transactions, so this PR will only reprocess transactions from the problematic timestamp

### DIFF
--- a/safe_transaction_service/history/services/index_service.py
+++ b/safe_transaction_service/history/services/index_service.py
@@ -13,6 +13,7 @@ from gnosis.eth import EthereumClient, EthereumClientProvider
 from ..models import EthereumBlock, EthereumTx
 from ..models import IndexingStatus as IndexingStatusDb
 from ..models import (
+    InternalTx,
     InternalTxDecoded,
     ModuleTransaction,
     MultisigConfirmation,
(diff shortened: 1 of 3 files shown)


Uxio0:
@moisses89 we should check in production after deploying this

falvaradorodriguez (approved):

safe_transaction_service/history/services/index_service.py
```diff
@@ -353,7 +353,38 @@ def _reprocess(self, addresses: List[str]):
...
 
-    def reprocess_addresses(self, addresses: List[str]):
+    @transaction.atomic
+    def fix_out_of_order(self, address: ChecksumAddress, internal_tx: InternalTx):
+        """
+        Fix a Safe that has transactions out of order (not processed transactions in between processed ones, usually
+        due a reindex)
+
+        :param address:
+        :return:
```
fmrsabino, line 363: nitpick: we are missing the `internal_tx` here

fmrsabino (approved):

safe_transaction_service/history/services/index_service.py
Uxio0: You are right, done!

PR #2129, opened 2024-07-09
Migrate from `EthereumClientProvider`

Uxio0:
- Related to https://github.com/safe-global/safe-eth-py/pull/1209


falvaradorodriguez:
Changes related to EthereumClientProvider LGTM, but other changes in the new version of `safe-eth-py` might not be compatible.

falvaradorodriguez (approved):

PR #2150, opened 2024-07-19
Remove webhooks support

### LINKED ISSUE #2149: Remove unused webhooks
# What is needed?
Webhooks logic were migrated to safe-events-service, but the safe transaction service continues creating and running tasks that don't do anything. 
We should remove webhooks and ensure that this tasks are never executed.

moisses89:
Closes #2149


safe_transaction_service/history/migrations/0058_alter_webhook_url.py
```diff
@@ -16,9 +16,7 @@ class Migration(migrations.Migration):
             name="url",
             field=models.CharField(
                 max_length=255,
-                validators=[
-                    safe_transaction_service.history.models._validate_webhook_url
-                ],
+                validators=[safe_transaction_service.history.utils.validate_url],
```
moisses89, line 19: @safe-global/core-api not sure if better approach would be remove this validator because the `_validate_webhook_url` was used just here.

Uxio0: Sorry, I'm not understanding this

safe_transaction_service/history/models.py
```diff
@@ -2043,7 +2042,7 @@ def previous(self) -> Optional["SafeStatus"]:
         )
 
 
-class WebHookType(Enum):
+class NotificationEventType(Enum):
```
Uxio0, line 2045:
```suggestion
class TransactionServiceEventType(Enum):
```

maybe?

safe_transaction_service/history/services/notification_service.py
```diff
@@ -25,7 +25,7 @@
 from safe_transaction_service.utils.ethereum import get_chain_id
 
 
-def build_webhook_payload(
+def build_message_payload(
```
Uxio0, line 28:
```suggestion
def build_event_payload(
```

Uxio0 (approved):

safe_transaction_service/history/migrations/0058_alter_webhook_url.py
moisses89: I mean, the function looks used just in the migration and maybe each time that a webhook is inserted, as we are removing the model this does not make sense, I just leave it because if not we will break the django migrations history. 
But how we are squashing the migrations in other PR this question is not relevant :)

Uxio0: Oh ok, makes sense!

PR #2176, opened 2024-08-08
Index internal txs in batch for safe events

Uxio0:
- Just leaving here as draft, as it does not get the speedup we were expecting


Uxio0:
Closed due to #2194 being a better solution

PR #2194, opened 2024-08-19
Optimize safe events indexer

### LINKED ISSUE #2181: Optimize  ´_process_decoded_element´
# What is needed?
With large number of events the decoded process is taking long times causing indexing delays. 
The main reason looks to be `_process_decoded_element`, then we should find out how to optimize this function.

moisses89:
Closes #2181
# Description 
This PR reduce the number of queries used to process the `SafeSetup` and `ProxyCreation` events, taking the advantage of commonly this two events are together in the same transaction or close in a batch of blocks. 
Is important to highlight if the first event is `SafeSetup` will be always together with a `ProxyCreation` event because is a result of call `DeployProxy` with initializer.
  
# Benchmark
The following sctipt was used to compare the results between the previous implementation and the implementation of this PR.
```
        SafeMasterCopy("0x29fcB43b46531BcA003ddC8FCB67FFE91900C762").save()
        SafeMasterCopy("0x3E5c63644E683549055b9Be8653de26E0B4CD36E").save()
        addresses = set(SafeMasterCopy.objects.all().values_list("address", flat=True))
        indexer = SafeEventsIndexerProvider()

        start_block = 1799327
        relevant = indexer.find_relevant_elements(
            addresses, start_block, start_block + 100
        )
        start = time.time()
        indexer.process_elements(relevant)
        print(f"Total time for {len(relevant)} {time.time() - start}")
``` 
Also used the same chain and RPC in both chains. 
## Results
Previous implementation:
```
2024-08-23 08:51:03,575 [INFO] [MainProcess] SafeEventsIndexer: Found 9288 events from block-number=1799327 to block-number=1799427 for 2 addresses
2024-08-23 08:51:11,350 [INFO] [MainProcess] Opening connection to Redis
Total time for process decoded elements 47.67031955718994
Total time for 9288 55.48148560523987
```
Current implementation:
```
2024-08-23 08:47:00,677 [INFO] [MainProcess] SafeEventsIndexer: Found 9288 events from block-number=1799327 to block-number=1799427 for 2 addresses
2024-08-23 08:47:10,637 [INFO] [MainProcess] Opening connection to Redis
2024-08-23 08:47:11,468 [INFO] [MainProcess] Inserted 9288 internal_txs 
2024-08-23 08:47:11,468 [INFO] [MainProcess] Inserted 4644 internal_decoded_txs
Total time for process decoded elements 4.111443758010864
Total time for 9288 11.770971059799194
```


safe_transaction_service/history/indexers/safe_events_indexer.py
```diff
@@ -485,3 +478,137 @@ def _process_decoded_element(
...
+                            # Usually SafeSetup is the first event and next is ProxyCreation when ProxyCreation is called with initializer.
+                            proxy_creation_event = events[1]
+                        elif (
+                            event_position == 1
+                            and events[0]["event"] == "ProxyCreation"
+                        ):
+                            # ProxyCreation first and SafeSetup later
+                            proxy_creation_event = events[1]
+                    else:
+                        # Proxy was created in previous blocks.
```
moisses89, line 543: The previous implementation always tried to remove the `InternalTx` generated from `ProxyCreation` just in case the `ProxyCreation` came first. https://github.com/safe-global/safe-transaction-service/blob/main/safe_transaction_service/history/indexers/safe_events_indexer.py#L366 
 In this case, because the `ProxyCreation` never comes later, the `InternalTx` generated by `SafeSetup` leaves without singleton (to) forever.  
 Maybe would be better try to get the Proxy InternalTx from database to insert the correct parameters of `SafeSetup`  what do you think? @safe-global/core-api 
 In the current implementation the `ProxyCreation` is storing the singleton addres in the `to` field of `InternalTx`, not sure if is completely right.

safe_transaction_service/history/indexers/safe_events_indexer.py
```diff
@@ -485,3 +478,137 @@ def _process_decoded_element(
...
+                        proxy_creation_event = None
+                        # TODO store decoded ProxyCreation event to get the singleton address
```
Uxio0, line 545: This was working before, right?

safe_transaction_service/history/indexers/safe_events_indexer.py
```diff
@@ -485,3 +478,137 @@ def _process_decoded_element(
...
+                            and events[1]["event"] == "ProxyCreation"
+                        ):
+                            # Usually SafeSetup is the first event and next is ProxyCreation when ProxyCreation is called with initializer.
+                            proxy_creation_event = events[1]
+                        elif (
+                            event_position == 1
+                            and events[0]["event"] == "ProxyCreation"
+                        ):
+                            # ProxyCreation first and SafeSetup later
+                            proxy_creation_event = events[1]
```
Uxio0, line 541: Am I mistaken or there's a typo?

```suggestion
                            proxy_creation_event = events[0]
```

If this was not detected by the tests a test should be added for this case

safe_transaction_service/history/indexers/safe_events_indexer.py
```diff
@@ -311,65 +358,23 @@ def _process_decoded_element(
...
-                # creates a Safe and configure it in the next transaction. Remove it if that's the case
-                InternalTx.objects.filter(contract_address=safe_address).delete()
-                internal_tx.contract_address = safe_address
-                internal_tx_decoded.function_name = "setup"
-                args["payment"] = 0
-                args["paymentReceiver"] = NULL_ADDRESS
-                args["_threshold"] = args.pop("threshold")
-                args["_owners"] = args.pop("owners")
+        if event_name == "ProxyCreation" or event_name == "SafeSetup":
+            # Will ignore this events because were indexed in process_safe_creation_events
```
Uxio0, line 375:
```suggestion
            # Will ignore these events because were indexed in process_safe_creation_events
```

safe_transaction_service/history/indexers/safe_events_indexer.py
```diff
@@ -485,3 +478,137 @@ def _process_decoded_element(
...
+                        tx_type=InternalTxType.CREATE.value,
+                        call_type=None,
+                    )
+                    internal_txs.append(internal_tx)
+                else:
+                    logger.error(f"Event is not a Safe creation event {event['event']}")
+
+        with transaction.atomic():
+            InternalTx.objects.bulk_create(internal_txs)
+            InternalTxDecoded.objects.bulk_create(internal_decoded_txs)
```
Uxio0, line 592: What happens if they are already inserted?

safe_transaction_service/history/indexers/safe_events_indexer.py
```diff
@@ -485,3 +478,137 @@ def _process_decoded_element(
...
+            logger.info(f"Inserted {len(internal_txs)} internal_txs ")
+            logger.info(f"Inserted {len(internal_decoded_txs)} internal_decoded_txs")
+
+        return internal_txs
+
+    def _process_decoded_elements(self, decoded_elements: list[EventData]) -> List[Any]:
+        processed_elements = []
+        # Extract Safe creation events from decoded_elements list
```
Uxio0, line 600:
```suggestion
        # Extract Safe creation events by Safe from decoded_elements list
```

Uxio0 (changes requested):

safe_transaction_service/history/indexers/safe_events_indexer.py
moisses89: yes, in that case the previous implementation did a delete to remove the proxyCreation event in the InternalTx, but I don't understand why.

Uxio0: Not sure why was the reason, but I bet it was due to testing and no one was complaining about how it worked, so in my opinion we should keep the same logic

safe_transaction_service/history/indexers/safe_events_indexer.py
moisses89: You were right and I added a proper test.

safe_transaction_service/history/indexers/safe_events_indexer.py
```diff
@@ -287,6 +287,54 @@ def _is_setup_indexed(self, safe_address: ChecksumAddress):
...
+            call_type=EthereumTxCallType.CALL.value,
+            trace_address=default_trace_address,
+            error=None,
+        )
+        # Overriding passed keys
+        for key, value in kwargs.items():
+            if hasattr(internal_tx, key):
+                setattr(internal_tx, key, value)
+            else:
+                raise AttributeError(f"Invalid atribute {key} for InternalTx")
```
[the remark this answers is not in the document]
moisses89: No strong opinion, this will raise the exception if any developer set wrong keyword.

safe_transaction_service/history/indexers/safe_events_indexer.py
moisses89: Raise an exception.
Ignoring conflicts in such case would fix the issue, https://github.com/safe-global/safe-transaction-service/pull/2194/commits/5eeabd4683d121afcf6d0247dc71152282efa1a6

moisses89: @Uxio0 `ignore_conflicts` fail because `InternalDecodedTx` have relation with `internal_tx`, basically the ORM try to prevent lose data.
We can use instead of that update_conflicts, but I'm concerned because I just discovered that bulk_create does not call save, and in that case post_save will not be fired. 
Will this affect our signals handle?

safe_transaction_service/history/indexers/safe_events_indexer.py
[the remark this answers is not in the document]
Uxio0: No it doesn't, you can use setattr in any object with any key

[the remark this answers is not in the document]
moisses89: right, is why I'm raising on the else of  ` if hasattr(internal_tx, key):`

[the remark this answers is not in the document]
Uxio0: Missed that, sorry Moisés!

safe_transaction_service/history/indexers/safe_events_indexer.py
Uxio0: `bulk_create` does in case we inherit our BulkCreateMixin

moisses89: I did a workaround in case that would be a collision inserting the same instance, shouldn't be any common because the bulk create is in atomic section and we are checking if the contract were indexed, basically if the `setup` and `proxyevent` were correctly inserted, shouldn't reach the insert section if would be indexed again.

safe_transaction_service/history/indexers/safe_events_indexer.py
moisses89: Done, I'm removing previous InternalTx.

safe_transaction_service/history/indexers/safe_events_indexer.py
```diff
@@ -485,3 +478,179 @@ def _process_decoded_element(
...
+                        logger.info(
+                            "Ignoring already processed InternalDecodedTx with tx-hash=%s and trace-address=%s",
+                            internal_decoded_tx.internal_tx.ethereum_tx_id.hex(),
+                            internal_decoded_tx.internal_tx.trace_address,
+                        )
+                        pass
+
+        return stored_internal_txs
+
+    def _process_decoded_elements(self, decoded_elements: list[EventData]) -> List[Any]:
```
Uxio0, line 638:
```suggestion
    def _process_decoded_elements(self, decoded_elements: list[EventData]) -> list[Any]:
```

If you are using `list` as type, you should use it everywhere 😉

safe_transaction_service/history/indexers/safe_events_indexer.py
```diff
@@ -485,3 +478,179 @@ def _process_decoded_element(
...
+
+        # Store which internal txs where inserted into database
+        stored_internal_txs: List[InternalTx] = []
+        with transaction.atomic():
+            try:
+                InternalTx.objects.bulk_create(internal_txs)
+                InternalTxDecoded.objects.bulk_create(internal_decoded_txs)
+                stored_internal_txs = internal_txs
+            except IntegrityError:
+                pass
```
Uxio0, line 608:
```suggestion
                logger.info("Cannot bulk create the provided Safe Creation Events, trying one by one")
```

Uxio0 (approved):

Uxio0 (approved):

PR #2208, opened 2024-09-17
Fix swagger locally is not working

moisses89:
# Description
Try endpoints from swagger always failed because the port was not included.
image

# Other issues
-  X_FORWARDED_HOST were configured in  nginx but were not enabled in Django.
- $server_name is not configured, using the host instead.


falvaradorodriguez (approved):
very useful change!

Uxio0:
Well done

PR #2220, opened 2024-09-23
Refactor processing of Internal Txs

### LINKED ISSUE #2214: Multiple indexing issues detected on Optimism
We detected multiple indexing issues:

- A Safe that migrates from `L1` to `L2` will always trigger and error and then call `reindex_master_copies` when processing, as the status is always corrupt before the `nonce` the Safe had when it was migrated. Transactions are missing as the events were not emitted, and service tries to find the missing events forever.
- `InternalTxDecoded.objects.safes_pending_to_be_processed()` is really slow, taking more than 50 seconds in network like Optimism. That prevents a quick processing of Safes.
- Banned Safes don't mark `InternalTxDecoded` as processed, making the index and the table grow but never consumed.

Uxio0:
- Task was refactored into a function in `IndexService`
- Automatic reindexing of Safes was disabled. Reindexing issues should be fixed by the scheduled task `reindex_mastercopies_last_hours_task`. Also some Safes were corrupted due to users trying to trick the indexer or Safes converted from L1 -> L2, and reindexing was triggered forever
- Closes #2214

### DIFF
--- a/safe_transaction_service/history/services/index_service.py
+++ b/safe_transaction_service/history/services/index_service.py
@@ -67,6 +67,7 @@ class IndexServiceProvider:
                 get_auto_ethereum_client(),
                 settings.ETH_REORG_BLOCKS,
                 settings.ETH_L2_NETWORK,
+                settings.ETH_INTERNAL_TX_DECODED_PROCESS_BATCH,
             )
         return cls.instance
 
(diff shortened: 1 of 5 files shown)


safe_transaction_service/history/tasks.py
```diff
@@ -272,6 +263,29 @@ def process_decoded_internal_txs_task(self) -> Optional[int]:
...
+
+    :param safe_address:
+    :param reindex_master_copies: Trigger auto reindexing if a problem is found
+    :return: Number of `InternalTxDecoded` processed
+    """
+    with contextlib.suppress(LockError):
+        with only_one_running_task(self, lock_name_suffix=safe_address):
+            logger.info("[%s] Start processing decoded internal txs", safe_address)
+            index_service: IndexService = IndexServiceProvider()
+            number_processed = index_service.process_address(safe_address)
```
moisses89, line 282:
```suggestion
            number_processed = index_service.process_decoded_txs(safe_address)
```
process_address sound for me too generic and confusing.

safe_transaction_service/history/services/index_service.py
```diff
@@ -388,6 +399,38 @@ def fix_out_of_order(
...
+        if InternalTxDecoded.objects.out_of_order_for_safe(safe_address):
+            logger.error("[%s] Found out of order transactions", safe_address)
+            self.fix_out_of_order(
+                safe_address,
+                InternalTxDecoded.objects.pending_for_safe(safe_address)[0].internal_tx,
+            )
+            self.tx_processor.clear_cache(safe_address)
+
+        # Use chunks for memory issues
+        number_processed = 0
```
moisses89, line 420:
```suggestion
        total_processed_txs = 0
```
For me number processed could indicate an ID for example, not the total amount of processed transactions.

moisses89 (changes requested):

Uxio0:
I've just applied your suggestions!

PR #2233, opened 2024-09-27
Decode creation data using MultiSend

### LINKED ISSUE #2197: Add saltNonce to api/v1/safes/<safe_address>/creation endpoint
# What is needed?
Research if is possible to add the saltNonce to the creation endpoint without affecting the endpoint performance too much. 
Maybe indexing it or getting the transaction data from the RPC.

Uxio0:
- Add `saltNonce` to the `/creation` endpoint, closes #2197
- Decode creation data for Safes created using `MultiSend`


safe_transaction_service/history/services/safe_service.py
```diff
@@ -45,10 +46,18 @@ class SafeCreationInfo:
...
+    salt_nonce: Optional[int]
     transaction_hash: str
     user_operation: Optional[aa_models.UserOperation]
 
 
+@dataclass
+class ProxyCreationData:
+    singleton: ChecksumAddress
+    initializer: bytes
+    salt_nonce: Optional[int]
```
moisses89, line 58: Should be Char as the serializer is expecting?

safe_transaction_service/history/services/safe_service.py
```diff
@@ -112,25 +121,28 @@ def get_safe_creation_info(self, safe_address: str) -> Optional[SafeCreationInfo
...
-                data = bytes(data_tx.data)
-                result = self._decode_proxy_factory(
-                    data
-                ) or self._decode_cpk_proxy_factory(data)
-                if result:
-                    master_copy, setup_data = result
-            if not (master_copy and setup_data):
+            results = self._decode_creation_data(data_tx.data)
+            if len(results) == 1:
+                # TODO Support multiple Proxy deployments in the same transaction
```
moisses89, line 135: I would expect this happen, I agree with you on do it in different PR but could you create an issue to track it?

moisses89 (approved):

safe_transaction_service/history/services/safe_service.py
Uxio0: Yes, being a `uint256`

safe_transaction_service/history/services/safe_service.py
Uxio0: Agree: https://github.com/safe-global/safe-transaction-service/issues/2250

PR #2258, opened 2024-10-15
Fix caching issues on tx-processor

Uxio0:
- TxProcessor was shared between tasks, so cache was deleted for all of them
- Now only the involved Safe is deleted from the processor
- If there's an error processing a transaction, previously cache could become broken. Now a `finally` is used to make sure it's always cleaned

### DIFF
--- a/safe_transaction_service/history/indexers/tx_processor.py
+++ b/safe_transaction_service/history/indexers/tx_processor.py
@@ -325,9 +325,15 @@ class SafeTxProcessor(TxProcessor):
     def process_decoded_transaction(
         self, internal_tx_decoded: InternalTxDecoded
     ) -> bool:
-        processed_successfully = self.__process_decoded_transaction(internal_tx_decoded)
-        internal_tx_decoded.set_processed()
-        self.clear_cache()
+        contract_address = internal_tx_decoded.internal_tx._from
+        self.clear_cache(safe_address=contract_address)
+        try:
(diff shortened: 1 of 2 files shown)


safe_transaction_service/history/indexers/tx_processor.py
```diff
@@ -325,9 +325,15 @@ def store_new_safe_status(
     def process_decoded_transaction(
         self, internal_tx_decoded: InternalTxDecoded
     ) -> bool:
-        processed_successfully = self.__process_decoded_transaction(internal_tx_decoded)
-        internal_tx_decoded.set_processed()
-        self.clear_cache()
+        contract_address = internal_tx_decoded.internal_tx._from
+        self.clear_cache(safe_address=contract_address)
```
falvaradorodriguez, line 329: Would it be enough to clear the cache in the `finally` part?

Uxio0: I'm worried about one process interrupted or a task killed and not executing the finally part

Uxio0: Do you think it's overkilling?

falvaradorodriguez: Taking into account your previous comment, maybe it's necessary. In my first review, I found it a bit strange.

falvaradorodriguez (approved):

PR #2259, opened 2024-10-15
Add countdown to delay reindex contracts

### LINKED ISSUE #2240: Reindex contracts get timeout for each delayed task due the amount of requests.
# Description
`reindex_contracts_without_metadata_task` search in Contracts database, for the contracts with missing ABIs and delay the task to be executed by celery. 
Depending on the configuration of `TASK_CONCURRENCY`, we would be able to request a big amount of contracts in a short times to the thirdparty services that offers the ABI of the contract and this could cause that the service cannot respond in time and reach the 60 seconds of timeout. 
The following picture show a case:
image

This assumption were demonstrated repeating with the same amount of contracts and giving a delay between contract requests of 3 seconds, every task finalized successfully.  

# Proposal improvement
Instead throw all the contracts to celery immediately, do it batching them and delay each batch some seconds. 
This solution could have two configurations the BATCH_SIZE and the DELAY_TIME (time delayed between batch) 
At this way we will request contract information in a friendly way.

moisses89:
# Description
It adds batch configuration to avoid requesting too many contracts metadata simultaneously to third parties (sourcify, etherscan).

Closes #2240


config/settings/base.py
```diff
@@ -590,3 +590,11 @@
 
 # Compression level – an integer from 0 to 9. 0 means not compression
 CACHE_ALL_TXS_COMPRESSION_LEVEL = env.int("CACHE_ALL_TXS_COMPRESSION_LEVEL", default=0)
+
+# Contracts configuration
+REINDEX_CONTRACTS_METADATA_BATCH = env.int(
+    "REINDEX_CONTRACTS_METADATA_BATCH", default=100
+)
```
Uxio0, line 597: Add a comment explaining a little

config/settings/base.py
```diff
@@ -590,3 +590,11 @@
...
+REINDEX_CONTRACTS_METADATA_COUNTDOWN = env.int(
```
Uxio0, line 598: Add a comment explaining a little

safe_transaction_service/contracts/tasks.py
```diff
@@ -98,16 +99,22 @@ def reindex_contracts_without_metadata_task() -> int:
...
         logger.info("Reindexing contract %s", address)
         create_or_update_contract_with_metadata_task.apply_async(
-            (address,), priority=1
+            (address,), priority=1, countdown=countdown
         )  # Lowest priority
-        i += 1
-    return i
+        total_addresses_reindexed += 1
+        countdown += (
+            countdown_increment if total_addresses_reindexed % batch_size == 0 else 0
```
Uxio0, line 115: If you reindex `50` addresses but the batch size is `100`, module is `50`, so you are increasing the countdown without reason

Uxio0 (changes requested):

safe_transaction_service/contracts/tasks.py
moisses89: It is increasing the coundown just when the module is equal to 0, so multiples of the `batch_size`.  
In your example, should be when `total_addresses_reindexed` is 100, 200, 300...

Uxio0 (approved):

PR #2265, opened 2024-10-17
Fix `null` confirmationsRequired

### LINKED ISSUE #2170: `confirmationsRequired` is `null`
**Describe the bug**

The multisig `confirmationsRequired` field sometimes returns `null`.

**To Reproduce**

Unknown.

**Expected behavior**

`confirmationsRequired` is an integer.

Frequently seen on staging/prod. Optimism, but occassionally on Mainnet.

Uxio0:
Fix `null` confirmationsRequired on txs endpoint

Now, for an historical transaction, we follow the same rules:
- If there's an entry on `SafeStatus` with matching `nonce` and `safe_tx_hash`, it's returned.
- If not, if there's an entry on `SafeLastStatus` with matching `nonce` and `safe_tx_hash`, it's returned.
- Otherwise, the number of confirmations for that transaction is returned.

Closes #2170


falvaradorodriguez (approved):
LGTM!

PR #2277, opened 2024-10-23
Load logo URIs from lists

### LINKED ISSUE #2103: [ API ] Get token logos from trusted lists
# What is needed?
Currently the logos need to be uploaded manually in the transaction service by the admin page. 
This is hard to manage when we need to update large quantity of tokens. 
My suggestion is take advantage of trusted lists that are configured in the service and get the logos from there when there is not any logo upload for the token. 
At this way the service would need less maintenance updating tokens logos. 

Example of list: https://tokens.coingecko.com/uniswap/all.json.

Uxio0:
- Add a `logo_uri` field to `Token`. If `logo` is not present, that URL will be used if defined.
- Closes #2103


safe_transaction_service/tokens/tasks.py
```diff
@@ -75,29 +94,21 @@ def update_token_info_from_token_list_task() -> int:
...
+        return 0
+
+    tokens_updated_count = 0
     with transaction.atomic():
         Token.objects.update(trusted=False)
-        return Token.objects.filter(address__in=token_addresses).update(trusted=True)
+        for token in filtered_tokens:
+            token_address = _parse_token_address_from_token_list(token["address"])
+            tokens_updated_count += Token.objects.filter(address=token_address).update(
+                logo_uri=token.get("logoURI") or "", trusted=True
```
falvaradorodriguez, line 112:
```suggestion
                logo_uri=token.get("logoURI", ""), trusted=True
```

falvaradorodriguez (approved):

safe_transaction_service/tokens/tasks.py
Uxio0: This will not work as expected. If in the JSON `logoUri` holds `null`, we will try to store a `None` instead of `empty string`

PR #2294, opened 2024-10-29
Add support for delegates to delete transactions proposed by themselves

### LINKED ISSUE #2284: Allow Proposal Delete from Delegate that created it
In the API docs for `v1_multisig-transactions_delete`

> Only the proposer can delete the transaction.
> If the transaction was proposed by a delegate, the Safe owner who delegated to the delegate must be used.

Since we have the proposer via #1592, the Delegate that proposed a transaction should be permitted to delete it. This would allow for automated management -- we have Discord-based management of some tasks, and being able to undo a mistake without an explicit reject transaction (and without using the parent key!) would be helpful.

falvaradorodriguez:
- Closes #2284
  - Delegate users can only delete transactions proposed by themselves.
  - Delegate users must still be valid at the time of deletion.

### DIFF
--- a/safe_transaction_service/history/serializers.py
+++ b/safe_transaction_service/history/serializers.py
@@ -571,19 +571,19 @@ class SafeMultisigTransactionDeleteSerializer(serializers.Serializer):
                 raise ValidationError("Only EOA and ETH_SIGN signatures are supported")
 
             # The transaction can be deleted by the proposer or by the delegate user who proposed it.
-            if safe_signature.owner == proposer:
+            owner = safe_signature.owner
+            if owner == proposer:
                 return attrs
-            if (
-                multisig_tx.proposed_by_delegate
-                and safe_signature.owner == multisig_tx.proposed_by_delegate
-            ):
+
+            proposed_by_delegate = multisig_tx.proposed_by_delegate
+            if proposed_by_delegate and owner == proposed_by_delegate:
                 delegates_for_proposer = (
                     SafeContractDelegate.objects.get_delegates_for_safe_and_owners(
                         safe_address, [proposer]
                     )
                 )
                 # Check if it's still a valid delegate.
-                if safe_signature.owner in delegates_for_proposer:
+                if owner in delegates_for_proposer:
                     return attrs
 
         raise ValidationError(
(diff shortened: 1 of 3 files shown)


safe_transaction_service/history/serializers.py
```diff
@@ -569,10 +569,26 @@ def validate(self, attrs):
                 SafeSignatureType.ETH_SIGN,
             ):
                 raise ValidationError("Only EOA and ETH_SIGN signatures are supported")
+
+            # The transaction can be deleted by the proposer or by the delegate user who proposed it.
             if safe_signature.owner == proposer:
                 return attrs
+            if (
+                multisig_tx.proposed_by_delegate
+                and safe_signature.owner == multisig_tx.proposed_by_delegate
```
Uxio0, line 578: IIRC `safe_signature.owner` involves computation. If you are using in multiple places please assign it to a variable first

safe_transaction_service/history/views.py
```diff
@@ -543,8 +543,8 @@ def get_queryset(self):
     def delete(self, request, safe_tx_hash: HexStr):
         """
         Removes the queued but not executed multi-signature transaction associated with the given Safe tansaction hash.
-        Only the proposer can delete the transaction.
-        If the transaction was proposed by a delegate, the Safe owner who delegated to the delegate must be used.
+        Only the proposer or the delegate user who proposed the transaction can delete it.
```
Uxio0, line 546:
```suggestion
        Only the proposer or the delegate who proposed the transaction can delete it.
```

safe_transaction_service/history/views.py
```diff
@@ -543,8 +543,8 @@ def get_queryset(self):
...
+        If the transaction was proposed by a delegate, must still be a valid delegate for the transaction proposer.
```
Uxio0, line 547:
```suggestion
        If the transaction was proposed by a delegate, it must still be a valid delegate for the transaction proposer.
```

Uxio0 (changes requested):

safe_transaction_service/history/serializers.py
falvaradorodriguez: Makes sense. Fixed [here](https://github.com/safe-global/safe-transaction-service/pull/2294/commits/3b80d157ad862d16ccae2182bbdacae4b0f3cec1)

Uxio0 (approved):

PR #2295, opened 2024-10-30
Move process_decoded_internal_txs_for_safe_task to processing queue

moisses89:
# Description
Adding the configurable queue names to be able to move some tasks to other workers. 
For now just process_safes have a different variable.


Uxio0 (changes requested):
Don't use environment variables. Queue names should be fixed and can be very confusing to be able to modify them

PR #2310, opened 2024-11-06
Add Delegates events on creation/deletion/update

### LINKED ISSUE #2299: Add Delegates events on creation/deletion/update
# What is needed?
Delegates _(proposers)_ is used on mobile clients only, and the traffic associated with the _delegates_ endpoint is low. However, after the delegates/proposers list is included in the Web UI, we expect a substantial increase in the traffic related to these endpoints.

The CGW strategy was not to cache the results for the delegates/proposers, but considering the traffic increase, it might make sense to add a caching/invalidation strategy. 

# Technical details
- Create a new event to signal the delegates/proposers' list has changed for a Safe address.
- Send the event to the general RabbitMQ, which is consumed by the CGW.

falvaradorodriguez:
- Closes #2299 

- Added the new event types: NEW_DELEGATE, UPDATED_DELEGATE and DELETED_DELEGATE


falvaradorodriguez:
I have continued the use of signals to manage the sending of these events to the queue because as we currently have multiple endpoints for delegate management the solution is simpler and clearer. At the moment, the delegates management is not high and there should not be performance issues with the use of signals.

safe_transaction_service/history/services/notification_service.py
```diff
@@ -204,3 +205,54 @@ def build_reorg_payload(block_number: int) -> ReorgPayload:
...
     )
+
+
+class DelegatePayload(TypedDict):
+    type: str
+    safeAddress: Optional[str]
+    delegate: str
+    delegator: str
+    label: str
+    expiryDate: Optional[int]
```
Uxio0, line 216: Maybe add the unit, for clarity. Like `expiryDateSeconds`

safe_transaction_service/history/services/notification_service.py
```diff
@@ -204,3 +205,54 @@ def build_reorg_payload(block_number: int) -> ReorgPayload:
...
+    """
+    Build a delegate payload with the specified event type and SafeContractDelegate instance data.
+
+    :param event_type: The transaction event type, restricted to NEW_DELEGATE, UPDATED_DELEGATE, or DELETED_DELEGATE.
+    :param instance: An instance of SafeContractDelegate.
+    :return: A DelegatePayload dictionary with details about the delegate.
+    """
+    return DelegatePayload(
+        type=event_type.name,
+        safeAddress=instance.safe_contract.address if instance.safe_contract else None,
```
Uxio0, line 236:
```suggestion
        safeAddress=instance.safe_contract_id if instance.safe_contract_id else None,
```

This can save a database lookup

safe_transaction_service/history/services/notification_service.py
```diff
@@ -204,3 +205,54 @@ def build_reorg_payload(block_number: int) -> ReorgPayload:
...
+        ),
+    )
+
+
+def build_save_delegate_payload(
+    instance: SafeContractDelegate, created: bool = True
+) -> DelegatePayload:
+    event_type = TransactionServiceEventType.NEW_DELEGATE
+    if not created:
+        event_type = TransactionServiceEventType.UPDATED_DELEGATE
```
Uxio0, line 251: Seems more clear this way

```suggestion
    if created:
        event_type = TransactionServiceEventType.NEW_DELEGATE
    else:
        event_type = TransactionServiceEventType.UPDATED_DELEGATE
```

Uxio0 (changes requested):

safe_transaction_service/history/services/notification_service.py
falvaradorodriguez: Done https://github.com/safe-global/safe-transaction-service/pull/2310/commits/e90cc8554af0d1d6fb5ada379215ba698099c7b2

Uxio0 (approved):

safe_transaction_service/history/services/notification_service.py
```diff
@@ -204,3 +205,55 @@ def build_reorg_payload(block_number: int) -> ReorgPayload:
...
     )
+
+
+class DelegatePayload(TypedDict):
+    type: str
+    safeAddress: Optional[str]
+    delegate: str
+    delegator: str
+    label: str
+    expiryDateSeconds: Optional[int]
```
iamacook, line 216: All of the existing events have a chain identifier. I suggest we add a `chainId` too, for example.

falvaradorodriguez: Yes, you are right. Good catch! 

Done https://github.com/safe-global/safe-transaction-service/pull/2310/commits/bd6310ee83c17efb76b4b1e0fbfd84775e71cdef

iamacook (approved):

PR #2311, opened 2024-11-06
Custom cache page

### LINKED ISSUE #2008: Refactor endpoints caching into a class
# What is needed?
Currently caching for endpoints is managed in different ways, could be good create a class to have together all the methods necessary for caching. 
We could extend the caching class for methods as multisig-transactions, module-transactions...

moisses89:
Closes #2008


safe_transaction_service/history/cache.py
```diff
@@ -0,0 +1,200 @@
...
+    # Cache tags
+    LIST_MULTISIGTRANSACTIONS_VIEW_CACHE_KEY = "multisigtransactionsview"
+    LIST_MODULETRANSACTIONS_VIEW_CACHE_KEY = "moduletransactionsview"
+    LIST_TRANSFERS_VIEW_CACHE_KEY = "transfersview"
+
+    def __init__(self, cache_tag: str, address: ChecksumAddress):
+        self.redis = get_redis()
+        self.address = address
+        self.cache_tag = cache_tag
+        self.cache_name = self.get_cache_name()
```
Uxio0, line 32: This can be a `cached property`

safe_transaction_service/history/cache.py
```diff
@@ -0,0 +1,200 @@
...
+    @cache
+    def is_enabled(self) -> bool:
+        """
+
+        :return: True if cache is enabled False otherwise
+        """
+        if settings.CACHE_VIEW_DEFAULT_TIMEOUT:
+            return True
+        else:
+            return False
```
Uxio0, line 53:
```suggestion
        return bool(settings.CACHE_VIEW_DEFAULT_TIMEOUT)
```

safe_transaction_service/history/cache.py
```diff
@@ -0,0 +1,200 @@
...
+        Calculate the cache_name from the cache_tag and address
+
+        :param cache_tag:
+        :param address:
+        :return:
+        """
+        return f"{self.cache_tag}:{self.address}"
+
+    @cache
+    def is_enabled(self) -> bool:
```
Uxio0, line 45: This should be `cached_property` called `enabled`

safe_transaction_service/history/cache.py
```diff
@@ -0,0 +1,200 @@
...
+    ModuleTransaction,
+    MultisigConfirmation,
+    MultisigTransaction,
+    TokenTransfer,
+)
+from safe_transaction_service.utils.redis import get_redis, logger
+
+
+class CacheSafeTxsView:
+    # Cache tags
```
Uxio0, line 23: Some docs explaining what is the purpose of this class would be very useful

safe_transaction_service/history/cache.py
```diff
@@ -0,0 +1,200 @@
...
+
+    :param timeout: Cache timeout in seconds.
+    :param cache_name: A unique identifier for the cache entry.
+    """
+
+    def decorator(view_func):
+        @wraps(view_func)
+        def _wrapped_view(request, *args, **kwargs):
+            # Get query parameters
+            query_params = request.request.GET.dict()
```
Uxio0, line 114: Shouldn't the `query params` be sorted? Same query params in a different order will use different cache keys

safe_transaction_service/history/cache.py
```diff
@@ -0,0 +1,200 @@
...
+            cache_path = f"{urlencode(query_params)}"
```
Uxio0, line 115: `urlencode` does not return a `str` already?

safe_transaction_service/history/cache.py
```diff
@@ -0,0 +1,200 @@
...
+            # Calculate cache_name
+            address = request.kwargs["address"]
```
Uxio0, line 117: Maybe `"address"` should be an optional parameter to the decorator, so we can use different keys in the future

Also this can throw an exception if `address` does not exist. It should be: `request.kwargs.get("address")` so you can emit a warning when you evaluate `if address:` later

safe_transaction_service/history/cache.py
```diff
@@ -0,0 +1,200 @@
...
+                response_data = cache_txs_view.get_cache_data(cache_path)
+                if response_data:
+                    return Response(
+                        status=status.HTTP_200_OK, data=json.loads(response_data)
+                    )
+
+            # Get response from the view
+            response = view_func(request, *args, **kwargs)
+            if response.status_code == 200:
+                # Just store success responses and if cache is enabled with DEFAULT_CACHE_PAGE_TIMEOUT !=0
```
Uxio0, line 138:
```suggestion
                # Just store success responses if cache is enabled with DEFAULT_CACHE_PAGE_TIMEOUT > 0
```

config/settings/base.py
```diff
@@ -683,6 +683,9 @@
 
 # Compression level – an integer from 0 to 9. 0 means not compression
 CACHE_ALL_TXS_COMPRESSION_LEVEL = env.int("CACHE_ALL_TXS_COMPRESSION_LEVEL", default=0)
+CACHE_VIEW_DEFAULT_TIMEOUT = env.int(
+    "DEFAULT_CACHE_PAGE_TIMEOUT", default=60
```
Uxio0, line 687: I would disable the cache by default and start testing it on some networks

Uxio0 (changes requested):

safe_transaction_service/history/cache.py
Uxio0: Either way, making `address` as a url path value configurable is just adding an optional parameter (by default `'address'`) to the decorator, something like `caching_value_name`

safe_transaction_service/history/cache.py
```diff
@@ -96,7 +96,9 @@ def remove_cache(self):
 
 
 def cache_txs_view_for_address(
-    cache_tag: str, timeout: int = settings.CACHE_VIEW_DEFAULT_TIMEOUT
+    cache_tag: str,
+    parameter_key: Optional["str"] = "address",
```
Uxio0, line 100: Parameter is not optional

```suggestion
    parameter_key: str = "address",
```

Uxio0 (changes requested):

Uxio0 (approved):

PR #2326, opened 2024-11-12
Add a table for not valid tokens

### LINKED ISSUE #2283: Balances really slow for one Safe in Base network
For https://safe-transaction-base.safe.global/api/v1/safes/0x32e52896663De88a65c2D94917b006404415A89f/balances/ is really slow (until cache is warm).

After debugging it, we realised is not an RPC or database optimization issue.

The issue is that for a lot of tokens they don't exist in the database so we try to recover them from blockchain, with no luck as they don't hold usual ERC20 information like `name` or `symbol`: https://basescan.org/address/0x20b09dCb0Ca884448acE9DDB10D7851Bfe26C261#readContract 

Method causing the bottleneck: https://github.com/safe-global/safe-transaction-service/blob/main/safe_transaction_service/history/services/balance_service.py#L293

Uxio0:
- Not valid tokens are retrieved again and again after cache expires
- Add a simple list of not valid tokens so RPC call is not required
- Also rename `filter_addresses` -> `filter_tokens`, which I feel is way more clear
- Closes #2283


falvaradorodriguez (approved):

Uxio0:
Thanks @falvaradorodriguez . I would like @moisses89 review here too as he was working in the balances stuff not so long ago 😉

moisses89 (approved):
Looks good to me! 
Just a minimal comment, do you think that makes sense add something like, why is not a token address included in the Safe balances? to the FAQ?

Uxio0:
> Looks good to me! Just a minimal comment, do you think that makes sense add something like, why is not a token address included in the Safe balances? to the FAQ?

Done!

PR #2347, opened 2024-11-27
Fix SafeRelevantTransaction default timestamp

falvaradorodriguez:
Sometimes when reindexing a safe, if a missing Tx appears, it is being inserted with the current date instead of the correct date. 

This causes an order problem in the `all-transactions` endpoint.

### DIFF
--- a/safe_transaction_service/history/indexers/tx_processor.py
+++ b/safe_transaction_service/history/indexers/tx_processor.py
@@ -722,7 +722,7 @@ class SafeTxProcessor(TxProcessor):
                 SafeRelevantTransaction.objects.get_or_create(
                     ethereum_tx=ethereum_tx,
                     safe=contract_address,
-                    defaults={"timestamp": ethereum_tx.block.timestamp},
+                    defaults={"timestamp": internal_tx.timestamp},
                 )
 
                 # Don't modify created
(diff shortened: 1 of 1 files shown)


safe_transaction_service/history/indexers/tx_processor.py
```diff
@@ -603,7 +603,7 @@ def __process_decoded_transaction(
                 SafeRelevantTransaction.objects.get_or_create(
                     ethereum_tx=ethereum_tx,
                     safe=contract_address,
-                    defaults={"timestamp": ethereum_tx.created},
+                    defaults={"timestamp": ethereum_tx.block.timestamp},
```
moisses89, line 606: It is the related field included in the queryset ethereum_tx?
If not this will create a new request to database for each ethereum_tx.

moisses89 (changes requested):

safe_transaction_service/history/indexers/tx_processor.py
Uxio0, line 606:
```suggestion
                    defaults={"timestamp": internal_tx.timestamp},
```

safe_transaction_service/history/indexers/tx_processor.py
```diff
@@ -722,7 +722,7 @@ def __process_decoded_transaction(
...
```
Uxio0, line 725:
```suggestion
                    defaults={"timestamp": internal_tx.timestamp},
```

safe_transaction_service/history/indexers/tx_processor.py
Uxio0: `timestamp` for `internal_tx` should be the same of the block (it's denormalized on the database for optimizing some queries). That way we save a database call to get the `block` for every `ethereum_tx`

Uxio0 (changes requested):

Uxio0 (approved):

moisses89 (approved):

PR #2348, opened 2024-11-27
Only reindex singletons/events if service is synced

Uxio0:


falvaradorodriguez (approved):

safe_transaction_service/history/tasks.py
```diff
@@ -309,12 +309,17 @@ def process_decoded_internal_txs_for_safe_task(
...
     """
     Reindexes last hours for master copies to prevent indexing issues
     """
     with contextlib.suppress(LockError):
         with only_one_running_task(self):
+            indexing_status = IndexServiceProvider().get_indexing_status()
+            if not indexing_status.master_copies_synced:
+                logger.warning(
+                    "Reindexing master copies will not be executed as service is out of sync"
+                )
```
moisses89, line 322: Should be a return here?

safe_transaction_service/history/tasks.py
```diff
@@ -337,12 +342,17 @@ def reindex_mastercopies_last_hours_task(self, hours: float = 2.5) -> Optional[i
...
     """
     Reindexes last hours for erx20 and erc721 to prevent indexing issues
     """
     with contextlib.suppress(LockError):
         with only_one_running_task(self):
+            indexing_status = IndexServiceProvider().get_indexing_status()
+            if not indexing_status.erc20_synced:
+                logger.warning(
+                    "Reindexing erc20/721 events will not be executed as service is out of sync"
+                )
```
moisses89, line 355: Should be a return here? and I also would add a return at the end of the task.

moisses89 (changes requested):

safe_transaction_service/history/tasks.py
Uxio0: Done

moisses89 (approved):

PR #2350, opened 2024-11-29
Add expirity and priority to handle Safes processing

moisses89:
# Description
- Add expirity time for process_decoded_internal_tx_task (if the task cannot be executed for more than one minute shouldn't be run) 
- Add priority for process Safes from process process_decoded_internal_tx_task.


Uxio0:
@moisses89 is this still relevant?

PR #2368, opened 2024-12-19
Remove ens client mainnet base url

falvaradorodriguez:
The URL https://api.thegraph.com/subgraphs/name/ensdomains/ens is no longer supported and it is necessary to use the ENS Mainnet subgraph (https://thegraph.com/explorer/subgraphs/5XqPmWe6gjyrJtFn9cLy237i4cWw2j9HcUJEXsP5qGtH) and the subgraph config to make the same queries to the TheGraph api.

- Related with https://github.com/safe-global/safe-eth-py/issues/1481

### DIFF
--- a/safe_transaction_service/history/services/collectibles_service.py
+++ b/safe_transaction_service/history/services/collectibles_service.py
@@ -172,11 +172,12 @@ class CollectiblesService:
                 api_key=api_key,
                 subgraph_id=subgraph_id,
             )
-        # Else, provide fallback for Sepolia, Holesky and Mainnet
+        # Else, provide fallback for Sepolia, Holesky or empty configuration.
         else:
             logger.warning(
                 "Using fallback EnsClient configuration. This configuration is not suitable for production and it is "
-                "recommended to setup a Subgraph API key. See https://docs.ens.domains/web/subgraph"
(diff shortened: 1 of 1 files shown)


safe_transaction_service/history/services/collectibles_service.py
```diff
@@ -197,9 +198,7 @@ def fallback_ens_client(self) -> EnsClient.Config:
                 "https://api.studio.thegraph.com/query/49574/ensholesky/version/latest",
             )
         else:
-            return EnsClient.Config(
-                "https://api.thegraph.com/subgraphs/name/ensdomains/ens/",
-            )
+            return EnsClient.Config("")
```
moisses89, line 201: Should we add a log indicating that fallback will not be configured because there is not any fallback for the network?

falvaradorodriguez: IMO it would already be well indicated with this [log](https://github.com/safe-global/safe-transaction-service/blob/cab037e76c299036e733c9121869880cb80c2b84/safe_transaction_service/history/services/collectibles_service.py#L177). WDYT?

moisses89: On that log we are saying that we will use a fallback but in in this case is not true, because ENS is not configured.

moisses89 (approved):

PR #2418, opened 2025-02-23
Add command to validate integrity for multisig transactions in queue

Uxio0:
- Validate integrity for `MultisigTransaction` and `MultisigConfirmation` in queue

### DIFF
--- a/safe_transaction_service/history/management/commands/validate_tx_integrity.py
+++ b/safe_transaction_service/history/management/commands/validate_tx_integrity.py
@@ -1,0 +1,107 @@
+from django.core.management.base import BaseCommand
+from django.db import connection
+
+from hexbytes import HexBytes
+from safe_eth.eth import get_auto_ethereum_client
+from safe_eth.safe import Safe
+from safe_eth.safe.safe_signature import SafeSignature
+
+from ...models import MultisigConfirmation, MultisigTransaction
+
+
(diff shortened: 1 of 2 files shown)


safe_transaction_service/history/management/commands/validate_tx_integrity.py
```diff
@@ -0,0 +1,101 @@
...
+
+        for multisig_transaction in self.get_multisig_transactions(safe_tx_hashes):
+            if not self.validate_safe_tx_hash(multisig_transaction):
+                self.stdout.write(
+                    self.style.WARNING(
+                        f"{multisig_transaction.safe_tx_hash} is not matching"
+                    )
+                )
+
+            for multisig_confirmation in multisig_transaction.confirmations.all():
```
moisses89, line 92: Could we do the same for multisigtransaction?

Uxio0: I think the relevant part are the confirmations

moisses89 (approved):

safe_transaction_service/history/management/commands/validate_tx_integrity.py
Uxio0: A not executed transaction shouldn't have signatures, but I will check that too

PR #2438, opened 2025-02-27
Ignore response validation of non executed transactions with old nonces

moisses89:
# Description
Don't validate response transactions which nonce is less than current nonce, so cannot be executed. 
Validates the owner of approveHash

### DIFF
--- a/safe_transaction_service/history/exceptions.py
+++ b/safe_transaction_service/history/exceptions.py
@@ -8,26 +8,48 @@ from rest_framework.views import exception_handler
 logger = logging.getLogger(__name__)
 
 
+def _get_exception_format(exception):
+    if str(exception):
+        exception_str = "{}: {}".format(exception.__class__.__name__, exception)
+    else:
+        exception_str = exception.__class__.__name__
+    return exception_str
+
+
+def _generate_response_data(exception_format, message):
+    return {
+        "exception": message,
+        "trace": exception_format,
+    }
+
+
+def _log_message(context, exception_format, exception):
+    logger.warning(
+        "%s - Exception: %s - Data received %s",
+        context["request"].build_absolute_uri(),
+        exception_format,
+        context["request"].data,
+        exc_info=exception,
+    )
+
+
 def custom_exception_handler(exc, context):
     if isinstance(exc, NodeConnectionException):
         response = Response(status=status.HTTP_503_SERVICE_UNAVAILABLE)
-
-        if str(exc):
-            exception_str = "{}: {}".format(exc.__class__.__name__, exc)
-        else:
-            exception_str = exc.__class__.__name__
-        response.data = {
-            "exception": "Problem connecting to Ethereum network",
-            "trace": exception_str,
-        }
-
-        logger.warning(
-            "%s - Exception: %s - Data received %s",
-            context["request"].build_absolute_uri(),
-            exception_str,
-            context["request"].data,
-            exc_info=exc,
+        exception_str = _get_exception_format(exc)
+        response.data = _generate_response_data(
+            exception_str, "Problem connecting to Ethereum network"
         )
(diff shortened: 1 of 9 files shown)


safe_transaction_service/history/exceptions.py
```diff
@@ -8,26 +8,48 @@
 logger = logging.getLogger(__name__)
 
 
+def _get_exception_format(exception):
```
Uxio0, line 11: If possible, could you add typing for the functions?

Uxio0: I would call this function `_get_exception_trace`

safe_transaction_service/history/exceptions.py
```diff
@@ -8,26 +8,48 @@
...
+        exception_format,
+        context["request"].data,
+        exc_info=exception,
+    )
+
+
 def custom_exception_handler(exc, context):
     if isinstance(exc, NodeConnectionException):
         response = Response(status=status.HTTP_503_SERVICE_UNAVAILABLE)
+        exception_str = _get_exception_format(exc)
```
Uxio0, line 39:
```suggestion
        exception_trace = _get_exception_format(exc)
```

safe_transaction_service/history/serializers.py
```diff
@@ -717,15 +717,17 @@ def get_confirmations(self, obj: MultisigTransaction) -> Dict[str, Any]:
         :param obj: MultisigConfirmation instance
         :return: Serialized queryset
         """
-        if obj.ethereum_tx_id:
+        safe_address = obj.safe
+        # Just get safe info for non executed transactions
+        if obj.ethereum_tx_id or obj.nonce < self.context.get("nonce", 0):
```
Uxio0, line 722:
```suggestion
        if obj.ethereum_tx_id or obj.nonce < self.context.get("current_nonce", 0):
```

safe_transaction_service/history/serializers.py
```diff
@@ -717,15 +717,17 @@ def get_confirmations(self, obj: MultisigTransaction) -> Dict[str, Any]:
...
             return SafeMultisigConfirmationResponseSerializer(
                 obj.confirmations, many=True
             ).data
 
         signature_owners_addresses = []
-        safe_address = obj.safe
+
         safe_tx_hash = obj.safe_tx_hash
-        safe_owners = get_safe_owners(safe_address)
+        safe_owners = self.context.get("owners", [])
```
Uxio0, line 730:
```suggestion
        safe_owners = self.context.get("current_owners", [])
```

scripts/check_signatures_file.py
Uxio0, line 1: Should this file be added?

safe_transaction_service/history/views.py
```diff
@@ -586,6 +586,18 @@ class SafeMultisigTransactionDetailView(RetrieveAPIView):
     lookup_field = "safe_tx_hash"
     lookup_url_kwarg = "safe_tx_hash"
 
+    def get_serializer_context(self):
```
Uxio0, line 589: If possible add some docs about why we are adding this

safe_transaction_service/history/views.py
```diff
@@ -649,6 +661,17 @@ class SafeMultisigTransactionListView(ListAPIView):
     ordering_fields = ["nonce", "created", "modified"]
     pagination_class = pagination.DefaultPagination
 
+    def get_serializer_context(self):
```
Uxio0, line 664: If possible add some docs about why we are adding this

Uxio0 (changes requested):

scripts/check_signatures_file.py
moisses89: Nop

Uxio0 (approved):

PR #2451, opened 2025-03-07
Fix ethereum indexer RPC calls exceptions

falvaradorodriguez:
Added Web3RPCError exception to avoid errors like `web3.exceptions.Web3RPCError: {'code': -32602, 'message': 'Query returned more than 10000 results. Try with this block range [x, y].'}`


Uxio0:
Good catch!

PR #2468, opened 2025-03-19
Update not valid transaction

moisses89:
### Description
Update to None the signatures of executed transactions after a reorg. 
This is a continuation of this PR https://github.com/safe-global/safe-transaction-service/pull/2461 with a different approach to avoid remove the non indexed data of a proposal.


safe_transaction_service/history/services/reorg_service.py
```diff
@@ -171,6 +177,16 @@ def recover_from_reorg(self, reorg_block_number: int) -> int:
...
         ).delete()
+
+        # Fix multisig transactions that had EthereumBlock removed
+        # We will remove signature to don't remove not indexed data as origin.
+        # Index from reorg will fill signatures field again
+        # `failed` must be `NULL` in the database if `ethereum_tx` is empty
+        # `ethereum_tx` and `failed` both have a database index
+        MultisigTransaction.objects.filter(ethereum_tx=None).exclude(
+            failed=None
+        ).update(signatures=None)
```
Uxio0, line 188:
```suggestion
        MultisigTransaction.objects.filter(ethereum_tx=None).exclude(
            failed=None
        ).update(signatures=None, failed=None)
```

Uxio0 (changes requested):

safe_transaction_service/history/services/reorg_service.py
moisses89: Good catch!

Uxio0 (approved):

PR #2469, opened 2025-03-20
Setup default singleton addresses if are deployed.

### LINKED ISSUE #2414: Setup safe-singleton addresses from canonical and eip155 addresses when safe-eth-py is not ready.
# What is needed?
Setup safe singleton addresses on transaction service even when safe-eth-py is not adding this addresses. 
This would reduce the dependencies and blocks to release new services on new chains.

moisses89:
Closes #2414

### DIFF
--- a/safe_transaction_service/history/management/commands/setup_service.py
+++ b/safe_transaction_service/history/management/commands/setup_service.py
@@ -226,10 +226,16 @@ class Command(BaseCommand):
                 self.style.SUCCESS(f"Setting up {ethereum_network.name} safe addresses")
             )
             self._setup_safe_singleton_addresses(MASTER_COPIES[ethereum_network])
-            self._setup_erc20_indexing()
+
         else:
+            self.stdout.write(
+                self.style.WARNING(
+                    f"Setting up {ethereum_network.name} default safe addresses from chain with unknown init block"
+                )
+            )
             self._setup_safe_singleton_addresses_from_chain(ethereum_client)
-            self._setup_erc20_indexing()
+
+        self._setup_erc20_indexing()
 
         if ethereum_network in PROXY_FACTORIES:
             self.stdout.write(
@@ -239,6 +245,11 @@ class Command(BaseCommand):
             )
             self._setup_safe_proxy_factories(PROXY_FACTORIES[ethereum_network])
         else:
+            self.stdout.write(
+                self.style.WARNING(
+                    f"Setting up {ethereum_network.name} default proxy factory addresses from chain with unknown init block"
+                )
+            )
             self._setup_safe_proxy_factories_from_chain(ethereum_client)
 
         if not (SafeMasterCopy.objects.count() and ProxyFactory.objects.count()):
@@ -265,7 +276,7 @@ class Command(BaseCommand):
                 safe_singleton_address.version != version
                 or safe_singleton_address.initial_block_number != initial_block_number
             ):
-                safe_singleton_address.version = initial_block_number
+                safe_singleton_address.initial_block_number = initial_block_number
                 safe_singleton_address.version = version
                 safe_singleton_address.save(
                     update_fields=["initial_block_number", "version"]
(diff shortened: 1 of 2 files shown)


safe_transaction_service/history/management/commands/setup_service.py
```diff
@@ -6,11 +6,36 @@
...
 
 from django_celery_beat.models import CrontabSchedule, IntervalSchedule, PeriodicTask
-from safe_eth.eth import get_auto_ethereum_client
+from safe_eth.eth import EthereumClient, get_auto_ethereum_client
+from safe_eth.eth.utils import fast_to_checksum_address
 from safe_eth.safe.addresses import MASTER_COPIES, PROXY_FACTORIES
 
 from ...models import IndexingStatus, IndexingStatusType, ProxyFactory, SafeMasterCopy
 
+default_safe_singletons = [
```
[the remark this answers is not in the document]
Uxio0: Yes. In `safe-deployments` there are `default` addresses (canonical, etc.), so we should store them and use that information. So they should be obtained from `safe-eth-py`

safe_transaction_service/history/management/commands/setup_service.py
```diff
@@ -200,28 +225,34 @@ def handle(self, *args, **options):
             self.stdout.write(
                 self.style.SUCCESS(f"Setting up {ethereum_network.name} safe addresses")
             )
-            self._setup_safe_master_copies(MASTER_COPIES[ethereum_network])
+            self._setup_safe_singleton_addresses(MASTER_COPIES[ethereum_network])
+            self._setup_erc20_indexing()
+        else:
+            self._setup_safe_singleton_addresses_from_chain(ethereum_client)
             self._setup_erc20_indexing()
+
```
Uxio0, line 233: `self._setup_erc20_indexing()` is called in both `if` branches, better call it once after `else` ends

safe_transaction_service/history/management/commands/setup_service.py
```diff
@@ -200,28 +225,34 @@ def handle(self, *args, **options):
             self.stdout.write(
                 self.style.SUCCESS(f"Setting up {ethereum_network.name} safe addresses")
             )
-            self._setup_safe_master_copies(MASTER_COPIES[ethereum_network])
+            self._setup_safe_singleton_addresses(MASTER_COPIES[ethereum_network])
+            self._setup_erc20_indexing()
+        else:
```
Uxio0, line 230: `self.stdout.write` should be called warning people that indexing will be started from scratch (block 0) as we are autodetecting the deployments

safe_transaction_service/history/management/commands/setup_service.py
```diff
@@ -231,12 +262,14 @@ def _setup_safe_master_copies(
...
             if (
-                safe_master_copy.version != version
-                or safe_master_copy.initial_block_number != initial_block_number
+                safe_singleton_address.version != version
+                or safe_singleton_address.initial_block_number != initial_block_number
             ):
-                safe_master_copy.version = initial_block_number
-                safe_master_copy.version = version
-                safe_master_copy.save(update_fields=["initial_block_number", "version"])
+                safe_singleton_address.version = initial_block_number
```
Uxio0, line 268: I guess tests were not covering this case 😉 

```suggestion
                safe_singleton_address.initial_block_number = initial_block_number
```

Uxio0 (changes requested):
Good job. We should do the same with Multicall, etc. contracts too

safe_transaction_service/history/management/commands/setup_service.py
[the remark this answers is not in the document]
moisses89: Done

safe_transaction_service/history/management/commands/setup_service.py
moisses89: Added a test to ensure that is being updated correctly.

Uxio0 (approved):

PR #2539, opened 2025-06-06
Fix 1271 contract signatures

Uxio0:
- Closes #2535 and #2526

### DIFF
--- a/safe_transaction_service/safe_messages/serializers.py
+++ b/safe_transaction_service/safe_messages/serializers.py
@@ -16,7 +16,10 @@ from safe_eth.util.util import to_0x_hex_str
 from safe_transaction_service.utils.serializers import get_safe_owners
 
 from .models import SIGNATURE_LENGTH, SafeMessage, SafeMessageConfirmation
-from .utils import get_hash_for_message, get_safe_message_hash_for_message
+from .utils import (
+    get_message_encoded,
+    get_safe_message_hash_for_message,
+)
 
 
 # Request serializers
@@ -105,12 +108,12 @@ class SafeMessageSerializer(SafeMessageSignatureParserMixin, serializers.Seriali
     def validate(self, attrs):
         attrs = super().validate(attrs)
         safe_address = self.context["safe_address"]
+        attrs["safe"] = safe_address
         message = attrs["message"]
         signature = attrs["signature"]
-        attrs["safe"] = safe_address
-        message_hash = get_hash_for_message(message)
+        message_encoded = get_message_encoded(message)
         safe_message_hash = get_safe_message_hash_for_message(
-            safe_address, message_hash
+            safe_address, message_encoded
         )
         attrs["message_hash"] = safe_message_hash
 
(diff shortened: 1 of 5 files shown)


safe_transaction_service/safe_messages/utils.py
```diff
@@ -1,22 +1,36 @@
...
+from eth_account.messages import encode_defunct
 from eth_typing import ChecksumAddress, Hash32
 from safe_eth.eth import get_auto_ethereum_client
-from safe_eth.eth.eip712 import eip712_encode_hash
+from safe_eth.eth.eip712 import eip712_encode
 from safe_eth.safe import Safe
 
 
-def get_hash_for_message(message: str | dict[str, Any]) -> Hash32:
+def encode_eip191_message(message: str) -> bytes:
```
mmv08, line 10: why is this needed?

safe_transaction_service/safe_messages/utils.py
```diff
@@ -1,22 +1,36 @@
...
+        + signable_message.header
+        + signable_message.body
+    )
+
+
+def encode_eip712_message(message: dict[str, Any]) -> bytes:
+    return b"".join(eip712_encode(message))
+
+
+def get_message_encoded(message: str | dict[str, Any]) -> bytes:
```
mmv08, line 24: why does a message require a certain encoding? this is the function for the raw message -> bytes message encoding, right? (not the SafeMessage EIP712 struct)

Uxio0: This is only to get an dictionary or a string and encode it (using either EIP712 or EIP191), but it's not the SafeMessage struct

safe_transaction_service/safe_messages/utils.py
Uxio0: I need to encode the provided message somehow, but I don't want the hash, so I had to write this

safe_transaction_service/safe_messages/serializers.py
```diff
@@ -158,11 +167,11 @@ def validate(self, attrs):
         attrs["safe_message"] = safe_message
         signature: HexStr = attrs["signature"]
         safe_address = safe_message.safe
-        message_hash = get_hash_for_message(safe_message.message)
+        message_encoded = get_message_encoded(safe_message.message)
         safe_message_hash = safe_message.message_hash
 
         safe_signatures = SafeSignature.parse_signature(
-            signature, safe_message_hash, safe_hash_preimage=message_hash
+            signature, safe_message_hash, safe_hash_preimage=message_encoded
```
nlordell: One thing that surprises me, is that you use `get_message_encoded` as the pre-image here, while above you use `get_safe_message_hash_and_preimage_for_message` which is quite surprising - I would expect these both to use the same pre-image value.

nlordell: Ah yes, I would expect this to be:

```python
message_encoded = get_message_encoded(safe_message.message)
_, safe_message_preimage = (
    get_safe_message_hash_and_preimage_for_message(
        safe_address, message_encoded
    )
)
safe_message_hash = safe_message.message_hash

safe_signatures = SafeSignature.parse_signature(
    signature, safe_message_hash, safe_hash_preimage=safe_message_preimage
)
```

Uxio0: Ops, I'm fixing it. In that case signature is only used to get the `owner` and the `signature_type`, so it's working

nlordell: LGTM!

Uxio0: It was fixed 😄

Issue #2472, opened 2025-03-24
Child Safes cannot register delegates


iamacook:
**Describe the bug**

If a Safe is primarily owned by a Safe with >1 owners, it cannot register delegates.

Delegate registration [parses the incoming signature](https://github.com/safe-global/safe-transaction-service/blob/7b459013259f986c67bedc22cdff487e7b80833d/safe_transaction_service/history/serializers.py#L466), [expecting there to be only one](https://github.com/safe-global/safe-transaction-service/blob/7b459013259f986c67bedc22cdff487e7b80833d/safe_transaction_service/history/serializers.py#L470). If the parental Safe has >1 owners, the proposed signature will parse into >1 signatures meaning it will fail.

**To Reproduce**

1. Create a (parent) Safe with >1 owners/threshold.
2. Create a (child) Safe owned by parent.
3. Create an message in parent for delegate registration in child.
4. Observe the Transaction Service preventing registration with said signature.

**Expected behavior**

Delegate is registered.

PR #2555, opened 2025-06-26
Fix 1271 signatures for delegates

Uxio0:
- Related with #2472 

Fixed the validation of EIP-1271 signatures so a nested safe can add a delegate user for its Safe owner using a 1271 signature.

Added a test case where the scenario described in the open issue is validated.


Uxio0 (commented):
LGTM

falvaradorodriguez (approved):

PR #2563, opened 2025-07-02
Optimize tx processing

Uxio0:
Process every Safe together:
- Less database sessions/transactions
- Less database queries to get internal txs for every Safe

### DIFF
--- a/safe_transaction_service/history/indexers/tx_processor.py
+++ b/safe_transaction_service/history/indexers/tx_processor.py
@@ -4,7 +4,7 @@ Contains classes for processing indexed data and store Safe related models in da
 
 import logging
 from abc import ABC, abstractmethod
-from typing import Iterator, Optional, Sequence, Union
+from typing import Optional, Sequence, Union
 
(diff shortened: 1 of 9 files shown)


falvaradorodriguez (approved):

safe_transaction_service/history/indexers/tx_processor.py
```diff
@@ -376,23 +376,43 @@ def process_decoded_transaction(
...
+        }
+        banned_addresses = set(
+            SafeContract.objects.get_banned_addresses(addresses=contract_addresses)
+        )
 
         try:
-            for internal_tx_decoded in internal_txs_decoded:
+            for internal_tx_decoded in internal_txs_decoded_list:
                 contract_address = internal_tx_decoded.internal_tx._from
                 contract_addresses.add(contract_address)
```
falvaradorodriguez: This add is redundant now?

Uxio0: You are right

PR #2565, opened 2025-07-04
Use hashes instead of full message for 1271

Uxio0:
- Revert to the old behaviour, so we don't break the frontend
- For 1271 `isValid` accepts hashes instead of full data

### DIFF
--- a/safe_transaction_service/safe_messages/serializers.py
+++ b/safe_transaction_service/safe_messages/serializers.py
@@ -114,11 +114,9 @@ class SafeMessageSerializer(SafeMessageSignatureParserMixin, serializers.Seriali
         signature = attrs["signature"]
         # Encode EIP-191 or EIP-712 original message as bytes
         # Use fast_keccak to maintain compatibility with the old version
-        message_encoded = fast_keccak(get_message_encoded(message))
+        message_hash = fast_keccak(get_message_encoded(message))
         safe_message_hash, safe_message_preimage = (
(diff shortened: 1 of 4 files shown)


safe_transaction_service/safe_messages/serializers.py
```diff
@@ -112,7 +113,8 @@ def validate(self, attrs):
         message = attrs["message"]
         signature = attrs["signature"]
         # Encode EIP-191 or EIP-712 original message as bytes
-        message_encoded = get_message_encoded(message)
+        # Use fast_keccak to maintain compatibility with the old version
+        message_encoded = fast_keccak(get_message_encoded(message))
```
moisses89:
```suggestion
        message_hash = fast_keccak(get_message_encoded(message))
```

safe_transaction_service/safe_messages/serializers.py
```diff
@@ -167,7 +169,7 @@ def validate(self, attrs):
         attrs["safe_message"] = safe_message
         signature: HexStr = attrs["signature"]
         safe_address = safe_message.safe
-        message_encoded = get_message_encoded(safe_message.message)
+        message_encoded = fast_keccak(get_message_encoded(safe_message.message))
```
moisses89:
```suggestion
        message_hash = fast_keccak(get_message_encoded(safe_message.message))
```

moisses89 (changes requested):

PR #2591, opened 2025-07-30
Only update modified fields for SafeLastStatus

Uxio0:
- `SafeLastStatus` rows are modified all the time. It's very important to only update modified fields, and not every one. That's very important for Safes with a lot of owners/modules


safe_transaction_service/history/indexers/tx_processor.py
```diff
@@ -349,17 +349,24 @@ def disable_module(
...
         Updates `SafeLastStatus`. An entry to `SafeStatus` is added too via a Django signal.
 
         :param safe_last_status:
         :param internal_tx:
+        :param modified_fields: only update modified fields (db optimization)
         :return: Updated `SafeLastStatus`
         """
         safe_last_status.internal_tx = internal_tx
-        safe_last_status.save()
+        if modified_fields:
```
moisses89, line 366:
```suggestion
        modified_fields.append("internal_tx") 
```
Why not include always the internal_tx in the modified_fields? the you could avoid this if and do always save(update_fields=modified_fields)

Uxio0: Because sometimes you want to insert the object (like in `setup`) and not update it

PR #2596, opened 2025-07-31
Export CSV main query fix

moisses89:
### ❗ Issue:

The previous approach split the logic into three separate blocks—one for each type of transfer (`erc20`, `erc721`, and `internaltx`).
This meant we had to write **9 subqueries** to retrieve all possible transfers with the required information.
The structure was unnecessarily complex, because each subquery had to check that the transaction **wasn't already included** in another table.
For example, the standalone queries had to exclude any transaction that appeared in either `multisigtransaction` or `moduletransaction`.

This logic led to **missing transactions**, as shown in the case below:

---

### ❌ Missing Incoming Transaction Example:

1. **Multisig Transaction Handling**:
    
    The original query only included ERC20 transfers where the `safe` field in the `multisigtransaction` matched the specified Safe address.
    However, it missed cases where the Safe was **the recipient** of a multisig transaction (i.e. `erc20.to = safe`), but not the one that initiated it.
    
2. **Standalone ERC20 Transfers**:
    
    These queries were designed to fetch ERC20 transfers that **were not part of a multisig transaction**, by excluding transfers with matching `ethereum_tx_id`s in the `multisigtransaction` table.
    
3. **Combination of Both**:
    
    The combination of this logic excluded **valid incoming transfers**, where:
    
    - The Safe address was the **recipient** (`erc20.to`),
    - But the **sender** was another Safe involved in a multisig transaction,
    - And the `ethereum_tx_id` existed in both the `erc20transfer` and `multisigtransaction` tables.
    
    As a result, these transactions were ignored, even though they should have been included.
    

---

### ⚙️ Performance Limitation:

On top of that, the **date filter** was applied **after** combining all subqueries with `UNION ALL`.

This meant the database had to sort and filter from a large, merged dataset, significantly increasing processing time and resource usage.

# Solution:

- Retrieve all types of transactions in a single SQL query by combining them into just three subqueries—one for each transfer type.
- Use `LEFT JOIN`s to include possible associated transactions and apply a priority-based `DISTINCT` logic to determine the correct transaction in case of duplicates.
  A transfer can belong to a `multisigtransaction`, `moduletransaction`, or be a `standalone` transaction.
  The `LEFT JOIN` allows for nulls in the joins—if the `module` join is not null, it's a module transaction; if the `multisig` is not null, it's a multisig transaction; otherwise, it's considered standalone.

- Apply date filtering within each subquery. This change was tested and reduced the execution cost from **255054.98** to **79615.97**—a **3x performance improvement**.
- As total performant improvement we pass from previous query cost 2.612.529 to cost 80.036 with the new query — a 32x performance improvent in total.


safe_transaction_service/history/services/transaction_service.py
```diff
@@ -403,337 +403,176 @@ def get_export_transactions(
...
-            AND NOT EXISTS (
-                SELECT 1 FROM history_moduletransaction modtx
-                JOIN history_internaltx itx ON modtx.internal_tx_id = itx.id
-                WHERE itx.ethereum_tx_id = erc721.ethereum_tx_id
-            )
-
-            UNION ALL
+                encode(COALESCE(mt.to, modtx.to), 'hex') as contract_address,
+
+                ROW_NUMBER() OVER (
```
falvaradorodriguez, line 545: Please, could you add a comment here explaining this logic?

safe_transaction_service/history/services/transaction_service.py
```diff
@@ -403,337 +403,176 @@ def get_export_transactions(
...
-                COALESCE(mt.origin->>'note', '') as note,
-                encode(mt.ethereum_tx_id, 'hex') as transaction_hash,
-                encode(mt.safe_tx_hash, 'hex') as safe_tx_hash,
-                null as method,
-                encode(mt.to, 'hex') as contract_address,
-                COALESCE(erc721.timestamp, mt.created) as sort_date
-            FROM history_multisigtransaction mt
-            JOIN history_ethereumtx et ON mt.ethereum_tx_id = et.tx_hash
-            JOIN history_erc721transfer erc721 ON et.tx_hash = erc721.ethereum_tx_id
+              encode(%s, 'hex') as safe_address,
```
falvaradorodriguez: IMO It's more readable without empty lines inside the SELECT. WDYT?

falvaradorodriguez (approved):

safe_transaction_service/history/services/transaction_service.py
```diff
@@ -403,337 +403,163 @@ def get_export_transactions(
...
+            -- ERC20 Transfers
             SELECT
-                encode(mt.safe, 'hex') as safe_address,
-                encode(erc20._from, 'hex') as from_address,
-                encode(erc20.to, 'hex') as to_address,
-                erc20.value::text as amount,
+                encode(%s, 'hex') as safe_address,
+                encode(COALESCE(erc20._from, modtx.module), 'hex') as from_address,
+                encode(COALESCE(erc20.to, modtx.to), 'hex') as to_address,
+                COALESCE(erc20.value::text, modtx.value::text) as amount,
```
moisses89: This is wrong should be always erc20 value because is the transfer.

PR #2611, opened 2025-08-11
Deprecate EtherscanV1

### LINKED ISSUE #2602: feat: remove Etherscan V1 integration.
# What is needed?
Etherscan is stopping their V1 API August 15th. Stop Etherscan tasks for fetching Decoded Contracts.

# Background
Decoding functionality will be removed from the Transaction Service and start a migration process to the Decoding Service as the way to go from now on. Decoding Service will still be able to fetch decoded contracts from Etherscan new API. We need to stop the fetching tasks and related Etherscan connections. We will keep other providers for now to make the migration as smooth as possible.

# Related issues
Paste here the related links for the issues on the clients/safe project if applicable. Please provide at least one of the following:
- Links to epics in your repository
- Images taken from mocks
- Gitbook or any form of written documentation links, etc. Any of these alternatives will help us contextualise your request.

# Endpoint
If applicable, description on the endpoint and the result you expect:

## URL
`GET /api/v1/safes/<address>/creation/`

## Response
```
{
  created: "<iso 8601 datetime>",
  transactionHash: "<keccak-hash>",
  creator: "<checksummed-address>"
}
```

moisses89:
Closes #2602

### DIFF
--- a/safe_transaction_service/contracts/tasks.py
+++ b/safe_transaction_service/contracts/tasks.py
@@ -8,7 +8,6 @@ from django.utils import timezone
 from celery import app
 from celery.utils.log import get_task_logger
 from eth_typing import ChecksumAddress
-from safe_eth.eth.clients import EtherscanRateLimitError
 from safe_eth.safe.multi_send import MultiSend
 
 from safe_transaction_service.history.models import (
@@ -119,7 +118,6 @@ def reindex_contracts_without_metadata_task() -> int:
 
 
 @app.shared_task(
-    autoretry_for=(EtherscanRateLimitError,),
     retry_backoff=10,
     retry_kwargs={"max_retries": 5},
 )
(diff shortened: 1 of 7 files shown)


safe_transaction_service/contracts/tasks.py
```diff
@@ -119,7 +118,6 @@ def reindex_contracts_without_metadata_task() -> int:
 
 
 @app.shared_task(
-    autoretry_for=(EtherscanRateLimitError,),
```
dasanra, line 122: This was only to autoretry when Etherscan was returning a RateLimitError, now would autoretry for any error, right?

moisses89: yes, it was to autoretry in case EtherscanRateLimitError, maybe a better approach would remove the retries.

falvaradorodriguez (approved):

PR #2616, opened 2025-08-18
Paginate Safes pending to be processed

moisses89:
# Description
The query pending Safes to be processed can be slow in case of a big amount of Safes pending. 
The idea of this PR is paginate the query without a distinct to avoid a big sort on disk. 
The distinct is being appliead in redis, with the following features:
- use O(1) operations, to ensure have not big blockers. 
- expiry on task timeou, to avoid acumulate keys on redis
- remove with `unlink` if the task is succesfully ending, at this way redis will do the delete on bakcgroung avoiding a big block.


falvaradorodriguez (approved):

safe_transaction_service/history/models.py
```diff
@@ -1250,6 +1250,14 @@ def safes_pending_to_be_processed(self) -> QuerySet[ChecksumAddress]:
             .distinct("internal_tx___from")
         )
 
+    def safes_pending_to_be_processed_without_distinct(
+        self,
+    ) -> QuerySet[ChecksumAddress]:
+        """
+        :return: List of Safe addresses with repeated address that have transactions pending to be processed
+        """
+        return self.not_processed().values_list("internal_tx___from", flat=True)
```
Uxio0: This is not sorted, so results are not stable. Any query can return Safes in a total different order

Uxio0 (changes requested):

PR #2671, opened 2025-10-07
use ruff

Uxio0:
- **Use Ruff instead of Flake8 + isort + black**
- **Format files and fix issues**
- **Ignore format comment**


safe_transaction_service/history/tests/test_tasks.py
```diff
@@ -149,9 +147,9 @@ def test_reindex_mastercopies_last_hours_task(
         self.assertFalse(reindex_mastercopies_last_hours_task())
         reindex_master_copies_mock.assert_not_called()
 
-        ethereum_block_0 = EthereumBlockFactory(timestamp=one_week_ago)
+        EthereumBlockFactory(timestamp=one_week_ago)
         ethereum_block_1 = EthereumBlockFactory(timestamp=one_day_ago)
-        ethereum_block_2 = EthereumBlockFactory(timestamp=one_hour_ago)
+        EthereumBlockFactory(timestamp=one_hour_ago)
```
moisses89, line 152: Even if the variable is not being used I think that is explanatory of the test.

PR #2773, opened 2026-02-05
fix: fix missing native transfers in all-transactions for multiSend o…

crlosif:
ref #2764 
…n chains without tracing

- [x] If you want to add your network to `setup_service.py`, provide a link to your

### What was wrong? 👾

Closes #

### How was it fixed? 🎯
- Decode multiSend batch data when a multisig tx has no indexed native transfers and add synthetic ETHER_TRANSFER entries from valueDecoded  (operation=CALL, value>0).
- Enrich transfer_dict only when ether_count is 0 to avoid duplicating transfers on chains that do provide tracing (e.g. Sepolia).
- Add _get_native_transfers_from_multisend and _enrich_transfers_from_multisend_decoded in TransactionService.
- Add test_multisend_native_transfers_enriched_when_no_traces.


Genmin:
I took a pass here because #2764 was also on my radar. The overall shape looks right to me: enriching only when indexed native transfers are absent should avoid duplicating traces on chains that already have internal tx indexing.

One hardening I would consider before merge: `_get_native_transfers_from_multisend` currently accepts any Safe transaction with `operation == DELEGATE_CALL` and then tries to decode `multisig_tx.data` as MultiSend data. `decode_multisend_data` will reject most malformed payloads, but a delegate-call to a non-MultiSend target with MultiSend-shaped calldata could still synthesize Safe -> recipient native transfers in the API. Since the canonical MultiSend deployments are known, I think this should also check `multisig_tx.to` against the known MultiSend address(es) for the chain before adding synthetic transfers.

A focused regression for that edge would be: mock `decode_multisend_data` to return one native CALL item, set `operation=DELEGATE_CALL`, but set `to` to a non-MultiSend contract; expected result is no synthetic `ETHER_TRANSFER`. That would keep this fix scoped to actual MultiSend batches while still solving the no-tracing chain case.

crlosif:
I have read the CLA Document and I hereby sign the CLA

moisses89:
Thanks for digging into this! While this does help all-transactions, it doesn't fix the root cause, native transfers simply aren't indexed on chains without tracing (for multisend. 
Since this only synthesizes data inside `TransactionService`, other endpoints like `transfers/` and `incoming-transfers/` will continue to be missing native transfers on these chains.

PR #2859, opened 2026-04-13
Tune DB pool, worker connections and add lock/idle timeouts

Uxio0:
## Summary

- Fix gunicorn env vars cast to `int` — `os.environ.get` returns `str` when the env var is set, causing type inconsistency when values are used elsewhere
- Drop `WEB_WORKER_CONNECTIONS` default from 1000 → 100 to match the DB pool ceiling; the extra greenlets were queuing on pool exhaustion rather than adding real concurrency
- Add `DB_LOCK_TIMEOUT` (5 s default) — prevents lock queue cascades when a slow transaction holds a row lock
- Add `DB_IDLE_IN_TRANSACTION_SESSION_TIMEOUT` (30 s default) — releases locks from connections that go idle mid-transaction (crash, network drop) before the pool recycles them via `max_lifetime`

### DIFF
--- a/config/gunicorn.py
+++ b/config/gunicorn.py
@@ -6,5 +6,5 @@ Store gunicorn variables in this file, so they can be read by Django
 import os
 
 gunicorn_request_timeout = int(os.environ.get("WEB_WORKER_TIMEOUT", 60))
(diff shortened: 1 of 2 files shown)


falvaradorodriguez (approved):
LGTM! We should take into account the maximum number of connections configured for each database.

PR #2874, opened 2026-04-30
Add backfill_delete_failed_ethereum_txs management command

Uxio0:
## Summary

- Adds a `backfill_delete_failed_ethereum_txs` management command that identifies `EthereumTx` rows stored with `status=1` whose on-chain receipt actually reports `status=0` (failure), and deletes them to resolve the DB/chain state mismatch believed to be causing reindex incidents (PLA-1011)
- Supports `--dry-run` to audit impact before committing, `--days` to limit the scan window, and `--retries`/`--backoff` for RPC resilience with exponential backoff
- Adds tests covering deletion, dry-run, missing receipts, retry-then-succeed, exhausted retries, and the days filter

Closes PLA-1404
Related to PLA-1011

### DIFF
--- a/safe_transaction_service/history/management/commands/backfill_delete_failed_ethereum_txs.py
+++ b/safe_transaction_service/history/management/commands/backfill_delete_failed_ethereum_txs.py
@@ -119,7 +119,7 @@ class Command(BaseCommand):
                 continue
 
             if tx_receipt.get("status") == 0:
-                self.stdout.write(f"Deleting {ethereum_tx.tx_hash.hex()}")
+                self.stdout.write(f"Deleting {ethereum_tx.tx_hash}")
                 if not dry_run:
                     ethereum_tx.delete()
                 deleted += 1
--- a/safe_transaction_service/history/management/commands/backfill_delete_failed_ethereum_txs.py
+++ b/safe_transaction_service/history/management/commands/backfill_delete_failed_ethereum_txs.py
@@ -1,0 +1,132 @@
+# SPDX-License-Identifier: FSL-1.1-MIT
+import time
+from datetime import timedelta
+
+from django.core.management.base import BaseCommand
+from django.utils import timezone
+
+try:
+    from safe_eth.eth import EthereumClientProvider
+except ImportError:  # pragma: no cover - fallback for environments without provider
+    from safe_eth.eth import get_auto_ethereum_client
+
+    def EthereumClientProvider():
+        return get_auto_ethereum_client()
+
+
+from ...models import EthereumTx
+
+
+class Command(BaseCommand):
+    help = (
+        "Delete recent EthereumTx rows marked successful in DB after confirming "
+        "failed receipts on RPC"
+    )
+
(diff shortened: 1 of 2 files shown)


safe_transaction_service/history/management/commands/backfill_delete_failed_ethereum_txs.py
```diff
@@ -0,0 +1,136 @@
...
+        )
+        parser.add_argument(
+            "--dry-run",
+            action="store_true",
+            help="Only report rows that would be deleted",
+        )
+        parser.add_argument(
+            "--retries",
+            type=int,
+            default=3,
```
falvaradorodriguez, line 43: Should this be validated as > 0?

safe_transaction_service/history/management/commands/backfill_delete_failed_ethereum_txs.py
```diff
@@ -0,0 +1,136 @@
...
+                )
+                rpc_errors += 1
+                continue
+            checked += 1
+
+            if not tx_receipt:
+                missing_receipts += 1
+                continue
+
+            if tx_receipt.get("status") == 0:
```
falvaradorodriguez, line 129: Is this always 0, or could it be something like `0x0` or `0`?

safe_transaction_service/history/management/commands/backfill_delete_failed_ethereum_txs.py
```diff
@@ -0,0 +1,136 @@
...
+                time.sleep(delay)
+
+    def handle(self, *args, **options):
+        days = options["days"]
+        dry_run = options["dry_run"]
+        retries = options["retries"]
+        backoff = options["backoff"]
+        ethereum_client = EthereumClientProvider()
+
+        queryset = EthereumTx.objects.filter(status=1).only("tx_hash")
```
falvaradorodriguez, line 82: IMO for `status=1` It's better to use a constant. WDYT?

falvaradorodriguez (dismissed):

safe_transaction_service/history/management/commands/backfill_delete_failed_ethereum_txs.py
Uxio0: We use EthereumClient -> Web3.py , so it should be `0`

safe_transaction_service/history/management/commands/backfill_delete_failed_ethereum_txs.py
Uxio0: As we are using status like this in all the codebase, I wouldn't change it at this point

safe_transaction_service/history/management/commands/backfill_delete_failed_ethereum_txs.py
Uxio0: Agree, done in 1464e8cd

falvaradorodriguez (approved):

PR #2878, opened 2026-05-08
feat: Add gas fee fields to CSV export endpoint

falvaradorodriguez:
- Related with https://linear.app/safe-global/issue/PLA-1419/history-csv-export-should-contain-paid-from-the-safe-info#comment-9b93eec1

## Summary                                                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                          
  - Adds `gas_token`, `fee`, `gas_token_symbol` and `gas_token_decimals` to the                                                                                                                                                                                                                                           
    `/api/v1/safes/{address}/export/` response (PLA-1419)
  - Follows the same semantics as `/api/v1/multisig-transactions/`: `gas_token`                                                                                                                                                                                                                                           
    is always returned when a multisig transaction exists (null for incoming or                                                                                                                                                                                                                                           
    standalone transfers); ~`fee` is `gas_used × eth_gas_price`~ payment, null when either                                                                                                                                                                                                                                          
    is not available                                                                                                                                                                                                                                                                                                      
  - `gas_token_symbol` and `gas_token_decimals` are resolved via the tokens table                                                                                                                                                                                                                                         
    (same pattern as the existing `asset_symbol`/`asset_decimals` fields),                                                                                                                                                                                                                                                
    allowing the gateway to call `formatUnits(fee, gas_token_decimals)` without
    additional lookups.

### DIFF
--- a/safe_transaction_service/history/services/transaction_service.py
+++ b/safe_transaction_service/history/services/transaction_service.py
@@ -456,11 +456,7 @@ class TransactionService:
                 COALESCE(mt.origin->> 'note', '') as note,
                 encode(erc20.ethereum_tx_id, 'hex') as transaction_hash,
                 encode(COALESCE(mt.to, modtx.to), 'hex') as contract_address,
-                -- Only expose gas fee fields when the multisig tx belongs to the queried Safe.
-                -- Edge case: when Safe A sends to Safe B, Safe B's export sees a multisig tx
-                -- from Safe A (with its nonce, gas_token and payment) — those should be null.
-                -- payment is the exact refund paid by the Safe to the executor as emitted in
-                -- the ExecutionSuccess/ExecutionFailure event (differs from gas_used * gas_price).
+                -- Only expose the nonces and gas fee fields when the multisig tx belongs to the queried Safe.
                 CASE WHEN mt.safe = %s THEN mt.nonce ELSE NULL END AS nonce,
                 CASE WHEN mt.safe = %s THEN encode(mt.gas_token, 'hex') ELSE NULL END AS gas_token,
                 CASE WHEN mt.safe = %s THEN mt.payment::text ELSE NULL END AS payment,
@@ -507,11 +503,7 @@ class TransactionService:
                 COALESCE(mt.origin->> 'note', '') as note,
                 encode(erc721.ethereum_tx_id, 'hex') as transaction_hash,
                 encode(COALESCE(modtx.to, mt.to), 'hex') as contract_address,
-                -- Only expose gas fee fields when the multisig tx belongs to the queried Safe.
-                -- Edge case: when Safe A sends to Safe B, Safe B's export sees Safe A's multisig
-                -- tx — nonce, gas_token and payment should be null from Safe B's perspective.
-                -- payment is the exact refund paid by the Safe to the executor as emitted in
-                -- the ExecutionSuccess/ExecutionFailure event (differs from gas_used * gas_price).
+                -- Only expose the nonces and gas fee fields when the multisig tx belongs to the queried Safe.
                 CASE WHEN mt.safe = %s THEN mt.nonce ELSE NULL END AS nonce,
                 CASE WHEN mt.safe = %s THEN encode(mt.gas_token, 'hex') ELSE NULL END AS gas_token,
                 CASE WHEN mt.safe = %s THEN mt.payment::text ELSE NULL END AS payment,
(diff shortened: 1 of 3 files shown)


safe_transaction_service/history/services/transaction_service.py
```diff
@@ -458,6 +458,10 @@ def get_export_transactions(
                 encode(COALESCE(mt.to, modtx.to), 'hex') as contract_address,
                 -- Just get the nonces from the provided Safe
                 CASE WHEN mt.safe = %s THEN mt.nonce ELSE NULL END AS nonce,
+                encode(mt.gas_token, 'hex') AS gas_token,
```
moisses89: Are you sure about this sentence in the PR description `gas_token is always returned when a multisig transaction exists (null for incoming or standalone transfers)`
Imagine a situation where a Safe sends ERC20 to another Safe, you will have a multisigtransaction with the gas token for the sender, so will not be null.
In the case of the nonce we are setting as Null for this edge case
`CASE WHEN mt.safe = %s THEN mt.nonce ELSE NULL END AS nonce, `

safe_transaction_service/history/services/transaction_service.py
```diff
@@ -458,6 +458,10 @@ def get_export_transactions(
...
+                (et.gas_used * et.gas_price)::text AS fee,
```
moisses89: I'm wondering if is right use the EthereumTransaction data for fee calculation, because ethereum transaction fees are always on native token, if we want to represent the fees on gast token I think that we should use the data from multisigtransaction (eg: gas_price).

safe_transaction_service/history/services/transaction_service.py
moisses89: This is in several parts of the code.

safe_transaction_service/history/services/transaction_service.py
moisses89: Also this is happening in several parts of the code.

moisses89 (changes requested):
Requesting changes to block the merge until ensure that we are getting the right data.

safe_transaction_service/history/services/transaction_service.py
falvaradorodriguez: You're right. I will replace fee with payment from MultisigTransaction, which is the exact refund emitted by ExecutionSuccess(bytes32 txHash, uint256 payment), correctly represents the cost in the gas token.

safe_transaction_service/history/services/transaction_service.py
falvaradorodriguez: Added here https://github.com/safe-global/safe-transaction-service/pull/2878/commits/9257ee1790bfc0aea70e1b43ed8ecd647c7c184d

safe_transaction_service/history/services/transaction_service.py
falvaradorodriguez: Added here https://github.com/safe-global/safe-transaction-service/pull/2878/commits/9257ee1790bfc0aea70e1b43ed8ecd647c7c184d

safe_transaction_service/history/services/transaction_service.py
```diff
@@ -456,8 +456,16 @@ def get_export_transactions(
                 COALESCE(mt.origin->> 'note', '') as note,
                 encode(erc20.ethereum_tx_id, 'hex') as transaction_hash,
                 encode(COALESCE(mt.to, modtx.to), 'hex') as contract_address,
-                -- Just get the nonces from the provided Safe
+                -- Only expose gas fee fields when the multisig tx belongs to the queried Safe.
```
moisses89: I think that with this comment is enough, becuase the edge case A to B is included on it.

safe_transaction_service/history/services/transaction_service.py
```diff
@@ -456,8 +456,16 @@ def get_export_transactions(
...
+                -- Edge case: when Safe A sends to Safe B, Safe B's export sees a multisig tx
+                -- from Safe A (with its nonce, gas_token and payment) — those should be null.
+                -- payment is the exact refund paid by the Safe to the executor as emitted in
+                -- the ExecutionSuccess/ExecutionFailure event (differs from gas_used * gas_price).
```
moisses89: `-- payment is the exact refund paid by the Safe to the executor as emitted in`
Documentation related fields should go into the model, I think that this is even documented there. 
`the ExecutionSuccess/ExecutionFailure event (differs from gas_used * gas_price)`
Again I think this is more interesting in the model than in a query.

falvaradorodriguez: To be honest, I don’t think the comments are out of place here, as the query is quite complex. 

In any case, I have no problem removing them. 

Done. https://github.com/safe-global/safe-transaction-service/pull/2878/changes/e6ed47520c27284c8ed53ab06b60d8084864f879

moisses89 (approved):

PR #2882, opened 2026-05-15
feat: Publish events to topic exchange with routing keys (PLA-1466)

Uxio0:
Switch the events producer from a fanout-only setup to a topic exchange so consumers can subscribe to specific event types / chains / safes instead of filtering every message in application code.

Routing key format:
- `"{chainId}.{type}.{address}" `(address lower-cased, "_" placeholder when absent).

The legacy fanout exchange is bound downstream of the topic exchange with routing_key="#" on first publish (idempotent), so existing consumers keep receiving every event during the migration window with no producer-side duplication or loss. Once consumers have migrated their bindings, the legacy exchange can be removed.

### DIFF
--- a/safe_transaction_service/events/services/queue_service.py
+++ b/safe_transaction_service/events/services/queue_service.py
@@ -15,7 +15,17 @@ logger = logging.getLogger(__name__)
 
 class QueueService:
     def __init__(self):
+        # Topic exchange — events are published here so consumers can subscribe
+        # by routing-key pattern. Routing keys are "{chainId}.{type}.{address}".
         self.exchange = Exchange(
+            settings.EVENTS_QUEUE_TOPIC_EXCHANGE_NAME,
+            type="topic",
+            durable=True,
+        )
+        # Legacy fanout exchange. Existing consumers stay bound to it; we bind
+        # it downstream of the topic exchange (routing_key="#") so every event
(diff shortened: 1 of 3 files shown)


safe_transaction_service/events/services/queue_service.py
```diff
@@ -24,23 +34,78 @@ def __init__(self):
...
+        except Exception as exc:
+            logger.error(
+                "Could not bind legacy fanout exchange to topic exchange: %s",
+                exc,
+                exc_info=True,
+            )
+            raise
+
+    @staticmethod
+    def _build_routing_key(payload: dict[str, Any]) -> str:
```
falvaradorodriguez, line 72: IMO, It would be nice to add a few unit tests for `_build_routing_key`, since it’s the core new logic here. Especially around edge cases like missing values  "_" fallbacks and address lowercasing.

falvaradorodriguez (dismissed):
Great contribution!

safe_transaction_service/events/services/queue_service.py
Uxio0: Added in b1a038d04b292eba7613eaab7c5013a17d09e2d5

falvaradorodriguez (approved):
LGTM!

PR #2885, opened 2026-05-19
fix: Auto-adjust block range and tolerate RPC errors during reindex

Uxio0:
## Summary
  - `reindex_master_copies` / `reindex_erc20_events` (commands, Celery tasks, and the "last hours" variants) now reuse the live indexer's self-tuning chunk size, so reindex throughput matches regular indexing on the same node.
  - Transient RPC errors (`OSError`, `ValueError`, `Web3RPCError`) no longer abort a reindex run. The chunk is retried with the minimum window, and auto-adjust grows it back as requests succeed — same recovery pattern the live indexer already uses.
  - Minor fixes to the `reindex_master_copies` management command: `--from-block-number 0` is now a valid input, duplicate log line dropped, and `--block-process-limit` becomes optional (falls back to the indexer's configured value).

Closes PLA-1493

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -7,6 +7,9 @@ It is the preferred place to capture shared context and expectations.
 ## Indexer Architecture (Summary)
 This repository includes multiple indexers under `safe_transaction_service/history/indexers/`.
 
+## Linear
+- All tickets go into `PLATFORM` team and with labels `Backend` and `Transaction Service`
+
 ### Base Classes
 - `ethereum_indexer.py`
   - Orchestrates block range selection and indexing flow.
@@ -35,7 +38,7 @@ This repository includes multiple indexers under `safe_transaction_service/histo
   - Indexes Safe `ProxyCreation` events, inserts `SafeContract`.
 
 ### Key Behavioral Constraints (Must Preserve)
-- L1 uses `InternalTxIndexer` and L2 uses `SafeEventsIndexer`. They do not run in parallel.
+- L1 uses `InternalTxIndexer` and L2 uses `SafeEventsIndexer`. They never run in parallel.
 - Internal traces are matched only by `SafeMasterCopy` addresses (no other address list).
 - `EthereumTx` should not exist without at least one related table row.
 - Do not index errored transactions.
@@ -54,7 +57,8 @@ This repository includes multiple indexers under `safe_transaction_service/histo
 
 ## Testing
 - Virtualenv is managed by uv at `.venv`. Activate with `source .venv/bin/activate` or prefix commands with `uv run`.
-- Use `pytest` (no parameters). It should auto-detect configuration.
+- Dependencies must be up before running tests: `docker compose --profile develop up db redis ganache rabbitmq -d`
+- Use `pytest` (no parameters). It should auto-detect configuration. `DJANGO_SETTINGS_MODULE=config.settings.test uv run python -m pytest``
 - Use factories (e.g. `SafeContractFactory`) instead of inline helpers or `get_or_create` calls in tests.
 
 ## Performance Conventions
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -58,7 +58,7 @@ This repository includes multiple indexers under `safe_transaction_service/histo
 ## Testing
 - Virtualenv is managed by uv at `.venv`. Activate with `source .venv/bin/activate` or prefix commands with `uv run`.
 - Dependencies must be up before running tests: `docker compose --profile develop up db redis ganache rabbitmq -d`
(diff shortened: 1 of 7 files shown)


safe_transaction_service/history/services/index_service.py
```diff
@@ -598,49 +602,68 @@ def _reindex(
...
+                    "Error reindexing block range [%d, %d]: %s. "
+                    "Retrying with block_process_limit=1",
                     block_number,
-                    len(elements),
+                    chunk_to,
+                    exc,
                 )
-                element_number += len(elements)
+                indexer.block_process_limit = 1
+                continue
```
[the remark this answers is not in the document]
Uxio0: Management task can be interrupted and Celery task has a timeout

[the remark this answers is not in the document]
Uxio0: Either way, added here: dad75cad

safe_transaction_service/history/services/index_service.py
```diff
@@ -578,17 +581,21 @@ def _reindex(
...
         :param from_block_number:
         :param to_block_number:
-        :param block_process_limit:
+        :param block_process_limit: Initial chunk size. If ``None``, the indexer's
+            configured ``block_process_limit`` is used. The indexer's auto-adjust
+            mechanism may grow or shrink the limit during the run.
         :param addresses:
         :return: Number of reindexed elements
         """
+        from ..indexers import FindRelevantElementsException
```
falvaradorodriguez, line 597: Should this import go at the beginning of the file?

AGENTS.md
```diff
@@ -54,7 +57,8 @@ This repository includes multiple indexers under `safe_transaction_service/histo
 
 ## Testing
 - Virtualenv is managed by uv at `.venv`. Activate with `source .venv/bin/activate` or prefix commands with `uv run`.
-- Use `pytest` (no parameters). It should auto-detect configuration.
+- Dependencies must be up before running tests: `docker compose --profile develop up db redis ganache rabbitmq -d`
+- Use `pytest` (no parameters). It should auto-detect configuration. `DJANGO_SETTINGS_MODULE=config.settings.test uv run python -m pytest``
```
falvaradorodriguez:
```suggestion
- Use `pytest` (no parameters). It should auto-detect configuration. `DJANGO_SETTINGS_MODULE=config.settings.test uv run python -m pytest`
```

safe_transaction_service/history/tests/test_commands.py
```diff
@@ -183,7 +183,6 @@ def test_reindex_master_copies(self, current_block_number_mock: PropertyMock):
                 stdout=buf,
             )
             self.assertIn("Setting block-process-limit to 11", buf.getvalue())
```
falvaradorodriguez, line 185: Should we add tests for new cases where the parameter isn't required?

safe_transaction_service/history/services/index_service.py
```diff
@@ -598,49 +605,85 @@ def _reindex(
...
-            )
-            logger.info("Start reindexing addresses %s", addresses_str)
-            current_block_number = self.ethereum_client.current_block_number
-            stop_block_number = (
-                min(current_block_number, to_block_number)
-                if to_block_number
-                else current_block_number
+            return 0
+
+        if block_process_limit is not None:
```
falvaradorodriguez, line 612: Although the `indexer` parameter specifies that a new instance must be provided, it might be a good idea to include a safeguard such as: 


```
original_limit = indexer.block_process_limit
  if block_process_limit is not None:
      indexer.block_process_limit = block_process_limit
  try:
      # ... loop ...
  finally:
      indexer.block_process_limit = original_limit
```

WDYT?

safe_transaction_service/history/services/index_service.py
Uxio0: It prevents a circular import, take a look at __init__ on the IndexService

AGENTS.md
Uxio0: I shouldn't update AGENTS manually 😂 , done

safe_transaction_service/history/services/index_service.py
Uxio0: I think this is confusing... A lot of things break if you use the same indexer (like current block number and variables), so this is not our biggest problem

safe_transaction_service/history/tests/test_commands.py
Uxio0: Yes, done 05123ebf5c3b21a91bf2459b5ab2f8d7ee02e0e2

AGENTS.md
Uxio0: 05123ebf5c3b21a91bf2459b5ab2f8d7ee02e0e2

falvaradorodriguez (approved):
LGTM!

PR #2888, opened 2026-05-21
fix: allow owners registered as delegates to sign with multiple signatures

### LINKED ISSUE #2638: 422 status error when Signer is also a Proposer
**Describe the bug**

This error is thrown:

https://github.com/safe-global/safe-transaction-service/blob/7e92a360480908136c5b914d1dbd508eee5c5c36/safe_transaction_service/history/serializers.py#L287

When the following conditions are met:

**To Reproduce**
1. Add a proposer address ("Wallet A") to an existing safe
1. Then add Wallet A as a signer. In my case the safe ended up as a 3 of 5 on Base network after adding Wallet A
1. Wallet B proposes batch call signatures.
1. Wallet A tries to sign and receives error shown in link above `"Just one signature is expected if using delegates"`

Alternative reproduction:
Wallet A proposes a cancellation of the batch transaction. Wallet B and Wallet C are unable to sign and see the same error.

Assumption: if wallet B, C, D propose and sign without Wallet A, this issue does not present itself to anyone.

**Expected behavior**
A helpful error should tell the team to remove Wallet A as a proposer.
Or, no error should be shown at all, and Safes should function well if a wallet is both a Proposer and a Signer.

falvaradorodriguez:
- Closes #2638 

When a address is both a Safe owner and a registered proposer (delegate), submitting a multisig transaction with multiple signatures returned a 422 error: "Just one signature is expected if using delegates".

The validation iterated over all signatures and raised an error if any signer was found in the delegates table, without checking whether that signer was also an owner. This caused legitimate owner signatures to be rejected whenever the owner happened to be registered as a proposer.

**Fix**

Added an owner not in safe_owners guard to the delegate signature restriction in `SafeMultisigTransactionSerializer.validate()`. The single-signature restriction now only applies to pure delegates (non-owners). If an address is both an owner and a delegate, it is treated as an owner with no signature count restriction.

The `save()` method already handled this correctly, when sender is in safe_owners, it is stored as proposer directly with proposed_by_delegate=None. The fix makes validate() consistent with that existing behavior.

### DIFF
--- a/safe_transaction_service/history/serializers.py
+++ b/safe_transaction_service/history/serializers.py
@@ -294,7 +294,12 @@ class SafeMultisigTransactionSerializer(SafeMultisigTxSerializer):
                     f"Signer={owner} is not authorized to interact with the service"
                 )
 
-            if owner in delegates and len(parsed_signatures) > 1:
+            if (
+                owner in delegates
+                # An address that is both owner and delegate is treated as owner
+                and owner not in safe_owners
+                and len(parsed_signatures) > 1
+            ):
                 raise ValidationError(
                     "Just one signature is expected if using delegates"
                 )
--- a/safe_transaction_service/history/tests/test_views.py
+++ b/safe_transaction_service/history/tests/test_views.py
@@ -2290,6 +2290,93 @@ class TestViewsV150(SafeTestCaseMixin, APITestCase):
             response.data["non_field_errors"][0],
         )
 
+    def test_post_multisig_transactions_with_owner_and_delegate(self):
+        """Owner that is also a delegate should be able to sign alongside other owners."""
+        safe_owners = [Account.create() for _ in range(4)]
+        safe_owner_addresses = [s.address for s in safe_owners]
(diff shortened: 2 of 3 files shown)


moisses89:
This is happening when a transaction is being created, normally we expect one signature there, so for this case the issue is creating a transaction with more than 1 signature, instead of create a transaction with the one required signature and add confirmations, right?

falvaradorodriguez:
> This is happening when a transaction is being created, normally we expect one signature there, so for this case the issue is creating a transaction with more than 1 signature, instead of create a transaction with the one required signature and add confirmations, right?

That's correct for the typical flow. However, the proposal endpoint intentionally supports multiple signatures, and some clients use it to bundle collected signatures before reaching the threshold. The bug affects that flow whenever an owner happens to also be registered as a proposer. The fix ensures the single signature restriction only applies to pure delegates, which is the original intent of that validation.

Uxio0 (approved):

PR #2897, opened 2026-05-28
fix: Add EIP-1271 ready preparedSignature for nested Safes (PLA-1534)

Uxio0:
## Summary
  - Expose a new `preparedSignatureEip1271` field on the `SafeMessage` response that is ready to be consumed as a single `SafeSignature` whose owner is the Safe (e.g. for the v2 delegate endpoint when the delegator is a Safe).
  - Keep `preparedSignature` unchanged for backwards compatibility. Its `help_text` now notes it only works for EOA-only confirmations and points readers to the new field.
  - Closes [PLA-1534](https://linear.app/safe-global/issue/PLA-1534).

### DIFF
--- a/safe_transaction_service/safe_messages/models.py
+++ b/safe_transaction_service/safe_messages/models.py
@@ -11,7 +11,11 @@ from safe_eth.eth.django.models import (
     HexV2Field,
     Keccak256Field,
 )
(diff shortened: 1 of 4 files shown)


falvaradorodriguez (approved):
LGTM!

PR #2904, opened 2026-06-09
fix: Only emit transfer/ether events for tracked Safes (PLA-1606)

Uxio0:
## What

Only emit `INCOMING_*` / `OUTGOING_*` events whose addressed side is a tracked Safe, so the queue no longer publishes transfer/ether events addressed to DEX routers, pools, WETH or the zero address.

## Why

`build_event_payload` emitted **both** a directional `INCOMING_*` (`address = instance.to`) and `OUTGOING_*` (`address = instance._from`) payload for every transfer / ether `InternalTx`, with no check that the addressed side is a tracked Safe.

- **Tokens (ERC20/721) & L2 ether:** the row is legitimately stored (one side *is* the Safe), but the other side still got a bogus event.
- **L1 ether:** the internal-tx indexer monitors Safe **master-copy** addresses. When a tx merely touches a master copy (e.g. a Safe receives a fee → its proxy delegatecalls the singleton), `InternalTx.is_relevant` stores **every** value hop in the tx — router→router hops included — each emitting junk events on both sides.

Confirmed end-to-end on `0xb6c9d812…`: a MetaMask smart account (a Safe) receives a 0.0006 ETH fee deep in the internal txs; that pulled the whole swap in and produced `OUTGOING_ETHER` events for the 1inch router, MetaMask Swaps Spender and Swap Router. Downstream (balance-service) then had to call `getSafeInfo` on every event and drop the non-Safes — wasteful at ~10M events/day.

Closes PLA-1606.

## How

Annotate each model instance at index time — where Safe membership is already known in memory — with two transient (non-DB) attributes, then gate emission in `build_event_payload`. The same instance flows from `bulk_create` to the `post_bulk_create` signal, so the annotation survives to payload building. No per-event DB query, no Redis set, no DB connection in the spawned greenlet.

- **`event_service.build_event_payload`** — emits `INCOMING_*` only if `_to_is_a_safe`, `OUTGOING_*` only if `_from_is_a_safe`. New `set_safe_membership` helper keeps the attribute names in one place.
- **ERC20/721 (`erc20_events_indexer`)** — membership against the indexer's in-memory address cache.
- **L1 ether (`internal_tx_indexer`)** — per-tx Safe proxies derived from the full (pre-filter) traces: the `from` of any non-errored delegatecall whose `to` is a monitored master copy.
- **L2 ether (`safe_events_indexer`)** — the Safe is the event emitter (`safe_address`).

**Fail-closed with observability:** indexers always set both attributes (each possibly `False`). An *annotated* transfer with neither side a Safe (router→router) emits nothing, silently. An *unannotated* instance reaching payload building emits nothing **and** logs `error("Lacking metadata to build event for transfer with hash …")` with the tx hash and log_index/trace_address, so any missed path is visible rather than silently over-emitting.

No stored data changes. (Trimming the L1 ether noise rows from storage is intentionally out of scope — separate follow-up.)

### DIFF
--- a/safe_transaction_service/history/indexers/erc20_events_indexer.py
+++ b/safe_transaction_service/history/indexers/erc20_events_indexer.py
@@ -22,6 +22,7 @@ from ..models import (
     SafeRelevantTransaction,
     TokenTransfer,
 )
+from ..services.event_service import set_safe_membership
 from .events_indexer import EventsIndexer
 
 logger = getLogger(__name__)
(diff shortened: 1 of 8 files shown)


safe_transaction_service/history/indexers/internal_tx_indexer.py
```diff
@@ -419,6 +438,41 @@ def process_elements(
...
+        safe_proxies: set[ChecksumAddress] = set()
+        if not traces:
+            return safe_proxies
+        for trace in traces:
+            action = trace.get("action", {})
+            if (
+                EthereumTxCallType.parse_call_type(action.get("callType"))
+                == EthereumTxCallType.DELEGATE_CALL
+                and not trace.get("error")
+                and action.get("to") in master_copies
```
moisses89, line 470: Here we just check if the destination is a master_copy but the from can be any contract that do a delegate_call to a master_copy, right? so no necessary a Safe proxy.

safe_transaction_service/history/indexers/erc20_events_indexer.py
```diff
@@ -149,21 +150,55 @@ def _process_decoded_element(self, decoded_element: EventData) -> None:
...
                 pass
 
+    def _get_safe_addresses_for_membership(self) -> set[ChecksumAddress] | None:
+        """
+        :return: The full set of monitored Safe addresses held in memory, or ``None`` if the
+            address cache has not been populated yet (abnormal: ``process_elements`` invoked
+            outside the normal indexing cycle). ``None`` makes ``_annotate_safe_membership`` a
+            no-op, so `build_event_payload` logs a "lacking metadata" error rather than guessing.
+        """
+        return self.addresses_cache.addresses if self.addresses_cache else None
```
moisses89, line 181: The cache is a snapshot taken at the start of each indexing batch. 
A Safe created in the same batch as its first token transfer will not yet be present, so that transfer silently produces no event.

safe_transaction_service/history/indexers/internal_tx_indexer.py
Uxio0: Well yes, but that's pretty unusual and I don't think it's worth it to check the proxy bytecode due to maybe one person trying to trick the indexer

safe_transaction_service/history/indexers/erc20_events_indexer.py
Uxio0: If the safe is not in the address list it will not be indexed by the erc20 indexer either

moisses89 (approved):

PR #2906, opened 2026-06-15
feat: Add trusted flag to INCOMING_TOKEN/OUTGOING_TOKEN events

falvaradorodriguez:
- Closes [PLA-1661](https://linear.app/safe-global/issue/PLA-1661/add-trusted-flag-to-incoming-token-outgoing-token-webhook-events)

Adds a `trusted` boolean for token transfer events from an in-memory TokenService cache, avoiding per-event DB lookups on the indexing path. 

Also wire up the tokens app `ready()` so its post_save signal (cache invalidation) is actually connected.

### DIFF
--- a/safe_transaction_service/tokens/apps.py
+++ b/safe_transaction_service/tokens/apps.py
@@ -5,3 +5,6 @@ from django.apps import AppConfig
 class TokensConfig(AppConfig):
     name = "safe_transaction_service.tokens"
     verbose_name = "Tokens for Safe Transaction Service"
+
+    def ready(self):
+        from . import signals  # noqa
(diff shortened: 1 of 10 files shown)


Uxio0 (approved):

safe_transaction_service/tokens/apps.py
```diff
@@ -5,3 +5,6 @@
 class TokensConfig(AppConfig):
     name = "safe_transaction_service.tokens"
     verbose_name = "Tokens for Safe Transaction Service"
+
+    def ready(self):
+        from . import signals  # noqa
```
Uxio0, line 10: Signals were not working before?

falvaradorodriguez: It seems that it hasn't been working for tokens until now.

PR #2921, opened 2026-06-26
feat: Replace buzzfeed SSO with Google OIDC JWT authentication for Django admin

luarx:
## Summary

- Replaces deprecated buzzfeed SSO (`CustomHeaderRemoteUserMiddleware`) with `GoogleOIDCMiddleware`, which verifies the RS256-signed Google JWT forwarded as `X-Enc-ID-Token` by the reverse proxy using `google.oauth2.id_token.verify_oauth2_token`
- `CustomRemoteUserBackend.configure_user()` grants staff/superuser only to emails listed in `SSO_ADMINS` env var (`SSO_ADMINS=alice@safe.global,bob@safe.global`); all others are deactivated
- Adds `google-auth==2.55.0` for JWT verification; removes `google-api-python-client` (not needed)
- Adds 11 unit tests covering JWT validation, domain/email checks, and access control — no DB or network required

### What was wrong? 👾

Closes #


safe_transaction_service/utils/auth.py
```diff
@@ -1,36 +1,63 @@
...
+    def __call__(self, request):
+        token = request.META.get("HTTP_X_ENC_ID_TOKEN", "")
+        if token and request.user.is_authenticated:
+            logger.debug(
+                "SSO skipping JWT, session already active email=%s",
+                request.user.username,
+            )
+        elif token and not request.user.is_authenticated:
+            logger.info("SSO JWT verification started")
+            claims = id_token.verify_oauth2_token(token, google_requests.Request())
```
Uxio0: **🔴 JWT audience is never verified — token replay / confused deputy.**

`verify_oauth2_token(..., audience=None)` skips the `aud` (OAuth client ID) check entirely. The only extra gates are `hd == "safe.global"` and `email_verified`, so **any** Google-signed ID token for **any** OAuth client — issued to a totally unrelated app — is accepted as long as it carries a verified `safe.global` email.

If an admin ever authenticates to a malicious/third-party Google app, that app's ID token can be replayed here to gain their session (and admin, if their email is in `SSO_ADMINS`).

Pass the expected client ID: `audience=settings.SSO_GOOGLE_CLIENT_ID`.

safe_transaction_service/utils/auth.py
```diff
@@ -1,36 +1,63 @@
...
-        By default, return the user unmodified.
-        """
-        user = super().configure_user(request, user)
+class CustomRemoteUserBackend(RemoteUserBackend):
+    def configure_user(self, request, user, created=True):
+        user = super().configure_user(request, user, created)
+        if created:
+            logger.info("SSO user created email=%s", user.username)
         if user.username in settings.SSO_ADMINS:
             user.is_superuser = True
```
Uxio0: **🔴 Deactivated user can never be re-enabled via SSO — permanent lockout.**

`configure_user` runs on *every* login (Django's `RemoteUserBackend.authenticate` calls it unconditionally). The non-admin branch below sets `is_active=False` and saves. This admin branch sets `is_superuser`/`is_staff=True` but **never sets `is_active=True`**.

Sequence: a `safe.global` user logs in once before being added to `SSO_ADMINS` → row saved with `is_active=False`. Later added to `SSO_ADMINS`. On next login this branch flips the staff/superuser flags but leaves `is_active=False`, so `user_can_authenticate()` returns False → `authenticate()` returns None → locked out forever until a manual DB edit.

Add `user.is_active = True` in this branch.

safe_transaction_service/utils/auth.py
```diff
@@ -1,36 +1,63 @@
...
+        if token and request.user.is_authenticated:
+            logger.debug(
+                "SSO skipping JWT, session already active email=%s",
+                request.user.username,
+            )
+        elif token and not request.user.is_authenticated:
+            logger.info("SSO JWT verification started")
+            claims = id_token.verify_oauth2_token(token, google_requests.Request())
+            if claims.get("hd") != "safe.global":
+                raise ValueError("JWT rejected: wrong hosted domain")
```
Uxio0: **🟠 Invalid/expired token → HTTP 500 instead of clean denial.**

`verify_oauth2_token` raises `ValueError` on expired/tampered/malformed tokens, and these explicit `raise ValueError` lines do the same for wrong `hd` / unverified email. None of it is caught, so it bubbles out of the middleware as an unhandled 500 (noisy tracebacks/Sentry, log spam, trivial error-amplification from garbage headers).

Expired tokens are *normal* (Google ID tokens last ~1h). Wrap in `try/except ValueError`, log, and fall through as `AnonymousUser` — matching the old middleware's behavior on bad input.

safe_transaction_service/utils/auth.py
```diff
@@ -1,36 +1,63 @@
...
+    before creating a Django session via RemoteUserBackend.
     """
 
-    header = settings.SSO_USERNAME_HEADER
+    def __init__(self, get_response):
+        self.get_response = get_response
 
+    def __call__(self, request):
+        token = request.META.get("HTTP_X_ENC_ID_TOKEN", "")
+        if token and request.user.is_authenticated:
```
Uxio0: **🟠 Session is never re-validated after first login — IdP revocation doesn't take effect.**

Once a Django session cookie exists, this `token and is_authenticated` guard short-circuits and the JWT is never checked again. If the proxy stops forwarding `X-Enc-ID-Token` (token expired, user disabled at Google, signed out), the Django session keeps full admin until it expires server-side.

The old `RemoteUserMiddleware` had `force_logout_if_no_header=True` and logged out the moment the trusted header disappeared — that invariant is lost here. Consider re-verifying when a token is present, and/or a short `SESSION_COOKIE_AGE` for the admin.

safe_transaction_service/utils/auth.py
```diff
@@ -1,36 +1,63 @@
...
-    def configure_user(self, request, user):
-        """
-        Configure a user after creation and return the updated user.
 
-        By default, return the user unmodified.
-        """
-        user = super().configure_user(request, user)
+class CustomRemoteUserBackend(RemoteUserBackend):
+    def configure_user(self, request, user, created=True):
+        user = super().configure_user(request, user, created)
```
Uxio0: **🟠 `configure_user` overwrites manual role changes and writes on every login.**

Since this runs on every authenticate, it re-derives `is_superuser`/`is_staff`/`is_active` from `SSO_ADMINS` and `save()`s each login. Two effects:

1. Any change made manually in the Django admin (e.g. demoting a superuser) is silently reverted on that user's next login.
2. On demotion `is_active` flips to False but `is_superuser`/`is_staff` are **never reset to False**, so the stale privileged flags survive and would re-grant admin if the account is reactivated by any other path.

Also a redundant DB write per login.

safe_transaction_service/utils/auth.py
```diff
@@ -1,36 +1,63 @@
...
+                request.user.username,
+            )
+        elif token and not request.user.is_authenticated:
+            logger.info("SSO JWT verification started")
+            claims = id_token.verify_oauth2_token(token, google_requests.Request())
+            if claims.get("hd") != "safe.global":
+                raise ValueError("JWT rejected: wrong hosted domain")
+            if not claims.get("email_verified"):
+                raise ValueError("JWT rejected: email not verified")
+            user = authenticate(request, remote_user=claims["email"])
```
Uxio0: **🟡 `claims["email"]` can `KeyError` → 500.**

If a verified token passes the `hd`/`email_verified` checks but lacks the `email` claim (only present when the `email` scope was granted), this subscript raises `KeyError`, again uncaught → 500. Use `claims.get("email")` and reject cleanly if absent.

safe_transaction_service/utils/auth.py
```diff
@@ -1,36 +1,63 @@
...
+            if user:
+                request.user = user
+                login(request, user)
+                logger.info("SSO authenticated email=%s", claims["email"])
```
Uxio0: **🟡 Emails (PII) logged at INFO on every login + content-free noise line.**

This and other INFO lines carry `email=...` on the normal success path, and `logger.info("SSO JWT verification started")` above adds a diagnostic-free line per attempt. INFO logs usually ship to retained central stores — a PII/compliance footprint. Demote email-bearing success logs to DEBUG (keep WARNING for the denied/failed cases where the email aids triage) and drop the "started" line.

safe_transaction_service/utils/auth.py
```diff
@@ -1,36 +1,63 @@
...
+        token = request.META.get("HTTP_X_ENC_ID_TOKEN", "")
+        if token and request.user.is_authenticated:
+            logger.debug(
+                "SSO skipping JWT, session already active email=%s",
+                request.user.username,
+            )
+        elif token and not request.user.is_authenticated:
+            logger.info("SSO JWT verification started")
+            claims = id_token.verify_oauth2_token(token, google_requests.Request())
+            if claims.get("hd") != "safe.global":
```
Uxio0: **🟡 Hosted domain hardcoded; no cert/session reuse.**

`"safe.global"` is hardcoded — adding a domain or rotating config needs a code change, inconsistent with the `env.list("SSO_ADMINS")` pattern next to it in `production.py`. Make it a setting.

Separately, a fresh `google_requests.Request()` is built per call (line 33), so Google's JWKS certs aren't reused across logins; a module-level `Request()` (carrying a `requests.Session`) lets HTTP caching amortize the fetch. Low impact — only fires on the session-establishing request.

safe_transaction_service/utils/auth.py
```diff
@@ -1,36 +1,63 @@
...
+        self.get_response = get_response
 
+    def __call__(self, request):
+        token = request.META.get("HTTP_X_ENC_ID_TOKEN", "")
+        if token and request.user.is_authenticated:
+            logger.debug(
+                "SSO skipping JWT, session already active email=%s",
+                request.user.username,
+            )
+        elif token and not request.user.is_authenticated:
```
Uxio0: **🟢 Redundant `if`/`elif`.**

Both branches require `token` truthy and differ only on `is_authenticated`; the first only emits a debug line. Collapses to a single early guard:

```python
if not token or request.user.is_authenticated:
    return self.get_response(request)
```

config/settings/production.py
```diff
@@ -55,26 +55,25 @@
 # https://docs.djangoproject.com/en/5.0/ref/settings/#csrf-trusted-origins
 CSRF_TRUSTED_ORIGINS = env.list("CSRF_TRUSTED_ORIGINS", default=[])
 
-# SSO (tested with https://github.com/buzzfeed/sso)
+# SSO — Google OIDC + JWT verification
 # ------------------------------------------------------------------------------
-# Be really careful when enabling SSO. If the `SSO_USERNAME_HEADER` can be spoofed
-# auth is broken and anyone will be able to log in as any user
+# The reverse proxy handles the Google OIDC flow and forwards the raw RS256-signed
```
Uxio0: **🟢 Header contract change is undocumented.**

`SSO_USERNAME_HEADER` and `CustomHeaderRemoteUserMiddleware` are fully removed (grep confirms no dangling refs), but nothing in the repo documents the new `X-Enc-ID-Token` header the proxy must forward, or that `SSO_USERNAME_HEADER` is now ignored. Operators upgrading get silent login failure. Worth a note in deployment docs / `.env` example.

(Not a bug: changing `SSO_ADMINS` default to `[]` is a good deny-by-default improvement — just ensure deployments set it.)

Uxio0 (commented):
Code review of the Google OIDC admin SSO change. The diff is small but security-critical — flagging findings inline, roughly ordered by severity. The top two (missing `aud` verification and the permanent-lockout path) are the ones I'd block on.

safe_transaction_service/utils/auth.py
luarx: 1. Intentional. SSO_ADMINS is the source of truth, manual changes in Django admin would drift from it. To demote someone, remove them from SSO_ADMINS.
2. Fixed.
3. Agreed it's a write per login, but this only runs on session creation and the code stays simpler without a dirty-check.

safe_transaction_service/utils/auth.py
luarx: We keep ValueError unhandled intentionally, a malformed or tampered token reaching the middleware indicates something unexpected upstream (APISIX should handle expiry/refresh before forwarding). A 500 makes that failure visible rather than silently swallowing it. We'd rather alert on it.

safe_transaction_service/utils/auth.py
luarx: Already fixed — the admin branch sets now is_active = True, is_superuser = True, and is_staff = True.

safe_transaction_service/utils/auth.py
```diff
@@ -1,36 +1,201 @@
...
+                "SSO_ADMINS membership changed, updating is_superuser",
+                extra={
+                    "extra_data": {
+                        "email": user.username,
+                        "old": user.is_superuser,
+                        "new": is_sso_admin,
+                    }
+                },
+            )
+            CustomRemoteUserBackend.apply_user_flags(user, is_sso_admin=is_sso_admin)
```
moisses89: When an admin is removed from SSO_ADMINS, they should lose all admin access. Currently only is_superuser is cleared, is_staff remains True, leaving them able to access Django admin.

safe_transaction_service/utils/auth.py
moisses89: > Separately, a fresh google_requests.Request() is built per call (line 33), so Google's JWKS certs aren't reused across logins; a module-level Request() (carrying a requests.Session) lets HTTP caching amortize the fetch. Low impact — only fires on the session-establishing request.

@luarx was this intentionally ignored?

safe_transaction_service/utils/auth.py
moisses89: @luarx was this considered?

safe_transaction_service/utils/auth.py
luarx: it is outdated this comment, the new code already supports this

safe_transaction_service/utils/auth.py
luarx: it is outdated this comment, new code already considers this

safe_transaction_service/utils/auth.py
moisses89: This is still recreated on every login, you could reuse the connection instead:
```
_google_request = google_requests.Request()

def _login(self, request, token):
    claims = id_token.verify_oauth2_token(
        token,
        _google_request,  # reused
        audience=settings.SSO_CLIENT_ID,
    )
```

moisses89 (dismissed):
I just left some suggestions

safe_transaction_service/utils/auth.py
luarx: Now we reuse the connection and reauthenticate the JWT token

moisses89 (dismissed):

Uxio0 (dismissed):

moisses89 (approved):

PR #2932, opened 2026-07-07
Parallelize trace_transactions chunks

Uxio0:
## Problem

`trace_transaction` is the most expensive call in the tracing indexing path, and `InternalTxIndexer.trace_transactions` fetched its chunks sequentially — the batches don't depend on each other, so this is wasted wall-clock on mainnet/gnosis catch-up.

## Fix

Fetch the chunks concurrently with a gevent pool, mirroring `EventsIndexer._do_node_query`'s existing `get_logs` parallelization: chunks are spawned into a `gevent.pool.Pool(...)`, joined with `raise_error=True`, and results are concatenated in chunk order so callers can still `zip` them against `tx_hashes_missing_traces`. The existing `OSError` logging / re-raise behavior is preserved.

Concurrency is controlled by a new setting `ETH_INTERNAL_TXS_TRACE_TXS_CONCURRENCY` (default `3`), wired the same way as `ETH_EVENTS_GET_LOGS_CONCURRENCY`.

## Deferred follow-up

A related `MultisigConfirmation` N+1 in `tx_processor` was **intentionally left out** of this PR. Both `get_or_create` sites carry update semantics (back-fill `ethereum_tx`, rewrite changed signatures) plus within-batch read-after-write, so a naive `bulk_create(ignore_conflicts=True)` would silently drop those updates when the row already exists (common via the queue-service flow) — a correctness regression. A proper fix needs its own PR: accumulate intended confirmations, one existence query per batch, then split into `bulk_create` + `bulk_update` with within-batch dedup.

### DIFF
--- a/safe_transaction_service/history/indexers/internal_tx_indexer.py
+++ b/safe_transaction_service/history/indexers/internal_tx_indexer.py
@@ -76,9 +76,7 @@ class InternalTxIndexer(EthereumIndexer):
 
         self.trace_txs_batch_size: int = settings.ETH_INTERNAL_TRACE_TXS_BATCH_SIZE
         # Number of concurrent `trace_transactions` batch requests
-        self.trace_txs_concurrency: int = (
-            settings.ETH_INTERNAL_TXS_TRACE_TXS_CONCURRENCY
-        )
+        self.trace_txs_concurrency: int = settings.ETH_INTERNAL_TRACE_TXS_CONCURRENCY
         self.number_trace_blocks: int = settings.ETH_INTERNAL_TXS_NUMBER_TRACE_BLOCKS
         self.tx_decoder = get_safe_tx_decoder()
 
@@ -292,16 +290,12 @@ class InternalTxIndexer(EthereumIndexer):
             return []
 
         batch_size = batch_size or len(tx_hashes)  # If `0`, don't use batches
-        tx_hash_chunks = [
-            list(tx_hash_chunk) for tx_hash_chunk in chunks(list(tx_hashes), batch_size)
-        ]
-
         gevent_pool = pool.Pool(self.trace_txs_concurrency)
         jobs = [
             gevent_pool.spawn(
                 self.ethereum_client.tracing.trace_transactions, tx_hash_chunk
             )
-            for tx_hash_chunk in tx_hash_chunks
+            for tx_hash_chunk in chunks(list(tx_hashes), batch_size)
         ]
 
         try:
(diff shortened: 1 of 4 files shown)


safe_transaction_service/history/indexers/internal_tx_indexer.py
```diff
@@ -271,22 +277,46 @@ def _get_internal_txs_to_decode(
...
-                    "Try lowering ETH_INTERNAL_TRACE_TXS_BATCH_SIZE",
-                    len(tx_hash_chunk),
-                    exc_info=True,
-                )
-                raise
+        tx_hash_chunks = [
+            list(tx_hash_chunk) for tx_hash_chunk in chunks(list(tx_hashes), batch_size)
+        ]
+
+        gevent_pool = pool.Pool(self.trace_txs_concurrency)
```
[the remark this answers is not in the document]
Uxio0: Fixed on 03a2a58c361fa2ac4c52a65eeda2143ea0466cc7

safe_transaction_service/history/indexers/internal_tx_indexer.py
```diff
@@ -271,22 +280,42 @@ def _get_internal_txs_to_decode(
...
+        jobs = [
+            gevent_pool.spawn(
+                self.ethereum_client.tracing.trace_transactions, tx_hash_chunk
+            )
+            for tx_hash_chunk in chunks(list(tx_hashes), batch_size)
+        ]
+
+        try:
+            gevent.joinall(jobs, raise_error=True)
```
falvaradorodriguez, line 307: gevent.joinall(jobs, raise_error=True) raises on the first chunk that finishes with an error (in completion order, not spawn order). It doesn't wait for or kill the remaining jobs. Two implications worth being explicit about, even though this mirrors EventsIndexer's existing get_logs pattern:                    
                                                                                                                                                                                                                                                                                                                            
1. One failing chunk aborts the whole trace_transactions() call, discarding results from chunks that already succeeded. Not a regression (the old sequential version had the same "one bad chunk fails the whole batch" contract), just worth calling out since it's now non-deterministic which chunk trips it.          
2. Sibling greenlets that are still in flight when the exception propagates are never killed. They'll finish on their own, but if one of them later raises with nobody watching (.get()/join()), gevent will print an unhandled-exception traceback to stderr, not a correctness issue.                       
                                                                                                                                                                                                                                                                                                                            
  Not a blocker for this PR since it's consistent with the existing events_indexer.py pattern, but a finally: gevent.killall(jobs) (applied to both indexers) would tighten this up as a follow-up.

Uxio0: Totally agree @falvaradorodriguez , fixed on 63bf62e7

falvaradorodriguez (approved):

PR #2933, opened 2026-07-07
Fix erc20 out-of-sync loop, conditional-indexing FK gap, batch delete

Uxio0:
Three related indexing fixes across `history/tasks.py` and `history/indexers/safe_events_indexer.py`.

## 1. Infinite loop in `index_erc20_events_out_of_sync_task`

The `while not updated:` loop swallowed `FindRelevantElementsException` with a bare `pass`. If the RPC kept failing, `updated` never became `True` and the loop spun forever, hammering the node until the Celery timeout killed the task. Added a consecutive-failure guard (`MAX_CONSECUTIVE_FAILURES = 3`): the counter resets after any successful `process_addresses`, and on the 3rd consecutive failure it logs an error and breaks, returning events processed so far.

## 2. FK integrity gap in conditional indexing

`_fetch_receipts_and_store` silently skipped any tx whose receipt fetch failed (warning only), but the corresponding events were still processed into `InternalTx` rows with `ethereum_tx_id` pointing at a tx that was never inserted → `IntegrityError`, forcing the slow one-by-one fallback and dropping the rows. Their blocks also missed the timestamp prefetch, hitting the `KeyError` error path.

Now `_fetch_receipts_and_store` returns the set of tx hashes actually stored. Events whose tx is not present in the DB (`db_txs` ∪ stored) are dropped before processing, and their log receipts are left **unmarked** so they get retried on a later run. Intentional drops (blocklisted / filtered-out receipts) stay marked as before — the exclusion is keyed on receipt-fetch failures only.

## 3. Per-Safe delete inside a loop

`_process_safe_creation_events` ran `InternalTx.objects.filter(contract_address=safe_address).delete()` once per Safe in the SafeSetup-without-ProxyCreation branch. Now the addresses are collected and deleted in a single `filter(contract_address__in=...).delete()` after the loop, before the store.

### DIFF
--- a/safe_transaction_service/history/indexers/safe_events_indexer.py
+++ b/safe_transaction_service/history/indexers/safe_events_indexer.py
@@ -228,21 +228,45 @@ class SafeEventsIndexer(EventsIndexer):
         ]
 
         # 11. Fetch receipts and store ONLY for txs with processable events
+        stored_fetched_tx_hashes: set[bytes] = set()
         if allowed_fetched_txs_filtered:
-            number_allowed_txs_inserted = self._fetch_receipts_and_store(
+            stored_fetched_tx_hashes = self._fetch_receipts_and_store(
                 allowed_fetched_txs_filtered
             )
             logger.debug(
-                "Conditional indexing: %d txs with processable events inserted",
-                number_allowed_txs_inserted,
+                "Conditional indexing: %d txs with processable events stored",
+                len(stored_fetched_tx_hashes),
             )
 
-        # 12. Process only the processable events
-        # (filtering already done by _get_processable_events)
+        # Txs whose receipt fetch failed were NOT stored. Processing their events
+        # would create InternalTx rows pointing at a missing EthereumTx
+        # (IntegrityError), so drop those events and leave their log receipts
+        # unprocessed so they get retried on a later run.
+        tx_hashes_present_in_db = set(db_txs.keys()) | stored_fetched_tx_hashes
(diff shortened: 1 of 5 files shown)


safe_transaction_service/history/indexers/safe_events_indexer.py
```diff
@@ -986,18 +1006,13 @@ def _fetch_receipts_and_store(self, txs: list[TxData]) -> int:
...
                 tx["hash"]
             )  # Retry if failed
-            if tx_receipt:
-                txs_with_receipts.append((tx, tx_receipt))
-            else:
-                logger.warning(
-                    "Conditional indexing: failed to fetch receipt for tx %s",
-                    to_0x_hex_str(tx["hash"]),
+            if not tx_receipt:
+                raise TransactionNotFoundException(
```
falvaradorodriguez, line 1010: This exception isn't in the tuple process_addresses catches in ethereum_indexer.py (FindRelevantElementsException, SoftTimeLimitExceeded, Timeout, ValueError, Web3RPCError), so it skips the block_process_limit = 1 reset and propagates without throttling the next attempt. Should it be managed there?

Uxio0: This behaviour was existing before... So I didn't want to touch it. We consider a missing transaction a RPC temporary issue and not a problem for the indexer to slow down

falvaradorodriguez (approved):

PR #2936, opened 2026-07-09
Avoid IntegrityError when storing already indexed InternalTxs

Uxio0:
## Summary

When reindexing (`reindex_master_copies_task`), most of the `InternalTx` are already stored, so the `bulk_create` in `store_internal_txs_and_decoded_in_db()` raises `IntegrityError` on almost every batch:

```
django.db.utils.IntegrityError: duplicate key value violates unique constraint "unique_internal_tx_trace_address"
```

The exception is handled by falling back to inserting the transactions one by one, but that fallback is slow (one transaction per row) and monitoring tools report the raised exception as an error, polluting error tracking.

## Changes

- Before inserting, fetch the `(ethereum_tx_id, trace_address)` pairs already in database (single query on the unique index) and filter out the matching `InternalTx` and their related `InternalTxDecoded`.
- The existing `IntegrityError` fallback is kept as a safety net for real races with the live indexer, but it should almost never be reached now.
- Add a regression test checking that already stored `InternalTx` and their `InternalTxDecoded` are filtered out.

## Details

New query plan result:

```sql
Index Only Scan using unique_internal_tx_trace_address on history_internaltx
    (cost=0.42..451.17 rows=517 width=35) (actual time=0.030..0.069 rows=500 loops=1)
  Index Cond: (ethereum_tx_id = ANY ('{...100 hashes...}'::bytea[]))
  Heap Fetches: 0
  Index Searches: 1
  Buffers: shared hit=9
Planning Time: 0.119 ms
Execution Time: 0.093 ms
```

Part of PLA-1742


falvaradorodriguez (approved):
LGTM!

PR #2938, opened 2026-07-10
Send queue events only after database commit

### LINKED ISSUE #2391: `PENDING_MULTISIG_TRANSACTION` event is emitted before confirmations are added
For sending notifications and events to the event queue we use [Django signals](https://docs.djangoproject.com/en/5.1/topics/signals/). `confirmations` are stored in a different table. So first we store `MultisigTransaction`, and then we create `Confirmation` and link it using a foreign key. The problem is that the event triggers when `MultisigTransaction` is created, so if a client checks the endpoint the transaction will miss the signers. Delay should be only a few seconds, but that's not acceptable.

## Solutions
- Invalidate signals and trigger then manually at the end: tricky and hacky solution
- Remove signals and manually manage when events should be emitted. This gives us total control but requires a big refactor of the service

Temporary we recommend to clients to add a small delay when they receive a `PENDING_MULTISIG_TRANSACTION` or `EXECUTED_MULTISIG_TRANSACTION` event

Uxio0:
Send queue events only after database commit

Related to #2391

## Motivation

Events are published to RabbitMQ from Django signals while the database transaction is still open. Subscribed services sometimes receive the event and query the API before the data is committed, getting a 404 or incomplete data. A rollback after publishing also produces events for data that never existed.

## Changes

- Add `send_events_on_commit`: payloads are built eagerly in the signal handlers (delete events need the rows before they are gone), and only the publish is deferred with `transaction.on_commit(robust=True)`. With no transaction open it publishes immediately. On rollback the events are discarded.
- Add `QueueService.send_events` to publish a batch of events acquiring a producer from the pool only once. `send_event` and `send_unsent_events` now share the same publish loop.
- Publishing never raises: unpublished events are always buffered. Before, an exhausted producer pool raised `LimitExceeded` (not an `OperationalError`), which would have escaped and lost the events.
- One unserializable payload no longer prevents its sibling events from being published.
- `post_bulk_create` events are processed synchronously by the same `process_event` receiver used for `post_save`, removing the fire-and-forget greenlets and their connection cleanup. The greenlets existed to take the per-row broker publish out of the indexing loop, but the publish is deferred to commit now. This also fixes a second race: the greenlets ran on their own database connection, so they could not see the still-uncommitted rows when building payloads. With no transaction open the events are sent directly, so `transaction.on_commit` does not force-open a database connection (`get_autocommit()` -> `ensure_connection()`).
- The ERC20/721 indexer stores transfers and `SafeRelevantTransaction` in one transaction, so transfer events are published only after all the data is visible, deferred and batched at commit instead of per row. On error nothing is stored and no event is sent, so retries cannot duplicate events.
- View cache is invalidated eagerly and again after commit, so a concurrent request repopulating the cache with pre-commit data does not leave it stale. The invalidation never raises (a Redis failure must not abort indexer batches now that dispatch is synchronous), uses a single variadic `UNLINK`, and the on-commit callback holds only the resolved cache tag and addresses, not the model instance.
- Tests wrap writes in `captureOnCommitCallbacks(execute=True)`, as `TestCase` never commits.

## Follow-ups (out of scope)

- The API serializers do multiple autocommit writes (`SafeMultisigTransactionSerializer.save` creates the transaction, then the confirmations), so the event can still go out between them. Wrapping those serializers in `transaction.atomic` is needed to fully close #2391 on the API path.
- Events registered in the same transaction are still published one batch per signal at commit. A per-transaction accumulator could publish them all with a single producer acquire.
- Exceptions building a payload now propagate into the caller (same as the `post_save` path always did): an indexer batch rolls back and retries instead of silently losing the event like the dead greenlet did.


safe_transaction_service/history/signals.py
```diff
@@ -168,46 +193,45 @@ def _process_event(
...
-                "[%s] Triggering send_event tasks for created=%s object=%s",
-                address,
-                created,
-                instance,
-            )
-            queue_service = get_queue_service()
-            queue_service.send_event(payload)
+    get_queue_service().send_events_on_commit(
+        [payload for payload in payloads if payload.get("address")]
+    )
```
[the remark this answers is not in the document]
Uxio0: Fixed in e6e60920

moisses89 (dismissed):
We should test this carefully, from the PR description I understand that you are aware of the following:
- we are adding one more call per instance to redis
- we are removing spawn, so we lost the concurrent unlink and send_event
in case that spwan removal can be problematic, I'would propose to call JOIN and wait until spawn finish, maybe this fix the orphaned spawns.

Uxio0:
Yes, I'm aware of both. But you made good points:

> we are adding one more call per instance to redis

Only when it's enabled, and it prevents race conditions that were not fixed previously. But I'm trying a better implementation.

> we are removing spawn

Agree. But:
- For not bulk create, `spawn` was not used
- For bulk create, `spawn` was used, but now everything will be done after transaction is committed in the same greenlet as the original task, but with everything already on database

Uxio0:
We are only doing one single redis call: dc53f2e1d504fbc3eebbc80a1dd7149226603f84

safe_transaction_service/events/services/queue_service.py
```diff
@@ -4,16 +4,61 @@
...
+        Payloads must be built eagerly by the caller (inside the transaction)
+        — only the publish is deferred. ``send_events`` never raises (failed
+        events are buffered), and ``robust=True`` keeps any unexpected failure
+        from preventing sibling ``on_commit`` callbacks.
+
+        :param payloads: Event payloads to publish after commit
+        """
+        if not payloads:
+            return
+        if transaction.get_connection().in_atomic_block:
```
falvaradorodriguez, line 50: For the "one batch per signal" follow-up you mention, since remove_cache_view_for_addresses already solves this same shape for the cache (cache.py:214, accumulating on connection.__dict__ and registering on_commit only once per transaction), it might be a small lift to reuse that same pattern here for queue events too, rather than a separate accumulator design. WDYT?

safe_transaction_service/events/services/queue_service.py
```diff
@@ -90,44 +135,79 @@ def _build_routing_key(payload: dict[str, Any]) -> str:
...
-            logger.debug("Adding %s to unsent messages", payload)
-            self.unsent_events.append((event, routing_key))
+        events: list[tuple[bytes, str]] = []
+        for payload in payloads:
+            try:
+                events.append((orjson.dumps(payload), self._build_routing_key(payload)))
+            except Exception:
+                logger.error("Cannot serialize payload %s", payload, exc_info=True)
+        if not events:
             return 0
```
falvaradorodriguez: if not events: return 0 skips send_unsent_events() when every payload in this batch fails to serialize. If there's a backlog in self.unsent_events from an earlier failure, it won't get flushed on this call. Maybe check self.unsent_events before returning here, even when events is empty?

falvaradorodriguez (dismissed):

safe_transaction_service/events/services/queue_service.py
Uxio0: Agree, good catch. Fixed on 813d290f58b944046c64287613615bf17df8a878

falvaradorodriguez (approved):

safe_transaction_service/events/services/queue_service.py
Uxio0: I thought about it, but RabbitMQ does not accept multiple events in one call, so we would be in a similar situation making the function more complex. Also in case of a rollback events would still be stored in the connection and can be sent, so we would be up for some nasty bugs

PR #2942, opened 2026-07-14
Lazily fetch token metadata on tokens detail endpoint miss

vseehausen:
## Summary

`GET /api/v1/tokens/{address}/` (`TokenView`) only served tokens already catalogued by the balances/collectibles endpoints (which populate them lazily via `create_from_blockchain`). Addresses that appear only in a Safe's transaction history — fully-sent tokens, spam/airdrops, NFTs, tokens the Safe no longer holds — are never populated, so per-transaction metadata lookups (e.g. from the client gateway) miss.

In production this endpoint runs ~9.7k req/hr across the tx-service fleet with **~85% returning 404** (~4.3M req/month). Since Django's `cache_page` only caches 200s, every miss re-hits the DB and repeats on each client poll.

## Changes

- `TokenView.get_object`: on a miss, fetch the token from the blockchain via `Token.objects.create_from_blockchain` and return it (populating + caching it). Invalid tokens are still recorded in `TokenNotValid` and return 404.
- An RPC/node failure degrades to a 404 instead of a 5xx and is **not** blacklisted, so it retries once the node recovers (throttled by the client's negative cache).
- Tests for the three paths: valid → lazily created + 200, invalid → 404 + recorded + no repeat RPC, RPC failure → 404 + not blacklisted.

## Details

- Only the detail view is affected (`RetrieveAPIView.get_object`); the list views are untouched.
- Concurrent first-requests for the same token are already handled by the `IntegrityError` / `transaction.atomic` guard in `create_from_blockchain`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/safe_transaction_service/tokens/tests/test_views.py
+++ b/safe_transaction_service/tokens/tests/test_views.py
@@ -10,17 +10,31 @@ from eth_account import Account
 from rest_framework import status
 from rest_framework.exceptions import ErrorDetail
 from rest_framework.test import APITestCase
-from safe_eth.eth.ethereum_client import Erc20Manager, InvalidERC20Info
+from safe_eth.eth.ethereum_client import (
+    Erc20Info,
+    Erc20Manager,
+    Erc721Manager,
+    InvalidERC20Info,
+    InvalidERC721Info,
+)
 from safe_eth.safe.tests.safe_test_case import SafeTestCaseMixin
 
-from ..models import Token
+from ..models import Token, TokenNotValid
 from .factories import TokenFactory, TokenListFactory
 
 logger = logging.getLogger(__name__)
 
 
 class TestTokenViews(SafeTestCaseMixin, APITestCase):
-    def test_token_view(self):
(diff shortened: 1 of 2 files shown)


vseehausen:
Closing — this solves a non-problem. The high 404 rate on `GET /api/v1/tokens/{address}/` is expected: CGW uses this endpoint (alongside `/safes/` and `/contracts/`) as an address type/metadata probe, so a 404 legitimately means "this address isn't a token," not missing data. Lazily fetching from the blockchain on a miss would add 1–2 RPC calls plus a `TokenNotValid` row for every unindexed non-token address probed (the majority), raising node load for no benefit on that path. The effective mitigation is CGW-side negative caching of these 404s; the root-cause improvement would be a single (batch) address-info/classification endpoint backed by the existing SafeContract/Contract/Token tables, so CGW stops fanning out to multiple endpoints and interpreting 404s. Thanks to the team for the context.

PR #2943, opened 2026-07-15
fix: Add CELERY_CONTROL_QUEUE_DURABLE setting to fix worker liveness probe on RabbitMQ >=4.3

luarx:
## Summary

RabbitMQ 4.3.0 denies "transient, non-exclusive" queue declarations by default. Celery's `inspect ping` command — which our worker liveness probe calls — declares exactly that kind of queue for its control/pidbox mailbox, so on any network running RabbitMQ >=4.3 the probe fails every time,

This adds `CELERY_CONTROL_QUEUE_DURABLE` (defaults to `true`), which makes that mailbox durable instead of transient, escaping RabbitMQ's restriction without touching real task queues/routing/delivery.

## Why durable, not exclusive

The other option Celery offers (`control_queue_exclusive`) also escapes the restriction, but ties the mailbox to a single AMQP connection — reproduced against a real RabbitMQ 4.3.2 container that this **crashes a worker outright** (`ResourceLocked 405`) if it restarts under the same hostname while the broker still has the old connection registered.

## Why default `true`

Durable+shared is RabbitMQ's oldest, most universally-supported queue type, so it's harmless on RabbitMQ <4.3 too. 

- [ ] You ran `./run_tests.sh`
- [ ] You ran `pre-commit run -a`

### What was wrong? 👾

Closes #

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -252,6 +252,17 @@ CELERY_BROKER_POOL_LIMIT = env.int("CELERY_BROKER_POOL_LIMIT", default=0)
 # https://docs.celeryq.dev/en/stable/userguide/configuration.html#broker-heartbeat
 CELERY_BROKER_HEARTBEAT = env.int("CELERY_BROKER_HEARTBEAT", default=120)
 
+# Not in Celery's official docs, but supported in `celery.app.control.Control` and
+# `kombu.pidbox.Mailbox`. RabbitMQ >=4.3 denies the "transient, non-exclusive" queue
+# that Celery's control/inspect mailbox (used by `celery inspect ping`, our worker
+# liveness probe) declares by default, breaking the probe. Making the mailbox durable
+# instead of exclusive avoids that RabbitMQ restriction without tying it to a single
+# connection, which would crash a worker that restarts under the same hostname.
+# Defaults to true: durable+shared is RabbitMQ's oldest, universally-supported queue
+# type, so it's harmless on RabbitMQ <4.3 too, and every network eventually drifts
+# onto >=4.3 since the RabbitMQ image tag isn't pinned.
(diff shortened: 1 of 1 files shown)


agusgonzaleznic (approved):

config/settings/base.py
```diff
@@ -252,6 +252,17 @@
...
+# Not in Celery's official docs, but supported in `celery.app.control.Control` and
+# `kombu.pidbox.Mailbox`. RabbitMQ >=4.3 denies the "transient, non-exclusive" queue
+# that Celery's control/inspect mailbox (used by `celery inspect ping`, our worker
+# liveness probe) declares by default, breaking the probe. Making the mailbox durable
+# instead of exclusive avoids that RabbitMQ restriction without tying it to a single
+# connection, which would crash a worker that restarts under the same hostname.
+# Defaults to true: durable+shared is RabbitMQ's oldest, universally-supported queue
+# type, so it's harmless on RabbitMQ <4.3 too, and every network eventually drifts
+# onto >=4.3 since the RabbitMQ image tag isn't pinned.
+CELERY_CONTROL_QUEUE_DURABLE = env.bool("CELERY_CONTROL_QUEUE_DURABLE", default=True)
```
moisses89, line 264: `durable+non-exclusive` orphaned worker queues never get cleaned up and will silently accumulate unacked messages indefinitely

moisses89 (changes requested):

config/settings/base.py
moisses89: After check it, celery have the following option `control_queue_expires` that is 10 by default. 
This option configure the `x-expire` on Rabbitmq https://www.rabbitmq.com/docs/ttl#queue-ttl
So the orphaned situation is managed correctly by celery in case of `durable+non-exclusive`.

moisses89 (approved):

config/settings/base.py
luarx: Thanks for double checking!

PR #2967, opened 2026-07-29
Decode Safe ProxyFactory 1.5.0 creation methods

Uxio0:
## What

Decode Safe `ProxyFactory` 1.5.0 creation methods in `get_safe_creation_info`.

1.5.0 adds `createProxyWithNonceL2` and `createChainSpecificProxyWithNonceL2`. `_decode_proxy_factory` only tried the 1.3.0 and 1.4.1 ABIs, so Safes created with the new methods returned no decoded creation info (singleton / initializer / saltNonce). The legacy `createProxyWithNonce` keeps the same signature across versions and was already decoding fine.

## Changes

- Build `proxy_factory_v1_5_0_contract` in `SafeService.__init__`.
- Rewrite `_decode_proxy_factory` to loop over the 1.3.0 → 1.4.1 → 1.5.0 ABIs, breaking on the first successful `decode_function_input`. The existing singleton/initializer/saltNonce extraction already covers the 1.5.0 arg names.
- Add `test_decode_creation_data_v1_5_0` covering `createProxyWithNonce`, `createProxyWithNonceL2`, and `createChainSpecificProxyWithNonceL2`.

## Notes

- The two new L2 methods internally call the non-L2 version, which still emits the standard `ProxyCreation(indexed proxy, singleton)` event. So the log-based multi-deployment matching in `_process_creation_data` and the indexers need no change — verified that the 1.5.0 `ProxyCreation` topic is identical to the existing one and that the extra `ProxyCreationL2` / `ChainSpecificProxyCreationL2` events are not double-counted.
- No selector collisions across the three ABIs, so loop order never mis-decodes.

Closes WA-2935.

### DIFF
--- a/safe_transaction_service/history/indexers/proxy_factory_indexer.py
+++ b/safe_transaction_service/history/indexers/proxy_factory_indexer.py
@@ -57,6 +57,9 @@ class ProxyFactoryIndexer(EventsIndexer):
             # event ProxyCreation(GnosisSafeProxy proxy, address singleton)
             proxy_factory_v1_3_0_contract.events.ProxyCreation(),
             # event ProxyCreation(SafeProxy indexed proxy, address singleton)
+            # v1.5.0 emits this same ProxyCreation signature (also from its
+            # createProxyWithNonceL2 / createChainSpecificProxyWithNonceL2 methods),
+            # so the v1.4.1 entry decodes 1.5.0 proxy creations too.
             proxy_factory_v_1_4_1_contract.events.ProxyCreation(),
         ]
 
--- a/safe_transaction_service/history/indexers/safe_events_indexer.py
+++ b/safe_transaction_service/history/indexers/safe_events_indexer.py
@@ -528,6 +528,8 @@ class SafeEventsIndexer(EventsIndexer):
                 # Change Master Copy
                 safe_v1_1_1_contract.events.ChangedMasterCopy(),
                 # Proxy creation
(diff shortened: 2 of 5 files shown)


falvaradorodriguez (dismissed):
`safe_events_indexer` and `proxy_factory_indexer` list `ProxyCreation` one entry per factory version (1.1.1 / 1.3.0 / 1.4.1), but 1.5.0 isn't there. Harmless today because  `ProxyCreation` topic is byte-identical to 1.4.1's, so that entry already matches the logs. But nothing in the code says so, and the next version that *does* change the signature gets silently missed.                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                            
  Add the 1.5.0 contract, or a one-line comment saying 1.4.1 covers it.

Uxio0:
> `safe_events_indexer` and `proxy_factory_indexer` list `ProxyCreation` one entry per factory version (1.1.1 / 1.3.0 / 1.4.1), but 1.5.0 isn't there. Harmless today because `ProxyCreation` topic is byte-identical to 1.4.1's, so that entry already matches the logs. But nothing in the code says so, and the next version that _does_ change the signature gets silently missed.
> 
> Add the 1.5.0 contract, or a one-line comment saying 1.4.1 covers it.

Agree, comment should be enough, done (ammended the commit by mistake)

falvaradorodriguez (dismissed):

falvaradorodriguez (approved):

PR #2714, opened 2025-11-24
V1.5.0 support

akshay-ap:
## Safe v1.5.0 Support

This PR implements support for Safe Smart Account v1.5.0 while maintaining backward compatibility with v1.4.1. The transaction service now supports both versions simultaneously.

### Changes

**Database Schema**
- Renamed `guard` field to `transaction_guard` in `SafeStatus` and `SafeLastStatus` models
- Added `module_guard` field to `SafeStatus` and `SafeLastStatus` models
- Migration: `0096_guard_changes_v150.py`

**Event Indexing**
- Added support for `ChangedModuleGuard` event in Safe v1.5.0
- Added `setModuleGuard` function handler in transaction processor
- Updated event indexer tests to support v1.5.0 contract behavior

**Account Abstraction**
- Updated `decode_init_code` helper to support both v1.4.1 and v1.5.0 proxy factories
- Added `get_contract_instances` function that dynamically selects appropriate factory and contract versions based on factory address
- Updated tests to deploy both v1.4.1 and v1.5.0 proxy factory contracts

**API Changes**
- Updated `SafeInfoResponseSerializer` to include:
  - `transaction_guard` (renamed from `guard`)
  - `module_guard` (new field)
- Updated `SafeInfo.get_safe_info()` to correctly map fields to SafeInfo constructor parameters

**Test Updates**
- Refactored event indexer tests to handle v1.5.0 contracts
- Updated test expectations for dual proxy factory deployment (v1.4.1 and v1.5.0)
- Fixed SafeInfo parameter ordering bug that caused incorrect version values

**Testing Gaps**
- Account abstraction `decode_init_code` test for v1.5.0 is currently not implemented because safe-eth-py lacks mock UserOperation data with initCode with v1.5.0 proxy factory
- Manual testing not done yet

### TODO
- [ ] This PR depends on unreleased changes in `safe-eth-py`. The PR cannot be merged/reviewed until a new safe-eth-py version is published with v1.5.0 support.
- [x] Add `ExtensibleFallabackHandler` events to indexing. Should we even do it?
   Answer: As discussed with team, we are not adding it now.
- [x] Additional test where both v1.4.1 and v1.5.0 must be supported simultaneously
   - Added in `test_views.py` and `test_views_v2.py`
- [ ] Manual testing
- [x] Replace safe-eth-py version with appropriate release version in `requirements.txt`

---

> [!NOTE]
> Introduces full Safe v1.5.0 compatibility while preserving support for prior versions.
> 
[...]

PR #2968, opened 2026-07-30
Skip redundant L2 re-index when adding a new master copy

Uxio0:
## What

Stop `setup_service` from forcing a full L2 re-index when a new `SafeMasterCopy` is added (e.g. the 1.5.0 singleton).

## Why

`SafeEventsIndexer` sets `IGNORE_ADDRESSES_ON_LOG_FILTER = True`, so its `eth_getLogs` query filters by **topic only** — addresses are ignored. The address set feeds a single thing: the scan window, which is `Min(tx_block_number)` over `SafeMasterCopy.objects.l2()`.

`_setup_safe_singleton_addresses` created the new row with `tx_block_number = <deploy block>`. On an already-synced L2 that drags the window back to the (old) deploy block and forces a full topic re-scan from there to head. It's idempotent (dedup + `ignore_conflicts`) but fully redundant — every new-version Safe event shares an existing topic and was already indexed in that range.

L1 (`InternalTxIndexer`) genuinely needs the deploy block because it matches traces by master-copy address, so its behavior is unchanged.

## Change

- Read the L2 **frontier** (`Min(tx_block_number)`) and **scan floor** (`Min(initial_block_number)`) over `l2()` once.
- A new L2 singleton starts at the frontier **when its deploy block is at or above the scan floor** (its events were already covered by the shared topic scan → no re-scan).
- It falls back to the deploy block for L1, a fresh DB, or an L2 singleton deployed **below** the scan floor (an unscanned gap exists, so its early events must still be indexed).
- `initial_block_number` always keeps the real deploy block (it's the reported `deployed_block_number` and the admin reindex floor).
- Existing rows are never modified.

## Notes

- The frontier is read once and not synchronized with a running indexer; a concurrent `setup_service` run could seed a slightly stale value, which self-heals on the next scan. Accepted (documented inline).
- Not addressed here (deeper alternative, out of scope): tracking L2 event-indexing progress with a single `IndexingStatus` row instead of per-master-copy `tx_block_number`.
- Also documents the reST docstring convention in `AGENTS.md`.

## Tests

- `test_setup_safe_singleton_addresses_l2_frontier` — synced chain: new L2 row starts at the frontier, `initial_block_number` keeps the deploy block, `l2()` window Min unchanged, L1 row still at deploy block, existing row untouched.
- `test_setup_safe_singleton_addresses_l2_fresh_db` — fresh DB → new L2 row starts from its deploy block.
- `test_setup_safe_singleton_addresses_l2_below_scan_floor` — L2 singleton deployed below the scan floor → starts from its deploy block, not the frontier.

Closes PLA-1810.

### DIFF
--- a/safe_transaction_service/history/management/commands/setup_service.py
+++ b/safe_transaction_service/history/management/commands/setup_service.py
@@ -232,38 +232,52 @@ class Command(BaseCommand):
         """
         Insert new ``SafeMasterCopy`` singletons.
 
-        A new L2 singleton whose deploy block is at or above the L2 scan floor starts at the
-        current L2 frontier (lowest ``tx_block_number`` among existing L2 singletons) instead of
-        its deploy block: those events are already covered by the shared topic scan (addresses are
-        ignored on L2), so a lower start would needlessly drag the scan window back. An L2 singleton
-        deployed below the scan floor, a fresh DB, or an L1 singleton fall back to the deploy block.
-        ``initial_block_number`` always records the real deploy block.
+        Only on an **L2 network** does ``SafeEventsIndexer`` run, filtering logs by topic while
+        ignoring addresses. There a new ``+L2`` singleton whose deploy block is at or above the
+        lowest block L2 indexing has scanned from starts at the current L2 indexing position
+        (lowest ``tx_block_number`` among existing L2 singletons), bounded below by its own deploy
+        block, instead of always the deploy block: those events are already covered by the shared
+        scan, so a lower start would needlessly drag the scan window back.
+
+        Everything else starts from the deploy block: an L1/tracing network (where
+        ``InternalTxIndexer`` matches traces by address, and ``+L2`` singletons are tracked too, so
+        skipping a range would lose them), a fresh DB, or a ``+L2`` singleton deployed below the
+        lowest scanned block (an unscanned gap exists). ``initial_block_number`` always records the
+        real deploy block.
 
         :param safe_singleton_addresses: ``(address, deploy_block_number, version)`` tuples
         """
-        # Frontier: highest fully-scanned block (lowest tx_block_number). Floor: lowest block
-        # ever scanned (lowest initial_block_number). Read once; a concurrent indexer run could
-        # leave the frontier slightly stale, which self-heals on the next scan.
-        l2_blocks = SafeMasterCopy.objects.l2().aggregate(
-            frontier=Min("tx_block_number"),
-            floor=Min("initial_block_number"),
-        )
-        l2_frontier_block_number = l2_blocks["frontier"]
-        l2_scan_floor_block_number = l2_blocks["floor"]
+        # The shortcut is only valid where SafeEventsIndexer (topic scan, addresses ignored) is
+        # the indexer, i.e. on L2 networks. `l2_current_block_number` is the current L2 indexing
+        # position (lowest tx_block_number); `l2_min_initial_block_number` is the lowest block L2
+        # indexing has scanned from (lowest initial_block_number). Read once; a concurrent indexer
+        # run could leave them slightly stale, which self-heals on the next scan.
+        if settings.ETH_L2_NETWORK:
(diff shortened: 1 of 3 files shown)


safe_transaction_service/history/management/commands/setup_service.py
```diff
@@ -229,14 +229,49 @@ def handle(self, *args, **options):
...
         for address, initial_block_number, version in safe_singleton_addresses:
+            is_l2 = version.endswith("+L2")
+            if (
+                is_l2
+                and l2_frontier_block_number is not None
+                and initial_block_number >= l2_scan_floor_block_number
+            ):
+                # Deploy block is within the already-scanned range, so its events are covered
+                # by the shared topic scan: start at the frontier, skipping a redundant re-scan.
+                tx_block_number = l2_frontier_block_number
```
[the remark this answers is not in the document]
Uxio0: We are aware

safe_transaction_service/history/management/commands/setup_service.py
```diff
@@ -229,14 +229,49 @@ def handle(self, *args, **options):
...
+        # leave the frontier slightly stale, which self-heals on the next scan.
+        l2_blocks = SafeMasterCopy.objects.l2().aggregate(
+            frontier=Min("tx_block_number"),
+            floor=Min("initial_block_number"),
+        )
+        l2_frontier_block_number = l2_blocks["frontier"]
+        l2_scan_floor_block_number = l2_blocks["floor"]
+
         for address, initial_block_number, version in safe_singleton_addresses:
+            is_l2 = version.endswith("+L2")
```
falvaradorodriguez, line 266: I think that here we have a problem. On a tracing network this skips traces that will never be indexed. is_l2 is the singleton flavour, not the network — +L2 singletons exist on mainnet too, and there SafeEventsIndexer never runs (index_safe_events_task is disabled, :105), so "addresses are ignored" doesn't apply: InternalTxIndexer matches by address. Should we add "is an L2 network" condition here?

safe_transaction_service/history/management/commands/setup_service.py
falvaradorodriguez: Minor, probably not worth fixing: tx_block_number can end up below initial_block_number, but only if the chain is still catching up (on an empty DB frontier is None, so everything falls back to the deploy block).                                                                                                    
                                                                                                                                                                                                                                                                                                                     
Example: rows are at their deploy blocks, the backfill is at ~15M, then a release adds 1.5.0+L2 (deploy 22838494). floor is 12504423, so the row is created at 15M instead of 22838494.                                                                                                                                 
                                                                                                                                                                                                                                                                                                                       
There is no DB constraint, so the only real effect is on the admin rewinds: Greatest(F("tx_block_number") - N, F("initial_block_number")) (admin.py:465-486) returns 22838494, so "reindex last 24 hours" moves the row forward over unscanned blocks. max(l2_frontier_block_number, initial_block_number) would cover it if you think it's worth it.

safe_transaction_service/history/management/commands/setup_service.py
Uxio0: Fixed on 56750929

safe_transaction_service/history/management/commands/setup_service.py
Uxio0: Fixed on 56750929

falvaradorodriguez (approved):

safe_transaction_service/history/management/commands/setup_service.py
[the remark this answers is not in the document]
Uxio0: Service has being indexing 1.5.0 events since the contracts were deployed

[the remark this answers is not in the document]
falvaradorodriguez: ChangedModuleGuard was added to the indexer on 2026-01-05 (7e4af798, #2714, released in v6.0.0).

PR #2987, opened 2026-08-18
Add sanctioned signer EOAs to BANNED_EOAS

gjeanmart:
Adds 16 sanctioned signer EOAs to the `BANNED_EOAS` setting so the service rejects transactions and messages signed by them, and moves the IC3 PSA source link next to the batch it actually documents.

Also documents the EIP-55 requirement for entries: they are compared against signature-recovered owners, which are always checksummed, so a badly-cased entry would silently never match.

Ref: PLA-1886

### What was wrong? 👾

Closes #

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -793,11 +793,13 @@ BANNED_SAFES_CACHE_TTL = env.int(
     "BANNED_SAFES_CACHE_TTL", default=60 * 5
 )  # Seconds the in-memory set of banned Safe addresses is cached for (default 5m)
 
-# Addresses not allowed to interact with the service
-# List taken from https://www.ic3.gov/PSA/2025/PSA250226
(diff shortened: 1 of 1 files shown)


falvaradorodriguez (approved):

config/settings/base.py
```diff
@@ -849,6 +851,23 @@
         "0xf03AfB1c6A11A7E370920ad42e6eE735dBedF0b1",
         "0xEB0bAA3A556586192590CAD296b1e48dF62a8549",
         "0xD5b58Cf7813c1eDC412367b97876bD400ea5c489",
+        # Sanctioned signer EOAs, see PLA-1886
```
moisses89, line 854: Have we a public list for this instead the private ticket?

gjeanmart: No we don't

PR #2998, opened 2026-09-09
feat: Decode creation data when using the Rhinestone relay

### LINKED ISSUE #2996: Decode creation data when using the Rhinestone relay
# What is needed?

`GET /api/v1/safes/<address>/creation/` should return `setupData` and `saltNonce` for Safes created through the Rhinestone relay.

# Background

`safe-client-gateway` switched its relay provider to Rhinestone in https://github.com/safe-global/safe-client-gateway/pull/3244 (2026-08-20). A Safe creation sent to `POST /v1/chains/10/relay` lands on chain wrapped in a forwarder call rather than as a direct factory call, so `_decode_creation_data` in `safe_transaction_service/history/services/safe_service.py` misses it.

The existing `_decode_gelato_relay` unwraps relayed creations with `gelato_relay_1_balance_v2_abi`. Calldata from the Rhinestone relay could not be resolved and trigger `ValueError`.

Reproduced on OP Mainnet, creation tx `0x44784c80464169647dc1346e9370a098e3a890efdc0764c106fd4f7866948269`.

# Endpoint

## URL

`GET https://api.safe.global/tx-service/oeth/api/v1/safes/0x25495528556D64A3551CF9F8dBEb7EC86325D728/creation/`

## Response

Today:

```
{
  "created": "2026-09-08T04:02:49Z",
  "creator": "0x46565eeDB0bb948aE787dD105cCE300f736dA28e",
  "transactionHash": "0x44784c80464169647dc1346e9370a098e3a890efdc0764c106fd4f7866948269",
  "factoryAddress": "0x4e1DCf7AD4e460CfD30791CCC4F9c8a4f820ec67",
  "masterCopy": "0x41675C099F32341bf84BFc5382aF534df5C7461a",
  "setupData": "0x",
  "saltNonce": null,
  "dataDecoded": null,
  "userOperation": null
}
```

Expected, with the values carried in the relayed calldata:

```
{
  ...
  "setupData": "0xb63e800d0000...0000",
  "saltNonce": "1788840152716",
  "dataDecoded": { "method": "setup", "parameters": [ ... 8 parameters ... ] }
}
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

dasanra:

### What was wrong? 👾

Closes #2996 

### How was it fixed? 🎯

Add the SafeRelayExecutor ABI and replace the Gelato-only unwrapper with a generic _decode_relay that tries each supported relayer before MultiSend and ProxyFactory decoding. Gelato stays supported: creation data is decoded
at read time from the stored transaction, so removing it would regress every Safe created through the Gelato relay.
Add a fixture from the OP Mainnet creation tx 0x44784c80464169647dc1346e9370a098e3a890efdc0764c106fd4f7866948269 and cover both relayers plus the non-relayed passthrough.


safe_transaction_service/history/services/safe_service.py
```diff
@@ -305,20 +319,26 @@ def _decode_creation_data(self, data: bytes) -> list[ProxyCreationData]:
...
-                self.gelato_relay_1_balance_v2_contract.decode_function_input(data)
-            )
-            return HexBytes(decoded_gelato_data["_data"])
-        except ValueError:
-            return data
+        for relay_contract, data_parameter in self.relay_contracts:
+            try:
+                _, decoded_relay_data = relay_contract.decode_function_input(data)
+                return HexBytes(decoded_relay_data[data_parameter])
+            except ValueError:
```
Uxio0, line 339: `except ValueError` does not catch every decoding error, so the loop can raise instead of falling through to the next relayer.

If the first 4 bytes match a relayer selector but the rest of the payload has a different shape, `web3` delegates to `eth_abi`, which raises `InsufficientDataBytes` or `NonEmptyPaddingBytes`. Both come from `eth_abi.exceptions.DecodingError`, which inherits from `Exception`, not from `ValueError`.

Easy to reproduce:

```python
contract.decode_function_input(bytes.fromhex("710a9f68" + "ff" * 96))
# eth_abi.exceptions.NonEmptyPaddingBytes
```

The error goes up through `_decode_creation_data` -> `_process_creation_data` -> `get_safe_creation_info`, and that `try` only handles `InternalTx.DoesNotExist` and `OSError`, so we return a 500 instead of `None`.

This was already the case for the old `_decode_gelato_relay` (and `_decode_proxy_factory` / `_decode_cpk_proxy_factory` do the same), but now we have two selectors and one of them is quite generic, so it is easier to hit. Suggestion:

```python
from eth_abi.exceptions import DecodingError

...
            except (ValueError, DecodingError):
                continue
```

safe_transaction_service/history/services/safe_service.py
```diff
@@ -305,20 +319,26 @@ def _decode_creation_data(self, data: bytes) -> list[ProxyCreationData]:
...
+        :return: Forwarded `data` if a relayer is detected, original `data` otherwise
         """
-        try:
-            _, decoded_gelato_data = (
-                self.gelato_relay_1_balance_v2_contract.decode_function_input(data)
-            )
-            return HexBytes(decoded_gelato_data["_data"])
-        except ValueError:
-            return data
+        for relay_contract, data_parameter in self.relay_contracts:
```
Uxio0, line 335: The unwrap only looks at the selector, and a single decoded result is never checked against the logs.

`_decode_relay` does not know the `to` of the transaction, so any calldata starting with `execute(uint256,address,bytes)` is unwrapped and the outer call is dropped. That signature is generic and used by other executors, not only Rhinestone.

Possible case: a tx where the top level call is an unrelated `execute(uint256,address,bytes)` carrying a `createProxyWithNonce` for Safe B, while Safe A is deployed in the same tx through another path. `_decode_creation_data` returns only one result (B) and `_process_creation_data` returns it directly in the `len(...) == 1` branch, so `get_deployed_proxies_from_logs()` is never called; the log check only runs when there is more than one result. Safe A would end up with the singleton, initializer and saltNonce of B.

It is an edge case, but the old Gelato `sponsoredCallV2` selector was specific enough that it could not happen. Two options to close it: check the decoded `target` against the known relayer/factory addresses, or also run the log check when there is a single result.

Uxio0 (commented):
Change looks correct. I checked the OP Mainnet tx used for the mock and the calldata is byte-identical, and `0x710a9f68` is indeed `execute(uint256,address,bytes)`. Two robustness notes below, neither is blocking.

safe_transaction_service/history/services/safe_service.py
dasanra: Fixed in the three functions

Uxio0 (approved):

safe_transaction_service/history/services/safe_service.py
dasanra: Fixed in https://github.com/safe-global/safe-transaction-service/pull/2998/commits/0d35b5cd5b7c1fc75bb8408063d560003e0daddc

